PR #4, opened 2021-08-03
Implement the WHATWG URL and URLSearchParam classes

### LINKED ISSUE #2: Missing globals `URL` and `URLSearchParams`
Hi, this is likely the wrong place to open this issue but I could not find a public repository for the `@fastly/js-compute` npm package

I wanted to report that `URL` and `URLSearchParams` are [documented on the website](https://js-compute-reference-docs.edgecompute.app/classes/url.html) to exist but they don't actually exist when running the code using `fastly compute serve` or deploying it to Fastly

I can see that the [compute-starter-kit-javascript-default](https://github.com/fastly/compute-starter-kit-javascript-default) uses Webpack and a polyfill from Core-JS to provide these globals - would it be possible to provide these polyfills within the `@fastly/js-compute` package itself?

https://github.com/fastly/compute-starter-kit-javascript-default/blob/22fe9f050d8a2c715d23da5ea29935765192566d/webpack.config.js#L15-L19

tschneidereit:
This is a full implementation of the WHATWG URL and URLSearchParam classes, based on the [rust-url](https://crates.io/crates/url), which is also used in Firefox.

It's quite a bit of code, but much of it is relatively boring plumbing. The noteworthy parts are around the integration of the Rust code in the `c-dependencies/js-compute-runtime/rust-url/` directory, and the use if `cbindgen` for the FFI.

Closes #2


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3238,6 +3282,583 @@ namespace FetchEvent {
...
+
+namespace URLSearchParams {
+  namespace Slots { enum {
+    Url,
+    Params,
+    Count
+  };};
+
+  namespace detail {
+    bool append(JSContext* cx, HandleObject self, HandleValue key, HandleValue val, const char* fun_name) {
```
cfallin, line 3584: `fun_name` is unused here -- was it meant to go into an error message of some sort maybe?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3238,6 +3282,583 @@ namespace FetchEvent {
...
+      const SpecSlice value = values.ptr[i];
+      str = JS_NewStringCopyUTF8N(cx, JS::UTF8Chars((char*)value.data, value.len));
+      if (!str) return false;
+
+      str_val.setString(str);
+      if (!JS_SetElement(cx, result, i, str_val))
+        return false;
+    }
+
+    free(values.ptr);
```
cfallin, line 3692: Do we have a guarantee that Rust-on-wasm uses wasi-libc's malloc/free? I suppose we'd have much deeper problems if that weren't the case, but this asymmetry still bothers me a tiny bit; I'd prefer for the `values` to be passed back into a Rust FFI call to be freed on that side.

cfallin (approved):
Largely LGTM! Just a few points below.

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -16,23 +12,39 @@ WASI_CXX ?= /opt/wasi-sdk/bin/clang++
...
 
 .PHONY: all clean
 
 all: js-compute-runtime.wasm
 
 -include $(FSM_DEP)
 
-%.o: $(FSM_SRC)%.cpp $(FSM_SRC)Makefile
-	$(WASI_CXX) $(CXX_FLAGS) $(DEFINES) -I $(SM_SRC)include -O3 -MMD -MP -c -o $@ $<
+$(RUST_URL_LIB): $(RUST_URL_SRC)/src/*.rs $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)/cbindgen.toml $(FSM_SRC)Makefile
```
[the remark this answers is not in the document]
tschneidereit: Good call, added!

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -16,23 +12,39 @@ WASI_CXX ?= /opt/wasi-sdk/bin/clang++
 WIZER ?= wizer
 
 CXX_FLAGS := -std=gnu++17 -Wall -Werror
+CXX_OPT ?= O2
+
+ifeq (,$(findstring g,$(CXX_OPT)))
+ifneq (, $(shell which wasm-opt))
```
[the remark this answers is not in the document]
tschneidereit: It seems like the whitespace here didn't make a difference, at least on my machine. But yeah, better safe than sorry, so I removed it.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: For this particular case, I was able to avoid this by just adding a `Drop` impl for `CVec`—which is the right thing to do anyway.

However, on a somewhat fundamental level we really can't avoid this: we have to pass strings to Rust that were created by SpiderMonkey (I think in all cases via the `encode` function). I guess we could in theory never have the Rust code take ownership of them and always copy, but that seems pretty unfortunate.

I also think this should be something we can indeed rely on: SpiderMonkey contains lots of Rust code precisely for string handling, so I would strongly assume that the build config sorts out all the allocator stuff :)

tschneidereit:
Thank you for the review! I pushed a small patch addressing your comments—mind taking another quick look, @cfallin?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
cfallin: Fair enough!

cfallin (approved):

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: ah, I forgot to address this. The parameter is required because `append` is used as a template parameter in `init_class`. I renamed it to `_` to make clear it's unused, though.

cfallin (approved):

PR #8, opened 2021-08-05
Make FetchEvent handling more spec-compliant

tschneidereit:
This introduces three significant changes:
1. Support multiple `"fetch"` event handlers instead of just one
2. Support `Promise<Response>` in addition to `Response` in `FetchEvent#respondWith`
3. Require `FetchEvent#respondWith` to be called synchronously from within a `"fetch"` handler

This last one is a breaking change: before, we supported invoking `respondWith` at any point. That isn't what the spec wants us to do, and it also means that we don't have a good way to determine when a `FetchEvent` has been fully handled. Our previous solution to that was to wait until all pending tasks have been completed, but that can potentially mean running an instance for much longer than the author intended.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -3238,6 +3238,10 @@ namespace FetchEvent {
     RootedObject event(cx, &js::GetFunctionNativeReserved(&args.callee(), 0).toObject());
 
     // Step 10.1
+    // Note: the `then` handler is only invoked after all Promise resolution has happened.
+    // (Even if there were multiple Promises to unwrap first.)
+    // That means that at this point we're guaranteed to have the final value instead of a
+    // Promise wrapping it, so either the value is a Response, or we have to bail.
     if (!Response::is_instance(args.get(0))) {
       JS_ReportErrorUTF8(cx, "FetchEvent#respondWith must be called with a Response "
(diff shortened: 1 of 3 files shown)


tschneidereit:
The breakage here is almost certainly caused by a mismatch between Rust versions used for compiling SpiderMonkey and the runtime. I'll look into either updating the former, or pinning the latter to an older version.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3209,28 +3222,33 @@ namespace FetchEvent {
...
-    METHOD_HEADER(1)
+  // Steps in this function refer to the spec at
+  // https://w3c.github.io/ServiceWorker/#fetch-event-respondwith
+  bool then_handler(JSContext* cx, unsigned argc, Value* vp) {
+    CallArgs args = CallArgsFromVp(argc, vp);
+    RootedObject event(cx, &js::GetFunctionNativeReserved(&args.callee(), 0).toObject());
 
-    // TODO: support Promise<Response>
+    // Step 10.1
     if (!Response::is_instance(args.get(0))) {
```
cfallin, line 3241: This was a bit confusing at first given the error message just below referring to promises. Could you add a comment here pointing to `respondWith` below and indicating that any promises are resolved by this point?

cfallin (approved):
LGTM! Just one request for a clarifying comment but the logic looks reasonable to me.

PR #13, opened 2021-08-09
Enable private class fields and methods, plus ergonomic brand checks

tschneidereit:
These features are stable and enabled in Firefox by default, but not in Spidermonkey's default configuration, so we have to manually enable them.

(Depends on #12, so only the last patch is relevant here.)


cfallin (approved):

tschneidereit:
Bleh, I accidentally included the full comment including the text for the previous commit that we rebased on in reality :(

PR #14, opened 2021-08-09
Implement FetchEvent#waitUntil

tschneidereit:
Before this patch, we always kept the service running until all async tasks had been processed. That can result in services running longer than expected though, e.g. when a "fire-and-forget" logging task is implemented using `fetch`.

The ServiceWorkers spec defines the `FetchEvent`'s lifetime to include resolution of any (potentially async) results passed to `FetchEvent#respondWith`, and of any promises passed to `FetchEvent#waitUntil`.

With this patch, we shut down the service as soon as the `FetchEvent`'s lifetime has ended. E.g., the following code will trigger a backend request, wait until it has returned, and send it on as a response:
```js
  event.respondWith(fetch("http://example.com", { backend: "example_backend" }));
});
```

This code however will trigger a backend request, but shut down right after the "fetch" event has completed dispatching, without waiting for the backend request to return (so the console message will never be logged):
```js
  let beReq = fetch("http://example.com", { backend: "example_backend" });
  beReq.then(beResp => console.log("Backend response received"));
  event.respondWith(new Response("synthetic"));
});
```

And finally, this code however will trigger a backend request, send off a synthetic response, but then wait until the "fetch" event has completed dispatching, and the attached `then` handler has run:
```js
  let beReq = fetch("http://example.com", { backend: "example_backend" });
  event.waitUntil(beReq);
  beReq.then(beResp => console.log("Backend response received"));
  event.respondWith(new Response("synthetic"));
});
```

### DIFF
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -1,18 +1,20 @@
 # DEFINES := -DDEBUG -DJS_DEBUG
 
 INIT_JS ?= test.js
+WIZER ?= wizer
 
 ROOT_SRC?=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))/..
 SM_SRC=$(ROOT_SRC)/spidermonkey/
-SM_OBJ=$(SM_SRC)lib/*
+SM_OBJ=$(SM_SRC)lib/*.o
+SM_OBJ+=$(SM_SRC)lib/*.a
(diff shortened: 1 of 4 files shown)


cfallin (approved):
LGTM. The refcounting looks right to me, but of course it'd be nice to have tests for this too, as it's somewhat subtle async code -- I know that there is a bigger plan for this so it's fine for now!

tschneidereit:
Yeah, very much agreed on the testing, and working that out will be my next highest priority after getting these breaking changes out the door :)

PR #15, opened 2021-08-09
Normalize HTTP method names when constructing Requests

### LINKED ISSUE #9: Support non-uppercase values for `method` in Request and Response
The fetch spec says to treat `method` as case-insensitive, but our host environment requires it to be all-uppercase, so we need to convert to that before passing values to the host.

tschneidereit:
The fetch spec [requires uppercasing](https://fetch.spec.whatwg.org/#concept-method-normalize) for exactly 6 HTTP method names, but not any others 🙃

Fixes #9

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1272,6 +1272,8 @@ static bool normalize_http_method(char* method) {
       if (strcmp(method, name) == 0) {
         return false;
       }
+
+      // Note: Safe because `strcasecmp` returning 0 above guarantees same-length strings.
       strcpy(method, name);
       return true;
     }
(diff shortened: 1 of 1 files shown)


tschneidereit:
Note: stacked on #12

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1259,6 +1260,26 @@ namespace CacheOverride {
...
+static bool normalize_http_method(char* method) {
+  static const char* names[6] = { "DELETE", "GET", "HEAD", "OPTIONS", "POST", "PUT" };
+
+  for (size_t i = 0; i < 6; i++) {
+    auto name = names[i];
+    if (strcasecmp(method, name) == 0) {
+      if (strcmp(method, name) == 0) {
+        return false;
+      }
+      strcpy(method, name);
```
cfallin, line 1275: For the benefit of any auditors in the future, can we add a safety comment here noting that `name` and `method` have the same length (because they compare equal modulo case) so `strcpy` is safe?

cfallin (approved):
LGTM!

PR #22, opened 2021-08-12
Add support for including binary files, and for using typed arrays as `Response` bodies

tschneidereit:
It's often desirable to be able to embed binary assets into a wasm file. Eventually we'll have a standardized way to do so, but for now we don't.

This patch thus adds `fastly.includeBytes` as the moral equivalent to Rust's [include_bytes!](https://doc.rust-lang.org/1.54.0/std/macro.include_bytes.html) macro. `includeBytes` can only be used during the initialization/wizening phase, because that's the only time filesystem is (through the one-line change in `src/main.rs`) available.

The way binary data is embedded is pretty close to optimal: the data is without any intermediate copies placed into a buffer underlying the `Uint8Array` that `includeBytes` returns. That way, the data is part of the wizer snapshot without any overhead.

Making full use of these included assets requires the ability to also directly return them as responses, so this patch also adds support for passing typed array, `DataView`, or `ArrayBuffer` instances to the `Response` constructor, which in turn passes them on to the host without any intermediate copies.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -17,6 +17,7 @@
 #include "js/JSON.h"
 #include "js/shadow/Object.h"
 #include "js/Stream.h"
+#include "js/experimental/TypedData.h"
 #include "js/Value.h"
 
 using JS::CallArgs;
@@ -321,6 +322,13 @@ static const uint32_t class_flags = 0;
     return false; \
   }
 
+#define INIT_ONLY(name) \
+  if (FetchEvent::instance()) { \
+    JS_ReportErrorUTF8(cx, "%s can only be used during initialization, " \
+                           "not during request handling", name); \
+    return false; \
+  }
+
 namespace Logger {
   namespace Slots { enum {
     Endpoint,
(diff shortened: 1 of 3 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -321,6 +322,13 @@ static const uint32_t class_flags = 0;
     return false; \
   }
 
+#define INIT_ONLY(name) \
+  if (FetchEvent::instance()) { \
```
cfallin, line 326: Just an idle thought, but perhaps it would be better (in this PR or later) to have a global "execution mode" enum that indicates whether we know we're wizening currently? This way we affirmatively assert that we're in the right mode rather than assume the absence of a fetch event implies it; I can imagine a distant future where there might be other-than-fetch events that also invoke the module.

cfallin (approved):
A very nice feature!

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: Good point, yes! I'll file an issue about this.

Issue #11, opened 2021-08-05
Improve handling of hostcall errors


tschneidereit:
In a lot of places, we currently turn a hostcall error code into a JS exception containing the error code and the C++ source location of the hostcall. That's very rarely helpful in debugging JS code, and we should improve on it :)

JakeChampion:
@tschneidereit I just came across the mapping of hostcall error codes within the [fastly-shared crate](https://docs.rs/fastly-shared/0.6.3/src/fastly_shared/lib.rs.html#35-96) and thought I'd have a go at contributing to this project -- I've started a [draft pull-request](https://github.com/fastly/js-compute-runtime/pull/24) with the error messages from the crate

PR #24, opened 2021-08-15
Improve handling of hostcall errors

JakeChampion:
Added more meaningful messages to hostcall errors

The mapping from error code number to message comes from https://docs.rs/fastly-shared/0.6.3/src/fastly_shared/lib.rs.html#35-96

Resolves https://github.com/fastly/js-compute-runtime/issues/11

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -111,12 +111,50 @@ SpecString encode(JSContext* cx, HandleValue val) {
 /* Returns false if an exception is set on `cx` and the caller should immediately
    return to propagate the exception. */
 static inline bool handle_fastly_result(JSContext* cx, int result, int line, const char* func) {
-    if (result == 0) {
-        return true;
-    } else {
-        fprintf(stdout, __FILE__":%d (%s) - Fastly error code %d\n", line, func, result);
-        JS_ReportErrorUTF8(cx, "Fastly error code %d", result);
-        return false;
+    switch(result)
+    {
+        case 0:
+            return true;
+        case 1:
+            JS_ReportErrorUTF8(cx, "%s: Generic error value. This means that some unexpected error occurred during a hostcall. - Fastly error code %d\n", func, result);
+            return false;
+        case 2:
+             JS_ReportErrorUTF8(cx, "%s: Invalid argument. - Fastly error code %d\n", func, result);
+            return false;
+        case 3:
+             JS_ReportErrorUTF8(cx, "%s: Invalid handle. Thrown when a request, response, dictionary, or body handle is not valid. - Fastly error code %d\n", func, result);
+            return false;
+        case 4:
+             JS_ReportErrorUTF8(cx, "%s: Buffer length error. Buffer is too long. - Fastly error code %d\n", func, result);
+            return false;
+        case 5:
+             JS_ReportErrorUTF8(cx, "%s: Unsupported operation error. This error is thrown when some operation cannot be performed, because it is not supported. - Fastly error code %d\n", func, result);
+            return false;
+        case 6:
+             JS_ReportErrorUTF8(cx, "%s: Alignment error. This is thrown when a pointer does not point to a properly aligned slice of memory. - Fastly error code %d\n", func, result);
+            return false;
+        case 7:
+             JS_ReportErrorUTF8(cx, "%s: HTTP parse error. This can be thrown when a method, URI, header, or status is not valid. This can also be thrown if a message head is too large. - Fastly error code %d\n", func, result);
+            return false;
+        case 8:
+             JS_ReportErrorUTF8(cx, "%s: HTTP user error. This is thrown in cases where user code caused an HTTP error. For example, attempt to send a 1xx response code, or a request with a non-absolute URI. This can also be caused by an unexpected header: both `content-length` and `transfer-encoding`, for example. - Fastly error code %d\n", func, result);
+            return false;
+        case 9:
+             JS_ReportErrorUTF8(cx, "%s: HTTP incomplete message error. A stream ended unexpectedly. - Fastly error code %d\n", func, result);
+            return false;
+        case 10:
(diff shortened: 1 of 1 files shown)


tschneidereit:
Thank you for doing this, this is already a great improvement over what we have now! ❤️

I think we should go ahead get this mergeable, but it doesn't quite resolve #11 yet: what I had in mind for that was to change the call sites to generate more meaningful JS exceptions.

To give just one somewhat random example, when trying to [set the HTTP status in the `Response` constructor](https://github.com/fastly/js-compute-runtime/blob/main/c-dependencies/js-compute-runtime/js-compute-builtins.cpp#L1754) to an invalid value, ideally we'd throw a JS exception saying something like `Response constructor: Invalid response status code` (which is what Firefox does) or even `Failed to construct 'Response': The status provided (9) is outside the range [200, 599]` (which is what Chrome does for the value `9`).

For this particular example, doing this wouldn't be terrifically hard, but also not entirely trivial. There are other cases where it'd be significantly more work, with various cases to consider, etc. So I expect this to be a gradual process of improvement over time :)

For now, let's focus on getting this landed, as it is very clearly an improvement!

There are two things I think we need to change: we do need to keep the `JS_ReportErrorUTF8`, because otherwise we'll not throw any exceptions anymore at all, and we need to have the function always return a `bool`.

I think what we can do is remove the `fprintf` calls entirely, and instead only throw the exception. To take `case 1` as an example, it could look something like this:
```c++
            JS_ReportErrorUTF8(cx, "%s: Generic error value. This means that some unexpected error occurred during a hostcall. - Fastly error code %d\n", func, result);
            return false;
```

Does that make sense? And if so, are you up for applying these changes?

JakeChampion:
@tschneidereit Thanks for the swift and detailed review -- I'll implement thoes changes 👍 (This is the first C++ project I've worked on, it's fun to move from writing JavaScript polyfills to making pull-requests to a JavaScript runtime)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -111,12 +111,51 @@ SpecString encode(JSContext* cx, HandleValue val) {
...
+        case 11:
+             JS_ReportErrorUTF8(cx, "%s: HTTP head too large error. This error will be thrown when the message head is too large. - Fastly error code %d\n", func, result);
+            return false;
+        case 12:
+             JS_ReportErrorUTF8(cx, "%s: HTTP invalid status error. This error will be thrown when the HTTP message contains an invalid status code. - Fastly error code %d\n", func, result);
+             return false;
+        default:
+            fprintf(stdout, __FILE__":%d (%s) - Fastly error code %d\n", line, func, result);
+            JS_ReportErrorUTF8(cx, "Fastly error code %d", result);
+            break;
```
tschneidereit, line 158: This should I think make the CI error go away:
```suggestion
            return false;
```

tschneidereit (changes requested):
Thank you, this is great! Just one tiny change needed, then this I think should be good to go 👏🏻

tschneidereit:
> @tschneidereit Thanks for the swift and detailed review -- I'll implement those changes 👍 (This is the first C++ project I've worked on, it's fun to move from writing JavaScript polyfills to making pull-requests to a JavaScript runtime)

You're more than welcome—thank *you* for working on this!

FWIW, we're working on getting the next release out the door to have all the changes that landed here in the last 2.5 weeks in developers' hands ASAP, but after that we'll put some work into adding a public test harness to this repo, so we don't have to rely on just the internal C@E test suite, run after changes have landed here. That should hopefully make contributing to the runtime a bit easier going forward :)

JakeChampion:
> FWIW, we're working on getting the next release out the door to have all the changes that landed here in the last 2.5 weeks in developers' hands ASAP, but after that we'll put some work into adding a public test harness to this repo, so we don't have to rely on just the internal C@E test suite, run after changes have landed here. That should hopefully make contributing to the runtime a bit easier going forward :)

That's excellent 🙌

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -111,12 +111,51 @@ SpecString encode(JSContext* cx, HandleValue val) {
...
-        return true;
-    } else {
-        fprintf(stdout, __FILE__":%d (%s) - Fastly error code %d\n", line, func, result);
-        JS_ReportErrorUTF8(cx, "Fastly error code %d", result);
-        return false;
+    switch(result)
+    {
+        case 0:
+            return true;
+            break;
```
tschneidereit, line 118: Tiniest of nits: 
```suggestion
```

tschneidereit (approved):
Just one tiny thing to go, and then I'll kick off the CI run and merge once it comes back green :)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -111,12 +111,50 @@ SpecString encode(JSContext* cx, HandleValue val) {
...
+             JS_ReportErrorUTF8(cx, "%s: A `None` error. This status code is used to indicate when an optional value did not exist, as opposed to an empty value. - Fastly error code %d\n", func, result);
+            return false;
+        case 11:
+             JS_ReportErrorUTF8(cx, "%s: HTTP head too large error. This error will be thrown when the message head is too large. - Fastly error code %d\n", func, result);
+            return false;
+        case 12:
+             JS_ReportErrorUTF8(cx, "%s: HTTP invalid status error. This error will be thrown when the HTTP message contains an invalid status code. - Fastly error code %d\n", func, result);
+             return false;
+        default:
+            fprintf(stdout, __FILE__":%d (%s) - Fastly error code %d\n", line, func, result);
```
JakeChampion, line 155: I kept the `fprintf` here as I found it was useful when opening an issue on this repository to be able to point to the line of code in which triggered the error code

tschneidereit: Yeah, that makes sense. It could even be a `MOZ_RELEASE_ASSERT`, as this really shouldn't happen—but that's absolutely not important, so we'll keep it as-is for this round :)

tschneidereit (approved):
Thank you so much!

PR #25, opened 2021-08-17
Add changelog for 0.2.0 release

tschneidereit:


torch2424 (approved):
Awesome! Thank you for updating the changelog! It's cool to see everything done in a single place! 😄 

~~I just wanted to confirm though, this Changelog should not be ingested by Devrel, or made public right? As since this is a private repo, and the Changelog formatting rules, I don't think they'd support links to the relevant PRs. But! If not (which I assume not), let's merge this!~~ 😄 🎉 

This repo is totally public 😂 Well! As long as we don't plan on having Fastly Docs ingest this Changelog, we should be good to go!

PR #38, opened 2021-09-13
Resolve URLs passed to `Request` and `fetch` relative to the client request URL's origin

tschneidereit:
Before this change, we didn't support relative URLs at all, which breaks expectations of some existing content—most notably the harness for WPT.

Per spec, the `Request` constructor resolves passed-in URLs relative to the environment's [API base URL](https://html.spec.whatwg.org/multipage/webappapis.html#api-base-url). In service workers, that is the service worker script's URL. The question is, what's the right value to use for C@E services? The script for a service doesn't have its own URL, exactly, after all.

The two options I can see are a) the URL of the incoming client request, and b) only the `origin` part (the protocol + host + port) of the incoming client request's URL. This patch implements b), as the other option would mean that the meaning of relative URLs changes based on the requested path, which doesn't seem desirable.

Since it might be useful to change the URL used as a base on a service-wide basis, this change also introduces `fastly.baseURL` to do so.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -418,6 +418,12 @@ namespace Logger {
   }
 }
 
+namespace URL {
+  bool is_instance(JS::Value);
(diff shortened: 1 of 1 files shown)


cfallin (approved):
LGTM!

PR #39, opened 2021-09-13
Enable defining a default backend using `fastly.defaultBackend`

tschneidereit:
When using existing libraries, it can be difficult to pass in the backend to use in `fetch`. With this change, it's possible to define a default backend to use when none is passed in to the `fetch` call.

Note that this isn't ideal in situations where multiple backends are used. An alternative that's been discussed is to derive the backend from the request URL's host name instead. We might do that before the next SDK release.

In the meantime, this helps get WPT tests running — see the next PR.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -538,6 +538,7 @@ namespace Fastly {
   };
 
   static PersistentRooted<JSObject*> baseURL;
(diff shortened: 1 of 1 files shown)


tschneidereit:
Note: this stacks on #38, so only the last commit is relevant here.

cfallin (approved):
LGTM. I share the feeling of slight discomfort with global/ambient state here and agree that something that derives backend config from the URL itself would be better; but I know those discussions are ongoing and this gets something to work now, so :+1:

PR #40, opened 2021-09-13
Add Web Platform Tests harness

tschneidereit:
Since much of our API surface aims to implement existing web specifications like [fetch](https://fetch.spec.whatwg.org/), [Service Workers](https://w3c.github.io/ServiceWorker/), [Streams](https://streams.spec.whatwg.org/), and [encoding](https://encoding.spec.whatwg.org/), the best way to get solid test coverage for our builtins is to run the official test suite for these specifications, [WPT](https://web-platform-tests.org/).

This change introduces a custom harness for
- running WPT tests
- evaluating the results against a set of expectations in a CLI to be used in CI
- evaluating the results against a set of expectations in an interactive web interface
- updating expectations

I think much of this will not be the final word on how to run WPT tests, but it works pretty well for the time being.

This change also comes with a list of tests to run (in `tests/wpt-harness/test.json`) and associated expectations (under `tests/wpt-harness/expecations`). These will be added to over time as we improve spec adherence.

And finally, this change also modifies the GitHub Actions test workflow to run the WPT harness to ensure that we don't accidentally regress spec adherence.


tschneidereit:
I had honestly not expected the tests to [pass](https://github.com/fastly/js-compute-runtime/pull/40/checks?check_run_id=3587591158#step:8:1) on the very first attempt 😬  Mainly because I wasn't sure I had everything right on preparing the `wpt.py` server to run and installing dependencies.

Speaking of dependencies, I want to get those cached before landing this: compiling Wizer and Viceroy takes up 17 of 18 minutes for this test ...

.github/workflows/main.yml
```diff
@@ -66,6 +66,41 @@ jobs:
...
+      with:
+        submodules: true
+
+    - name: Download Engine
+      uses: actions/download-artifact@v1
+      with:
+        name: engine
+
+    - name: Install Viceroy
+      run: cargo install viceroy --version 0.2.4
```
JakeChampion, line 84: This is how I would cache the cargo install, I do similar things in polyfill-library's workflows
```suggestion
      - name: add viceroy-version env
        run: echo "viceroy-version=0.2.4" >> $GITHUB_ENV
      - name: cache viceroy
        id: cache-viceroy
        uses: actions/cache@v2.1.5
        with:
          path: "/home/runner/.cargo/bin/viceroy"
          key: cache--dist--${{ env.viceroy-version }}
      - name: Install Viceroy
        run: cargo install viceroy --version 0.2.4
        if: steps.cache-viceroy.outputs.cache-hit != 'true'
```

tschneidereit: Thank you, this was an excellent starting point! I ended up doing [something slightly more involved](https://github.com/fastly/js-compute-runtime/pull/40/commits/175fc53dc9fafbd1aca6dca73ee8b3a925b3e151), both because we have multiple crates to install, and because I want them to build in parallel, but it would've taken me much longer to get to a solution without this suggestion <3

tschneidereit:
Ok, this all seems to work surprisingly well! 🥳  I have to add a bunch of documentation, and do some cleanup, but once that's done, this should be ready to go

tests/wpt-harness/viceroy.toml
```diff
@@ -0,0 +1,4 @@
+[local_server]
+  [local_server.backends]
+    [local_server.backends.wpt]
+      url = "http://web-platform.test:8000"
```
torch2424, line 4: Does this URL require any special changes to the /etc/hosts file? 😯

tests/wpt-harness/tests.json
```diff
@@ -0,0 +1,162 @@
+
```
torch2424, line 1: Just want to make sure that these files should point to the `.js` and not the `.json` correct? 😄  (I've never used the wpt toests, so I genuinely don't know 😅 )

torch2424 (approved):
This Looks great to me! Thank you for all the hard work here! 😄 

This PR seems to mostly be JSON test validationg, but I left a few comments, none of them should be blocking 😄

tests/wpt-harness/viceroy.toml
tschneidereit: it does, yes. Those changes happen [here](https://github.com/fastly/js-compute-runtime/pull/40/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R150).

tests/wpt-harness/tests.json
tschneidereit: yeah, these are the actual test files to run. The `json` files are the expectations we store for the test runs, and we derive their paths from the `js` file path. Could go the other way in theory, but since we can run tests that don't yet have expectations attached, that'd be a bit weird :)

PR #41, opened 2021-09-29
Return null instead of throwing on missing key in `Dictionary#get`

### LINKED ISSUE #37: Change `Dictionary#get` to return `null` on missing keys instead of throwing an exception
That'd be more aligned with what other collection types do, including `Map` and `Set`.

tschneidereit:
This only returns `null` when the key is missing and still throws when the dictionary itself isn't found. We should catch the latter in `new Dictionary()`, but that's for another patch.

Fixes #37


tschneidereit:
I'll merge this despite the `cargo audit` failure: we know what the new wasmtime release is about, and that the security issue it fixes doesn't affect our use here.

PR #44, opened 2021-10-14
Report uncaught exceptions in the request handler to stderr

tschneidereit:
More precisely, this will log the error message and stack if the promise passed to `FetchEvent#respondWith` is rejected.

One common scenario in which this happens is when passing the return value of an async function to `respondWith`, and that async function then throws an uncaught exception. E.g., this previously failed silently (besides sending a response with a `500` http status and an empty body):

```js
addEventListener("fetch", event => event.respondWith(handleRequest(event)));

  throw new Error("The developer should really see this");
}
```

With this change, it will log the following to `stderr`:
```
Error while running request handler: The developer should really see this
Stack:
  handleRequest@<stdin>:4:9
  @<stdin>:1:54
```


JakeChampion (approved):
I've definitely hit those silent errors many times, this is a great addition!

PR #45, opened 2021-10-17
Update to SpiderMonkey 94 beta

tschneidereit:


tschneidereit:
A review here could only be a rubberstamp, since we're not going to review the many many changes made to SpiderMonkey itself, so I'm going ahead with merging this without that rubberstamp.

PR #51, opened 2021-11-10
Update changelog and Cargo.toml to version 0.2.2

tschneidereit:
Note: I had previously published version 0.3.0, but given that there are no breaking changes here, and that the previously published version was somewhat broken, I'm retracting that and only bumping the minor version here instead.


tschneidereit:
I'm merging this without review as it's been discussed and there aren't any functional changes

PR #60, opened 2022-01-14
Loose set of small fixes and improvements

tschneidereit:
These are some small changes I amassed in working on TransformStream that were easy to separate out, so I moved them into a separate pull request for easier reviewing.

PR #61, opened 2022-01-14
Implement TransformStream

tschneidereit:
SpiderMonkey doesn't have a `TransformStream` implementation, and doing a proper one that'd work for Firefox as well would be a huge amount of work. So instead, we're adding our own implementation, using SpiderMonkey's `ReadableStream` and `WritableStream` implementations.

The goal is for the implementation to be safe, efficient, and spec-compliant, so we're not taking undue shortcuts here. The difference between this and an in-SpiderMonkey implementation is that we don't have to worry about multiple compartments and different kinds of cross-compartment wrappers, which simplifies the implementation significantly.

The implementation passes almost all spec tests (CI support for which will be added in a separate PR), and is I think reasonably efficient.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -754,8 +754,8 @@ namespace NativeStreamSource {
 
   JSObject* create(JSContext* cx, HandleObject owner, HandleValue startPromise,
                    PullAlgorithm* pull, CancelAlgorithm* cancel);
-  static JSObject* get_stream_source(JSObject* stream);
-  bool stream_has_native_source(JSObject* stream);
+  static JSObject* get_stream_source(JSContext* cx, HandleObject stream);
+  bool stream_has_native_source(JSContext* cx, HandleObject stream);
   bool lock_stream(JSContext* cx, HandleObject stream);
 }
 
@@ -820,9 +820,10 @@ namespace RequestOrResponse {
     return val.isNullOrUndefined() ? nullptr : &val.toObject();
   }
 
-  JSObject* body_source(JSObject* obj) {
+  JSObject* body_source(JSContext* cx, HandleObject obj) {
     MOZ_RELEASE_ASSERT(has_body(obj));
-    return NativeStreamSource::get_stream_source(body_stream(obj));
+    RootedObject stream(cx, body_stream(obj));
+    return NativeStreamSource::get_stream_source(cx, stream);
   }
 
   bool mark_body_used(JSContext* cx, HandleObject obj) {
@@ -1035,7 +1036,7 @@ namespace RequestOrResponse {
     // This assert guarantees that we fail early in case SpiderMonkey ever changes how
     // the underlying source for streams is stored, instead of operating on invalid
     // values.
-    MOZ_RELEASE_ASSERT(NativeStreamSource::get_stream_source(body_stream) == source);
+    MOZ_RELEASE_ASSERT(NativeStreamSource::get_stream_source(cx, body_stream) == source);
 
     JS_SetReservedSlot(owner, RequestOrResponse::Slots::BodyStream, JS::ObjectValue(*body_stream));
     return body_stream;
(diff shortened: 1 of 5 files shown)


tschneidereit:
A couple of notes on reviewing this: it builds on #60, so only cb0d778 is relevant here. Also, I validated spec correctness using the WPT harness (which I'll add in a separate PR) so the key things to focus on are implementation quality, memory and type safety, etc.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1210,6 +1212,1279 @@ namespace NativeStreamSource {
...
+    RootedObject readableController(cx, JS::ReadableStreamGetController(cx, readable));
+    MOZ_ASSERT(readableController);
+
+    // 3.  Perform ! [ReadableStreamDefaultControllerClose](readableController).
+    // Note: in https://github.com/whatwg/streams/pull/1029, the spec was changed to
+    // make ReadableStreamDefaultControllerClose (and -Enqueue) return early if
+    // ReadableStreamDefaultControllerCanCloseOrEnqueue is false. SpiderMonkey hasn't been
+    // updated accordingly, so it'll throw an exception instead.
+    // To avoid that, we do the check explicitly. While that also throws an exception, we can
+    // just clear it and move on.
```
cfallin, line 1640: Comment here on what happens when SpiderMonkey is updated? I guess then we just have a no-op `ClearPendingException` (which is harmless)?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1210,6 +1212,1279 @@ namespace NativeStreamSource {
...
+      JS_ClearPendingException(cx);
+    }
+
+    // 4.  Let error be a `[TypeError](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)` exception indicating that the stream has been terminated.
+    RootedValue error(cx);
+    JS_ReportErrorLatin1(cx, "The TransformStream has been terminated");
+    if (!JS_GetPendingException(cx, &error)) {
+      return false;
+    }
+    JS_ClearPendingException(cx);
```
cfallin, line 1655: This sequence is somewhat confusing at least to me: we report an error (which creates an exception on the context) then clear it? Perhaps a comment here to describe intent would help...

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1210,6 +1212,1279 @@ namespace NativeStreamSource {
...
+    }
+
+    // 5.1.2.  Perform ! [ReadableStreamDefaultControllerClose](readable.[controller]).
+    RootedObject readableController(cx, JS::ReadableStreamGetController(cx, readable));
+    // Note: in https://github.com/whatwg/streams/pull/1029, the spec was changed to
+    // make ReadableStreamDefaultControllerClose (and -Enqueue) return early if
+    // ReadableStreamDefaultControllerCanCloseOrEnqueue is false. SpiderMonkey hasn't been
+    // updated accordingly, so it'll throw an exception instead.
+    // To avoid that, we do the check explicitly. While that also throws an exception, we can
+    // just clear it and move on.
```
cfallin, line 2233: Same comment here as above: we should note what will happen when SM is updated, so that this code is future-proof.

cfallin (approved):
This looks overall reasonable, thanks!

I'm leaning pretty heavily on "passes the test suite" for confidence as I don't claim to understand much about the Streams APIs; I mostly reviewed for JSAPI idioms and made sure that the code seems to correspond with the quoted steps from the spec in the comments (this makes it very easy to verify, thanks!).

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: I added the following comment:
```c
    // Note that this is future-proof: if SpiderMonkey is updated accordingly, it'll simply
    // stop throwing an exception, and the `else` branch will never be taken.
```

PR #62, opened 2022-01-14
Support directly piping entire body streams to Request or Response objects instead of copying them by the chunk

tschneidereit:
This PR adds support for detecting when a Request or Response's body stream is piped to another Request or Response's body via a TransformStream. In that case, instead of reading every chunk from the body individually and then writing it to the destination body, we use the `xqd_body_append` hostcall to append the entire body in one go.

In situations where multiple bodies are piped to the same destination, this reduces the complexity from O(number of streams * number of chunks per stream) to O(number of streams), which in most cases basically takes JS execution out of the equation.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -307,8 +307,8 @@ static const uint32_t class_flags = 0;
     return val.isObject() && is_instance(&val.toObject()); \
   } \
  \
-  bool check_receiver(JSContext* cx, HandleObject self, const char* method_name) { \
-    if (!is_instance(self)) { \
+  bool check_receiver(JSContext* cx, HandleValue receiver, const char* method_name) { \
+    if (!is_instance(receiver)) { \
       JS_ReportErrorUTF8(cx, "Method %s called on receiver that's not an instance of %s\n", \
                          method_name, class_.name); \
       return false; \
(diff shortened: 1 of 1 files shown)


tschneidereit:
A note on reviewing: this builds on #61, so only the last three patches are relevant. They can be reviewed independently if desired.

AndrewBarba:
This is huge. Nice work

AndrewBarba:
@tschneidereit Is there anyway to set a response content-length when the content-length of each appended stream is already known? CF has the concept of a `FixedLengthStream` which is identical to an identity TransformStream except constructor takes in the known content-length so the host can correctly return it.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -997,9 +1047,30 @@ namespace RequestOrResponse {
     return true;
   }
 
-  bool body_source_pull_algorithm(JSContext* cx, CallArgs args, HandleObject stream,
-                                  HandleObject owner, HandleObject controller)
+  bool body_source_pull_algorithm(JSContext* cx, CallArgs args, HandleObject source,
+                                  HandleObject body_owner, HandleObject controller)
   {
+    // If the stream has been piped to a TransformStream whose readable end was then passed
```
cfallin, line 1053: Is there any sort of reordering issue that could occur here, because of the multi-step discovery of the shortcut? What I mean is that the flow seems to be (i) pipe the stream to the body output, setting this property under the hood; (ii) later, the "pull algorithm" discovers this and does the direct-path optimization. But what if multiple streams are piped to the output, hitting step (i) for multiple sources, and then the pulls happen in some other arbitrary order?

Perhaps this doesn't happen either because the pull is performed right away, or the output has some mutual exclusion; I don't know enough about the overall APIs to say. But a comment somewhere describing how order of multiple appends is preserved would be helpful :-)

cfallin (approved):
Looks good! One request for clarification in a comment below, but otherwise, 👍

tschneidereit:
@AndrewBarba thank you so much for the kind words! <3

We're evaluating that part separately, and I'll follow up once we have an answer fully sorted out for that part.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: Great question! That is indeed ruled out by how piping works: `pipeTo` locks the target stream, and it'll only unlock once the input stream closes, at which point you can call `pipeTo` with the next stream. Does that address your concern? If so, I'll add that to the comments somewhere.

cfallin: Ah, yes, that makes perfect sense. A note somewhere in the source making that connection would be great, thanks!

PR #64, opened 2022-01-28
Significantly improve spec-compliance of Request and Response objects

tschneidereit:
This pair of commits contains a ton of small changes to `Request` and `Response` objects to make them more spec compliant, with key aspects being
- spec-compliant constructors, enabling proper application of properties on the `input` and `init` arguments
- support for `URLSearchParams` as bodies
- support for content-provided bodies in `{Request,Response}.{text,json,arrayBuffer}()`

The diff is pretty big, but many of the added lines are just comments documenting the spec steps implemented, so I hope it won't be too annoying to review.

PR #65, opened 2022-01-28
Set of added or improved smaller builtin objects/functions

tschneidereit:
This PR contains a smattering of smaller builtins, and some improvements to the behavior of existing ones. They're small enough in sum that I didn't feel it makes sense to split them into multiple PRs, but let me know if you disagree.

Stacked on #64, so the first two commits don't need reviewing.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -482,7 +482,7 @@ namespace Env {
   }
 
   const JSFunctionSpec methods[] = {
-    JS_FN("get", env_get, 1, 0),
+    JS_FN("get", env_get, 1, JSPROP_ENUMERATE),
   JS_FS_END};
 
   JSObject* create(JSContext* cx) {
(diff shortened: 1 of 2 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6647,66 +6668,310 @@ namespace URLSearchParams {
...
+
+    if (!JS_DefineProperty(cx, global, "self", args[0], JSPROP_ENUMERATE)) {
+      return false;
+    }
+
+    args.rval().setUndefined();
+    return true;
+  }
+
+  #define SCTAG_DOM_URLSEARCHPARAMS 0xffff8014 // Same tag as used by Gecko.
```
cfallin, line 6871: A slightly longer comment here might help understanding -- something like "Magic number used in structured cloning as a tag to identify a URLSearchParam"?

Also, regarding use of the same tag as in Gecko -- could you add a comment why? Is it just a convenience to use the same tag-value assignments (and we could choose any value but we arbitrarily choose this one), or is there some expectation that the serialized form could be exposed someday and the tag/format matters?

cfallin (approved):
LGTM! Small question/request for clarification in the structured-clone impl but otherwise seems fine.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: Both very good points! :) There is absolutely no technical reason to reuse Gecko's value here—I just didn't want to think about it 😬 But cargo-culting without explanation is annoying, so I changed this to the first available tag, conveniently identified by `JS_SCTAG_USER_MIN`, and added the comment you suggested.

PR #67, opened 2022-02-08
Support streaming upstream request bodies

tschneidereit:
Until this PR, we only fully supported sending responses downstream with a streaming body. This adds the ability to do the same with upstream requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.2.4 (2022-02-09)
+
+### Enhancements
+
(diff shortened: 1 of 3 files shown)


tschneidereit:
@cfallin most of the diff here is code motion, which I split off into the first of the two patches. it might make sense to only really look at the second patch with the much smaller diff.

cfallin (approved):
This all looks reasonable to me -- I don't claim to grok every detail but the JSAPI-level details look right and the overall flow looks reasonable!

PR #71, opened 2022-03-10
increase max uri length to 8k to match the limit documented on https://docs.fastly.com/en/guides/resource-limits

JakeChampion:
At time of writing, the maximum uri allowed on fastly is 8kb -- https://docs.fastly.com/en/guides/resource-limits

I came across this issue when replaying production traffic from https://poyfill.io onto our js compute-at-edge implementation, which is currently running on https://qa.polyfill.io


tschneidereit (approved):
Makes sense, thank you! <3

PR #73, opened 2022-03-21
Change error message which is given when constructing a Response instance with a body and a null body status

JakeChampion:
previously this error message said "Request constructor", which was misleading as this error is within the Response constructor.

I've used the same error message text as Firefox does.

We could modify the error message some more to mention which status codes are 'null body' statuses, as a null body status is a status that is 101, 204, 205, or 304, and the user reading the error message may not know what is meant by `null body status`.


tschneidereit (approved):
Oops, great catch—thank you! <3

JakeChampion:
@tschneidereit I rebased this and now it all passes :tada:

tschneidereit (approved):
Thank you!

PR #74, opened 2022-03-23
Move the JS SDK into this JS Compute Runtime Repo (SDK and Integration Tests)

torch2424:
**Note: @tschneidereit I'll need access to the repo secrets (Possibly admin access to the repo?), so the tests and Github Action I am bringing over can run**.

As requested, I went ahead and brought over our SDK into our JS runtime repo, that way we can simplify the development of the SDK and the runtime.

This also includes our integration_tests, that uses our soon-to-be-public CI solution, that's built on top of Viceroy and C@E 😄

**Also, we'll need some follow up PRs for the following things:** 

* Reference Docs
* Auto deploy on Github Releases
* Having the JS SDK point properly to our build artifacts, now that we don't need to move them into the SDK manually anymore

P.S It's also worth noting, I didn't moved over the `streaming_response` test, because it's functionality is already covered by the `streaming_close` test.


torch2424:
@tschneidereit The GIthub Action is now passing! 😄 🎉 

I don't know what the cargo warnings are for, but all the JS SDK stuff is all good to go here! 😄

elliottt:
Closing in favor of #99.

PR #84, opened 2022-05-02
Implement CompressionStream builtin

tschneidereit:
This adds an implementation of and test expectations for the [CompressionStream](https://wicg.github.io/compression/#compression-stream) builtin allowing for compression of buffers or streams using the gzip or deflate compression formats.

Some of the changes involve how streams interact with host-side bodies, and fix some issues uncovered through tests of compression streams, that didn't otherwise show up.


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3166,6 +3235,315 @@ namespace TransformStream {
...
+        uint8_t* out_buffer = JS_GetUint8ArrayData(out_obj, &is_shared, nogc);
+        memcpy(out_buffer, buffer, bytes);
+        RootedValue out_chunk(cx, ObjectValue(*out_obj));
+        if (!TransformStreamDefaultController::Enqueue(cx, controller, out_chunk)) {
+          return false;
+        }
+      }
+
+      // 3.  If _buffer_ is empty, return.
+    } while (zstream->avail_out == 0);
```
cfallin, line 3400: I'm not sure I understand the loop condition `zstream->avail_out == 0` here: does this mean that we keep invoking `deflate` as long as it *fully* fills the output chunk? Is it possible for `deflate` to fill, say, all but a few bytes of the output (maybe there isn't enough space to fully encode a symbol) and then return while there is still more available input?

Said another way, what prevents the loop condition from being `while (zstream->avail_in > 0)` ? Or `while ((bytes > 0) || (zstream->avail_in > 0))` (i.e., as long as either there is more input to process, or this invocation produced some output)?

cfallin (commented):
Looks generally good, just one uncertainty below related to the compression loop:

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
tschneidereit: Very good question! The trivial answer to "why did I do it this way" is that that's how Gecko and [Chromium](https://chromium.googlesource.com/chromium/src/+/e8564ac44910253b6daada89fd67405c3a886f57/third_party/blink/renderer/modules/compression/deflate_transformer.cc#99) (in this case for exactly the same purpose) use zlib.

I added a comment explaining this and referring to the zlib docs which [explicitly state](https://searchfox.org/mozilla-central/rev/87ecd21d3ca517f8d90e49b32bf042a754ed8f18/modules/zlib/src/zlib.h#319-324) that this is what one should do.

cfallin (approved):
Sounds good, thanks for the clarifications!

PR #86, opened 2022-05-20
Throw error if specific constructors are called without `new`

JakeChampion:
The below constructors will now throw an error if called without `new` as per their specifications:
- CompressionStream
- Headers
- Request
- Response
- TextDecoder
- TextEncoder
- TransformStream
- URL
- URLSearchParams

I wasn't sure what we want to do with the Fastly specific constructors such as CacheOverride so I have left them as-is

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -182,6 +182,15 @@ uint8_t* value_to_buffer(JSContext* cx, HandleValue val, const char* val_desc, s
   return data;
 }
 
+inline bool ThrowIfNotConstructing(JSContext* cx, const CallArgs& args,
+                                   const char* builtinName) {
+  if (args.isConstructing()) {
+    return true;
+  }
+  JS_ReportErrorASCII(cx, "Constructor %s requires 'new'", builtinName);
+  return false;
+}
+
 /* Returns false if an exception is set on `cx` and the caller should immediately
    return to propagate the exception. */
 static inline bool handle_fastly_result(JSContext* cx, int result, int line, const char* func) {
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -2566,6 +2566,9 @@ namespace TransformStream {
    */
   bool constructor(JSContext* cx, unsigned argc, Value* vp) {
     CallArgs args = CallArgsFromVp(argc, vp);
+    if (!ThrowIfNotConstructing(cx, args, "TransformStream")) {
+      return false;
+    }
 
     RootedObject startFunction(cx);
     RootedObject transformFunction(cx);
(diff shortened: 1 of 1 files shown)


JakeChampion:
I thought this would increase the amount of passed tests from web-platform-tests but it looks like this part of the WebIDL specification wasn't being fully tested within web-platform-tests.

I've made a [pull-request to web-platform-tests](https://github.com/web-platform-tests/wpt/pull/34157) which tests this part, if/when that gets merged I can update our version of web-platform-tests and then see how many more tests we pass :+1:

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -2557,6 +2566,9 @@ namespace TransformStream {
    */
   bool constructor(JSContext* cx, unsigned argc, Value* vp) {
     CallArgs args = CallArgsFromVp(argc, vp);
+    if (!ThrowIfNotConstructing(cx, args, "TransformStream")) {
+      return false;
+    }
```
tschneidereit, line 2571: With my `CTOR_HEADER` suggestion, this would become
```suggestion
    CTOR_HEADER("TransformStream", 0)
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3300,6 +3312,9 @@ namespace CompressionStream {
   bool constructor(JSContext* cx, unsigned argc, Value* vp) {
     // 1.  If _format_ is unsupported in `CompressionStream`, then throw a `TypeError`.
     CallArgs args = CallArgsFromVp(argc, vp);
+    if (!ThrowIfNotConstructing(cx, args, "CompressionStream")) {
+      return false;
+    }
     if (!args.requireAtLeast(cx, "CompressionStream constructor", 1)) {
       return false;
     }
```
tschneidereit, line 3320: ... and this would become
```suggestion
    CTOR_HEADER("CompressionStream", 1)
```

tschneidereit (changes requested):
This is great, thank you!

I have two notes that should be fairly simple to address before landing this:
1. I think we should apply this to Fastly-specific constructors as well
2. how would you feel about introducing a `CTOR_HEADER` macro, parallel to the `METHOD_HEADER` one, and using that instead?

The obvious difference is that we can't infer the name to use from `__func__`, but that seems ok to me.

JakeChampion:
> This is great, thank you!
> 
> I have two notes that should be fairly simple to address before landing this:
> 
>     1. I think we should apply this to Fastly-specific constructors as well
> 
>     2. how would you feel about introducing a `CTOR_HEADER` macro, parallel to the `METHOD_HEADER` one, and using that instead?
> 
> 
> The obvious difference is that we can't infer the name to use from `__func__`, but that seems ok to me.

Oh that is a good suggestion! :relaxed: 
I've made those changes and pushed it up now :+1:

tschneidereit (approved):
Thank you!

tschneidereit:
Oh, turns out `URLSearchParams` doesn't require any arguments, so we have a bunch of test failures :)

JakeChampion:
> Oh, turns out `URLSearchParams` doesn't require any arguments, so we have a bunch of test failures :)

Hooray for tests! :dancers:

PR #91, opened 2022-05-31
Add support for generating a compile_commands.json for c-dependencies/js-compute-runtime

elliottt:
Add a new `compile_commands.json` target in `c-dependencies/js-compute-runtime/Makefile` for generating  the build config that `clangd` requires. Typing `make compile_commands.json` will generate the file, and enable the use of `clangd` when working on `js-compute-runtime`.

Also make some small changes to the `Makefile`:

* Replace uses of `=` with `:=` when possible to avoid accidentally introducing macros
* Add `clean` and `distclean` targets
* Add entries for generated intermediates to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,13 @@
 /target
 /dist
 .DS_Store
+
+# clangd state
+c-dependencies/js-compute-runtime/compile_commands.json
+c-dependencies/js-compute-runtime/.cache/
+
+c-dependencies/js-compute-runtime/rusturl/
+c-dependencies/js-compute-runtime/compiler_flags
+c-dependencies/js-compute-runtime/*.o
+c-dependencies/js-compute-runtime/*.d
+c-dependencies/js-compute-runtime/*.wasm
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,3 @@
 /target
 /dist
 .DS_Store
-
-# clangd state
-c-dependencies/js-compute-runtime/compile_commands.json
-c-dependencies/js-compute-runtime/.cache/
-
-c-dependencies/js-compute-runtime/rusturl/
-c-dependencies/js-compute-runtime/compiler_flags
-c-dependencies/js-compute-runtime/*.o
-c-dependencies/js-compute-runtime/*.d
-c-dependencies/js-compute-runtime/*.wasm
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -58,8 +58,8 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
 	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) -I $(SM_SRC)include -MMD -MP -c -o $@ $<
 
 js-compute-runtime.wasm: $(FSM_OBJ) $(SM_OBJ) $(RUST_URL_LIB)
-	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) $(LD_FLAGS) -o $@ $^ && \
-		$(call WASM_STRIP,$@)
+	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) $(LD_FLAGS) -o $@ $^
+	$(call WASM_STRIP,$@)
 
 initialized-js-compute-runtime.wasm: js-compute-runtime.wasm $(INIT_JS)
 	cat $(INIT_JS) | $(WIZER) --allow-wasi --dir=. -r _start=wizer.resume -o $@ $<
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -70,8 +70,8 @@ clean:
 distclean: clean
 	$(RM) $(FSM_DEP) compiler_flags
 
-# technically this depends on the value of the FSM_CPP variable
-compile_commands.json: compiler_flags
+.PHONY: compile_commands.json
+compile_commands.json:
 	$Q ( \
 		sep="["; \
 		for file in $(FSM_CPP); do \
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -9,8 +9,8 @@ ifdef DEBUG
   CXX_OPT := $(CXX_OPT) -DDEBUG -DJS_DEBUG
   Q :=
 else
-  MODE=release
-  CARGO_FLAG=--release
+  MODE := release
+  CARGO_FLAG := --release
   Q := @
 endif
 
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -34,7 +34,6 @@ all: js-compute-runtime.wasm
 compiler_flags:
 	echo '$(CXX_OPT) $(CXX_FLAGS)' | cmp -s - $@ || echo '$(CXX_OPT) $(CXX_FLAGS)' > $@
 
-WASM_STRIP = true
 ifeq (,$(findstring g,$(CXX_OPT)))
 ifneq (,$(shell which wasm-opt))
 WASM_STRIP = wasm-opt --strip-debug -o $1 $1
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -3,36 +3,41 @@ WIZER ?= wizer
 
 CXX_OPT ?= -O2
 
-MODE=release
-CARGO_FLAG=--release
 ifdef DEBUG
-  MODE=debug
-  CARGO_FLAG=
+  MODE := debug
+  CARGO_FLAG :=
   CXX_OPT := $(CXX_OPT) -DDEBUG -DJS_DEBUG
+  Q :=
+else
+  MODE=release
+  CARGO_FLAG=--release
+  Q := @
 endif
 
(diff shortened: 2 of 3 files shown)


c-dependencies/js-compute-runtime/Makefile
```diff
@@ -53,8 +58,33 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
 	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) -I $(SM_SRC)include -MMD -MP -c -o $@ $<
 
 js-compute-runtime.wasm: $(FSM_OBJ) $(SM_OBJ) $(RUST_URL_LIB)
-	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) $(LD_FLAGS) -o $@ $^
-	$(WASM_STRIP)
+	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) $(LD_FLAGS) -o $@ $^ && \
+		$(call WASM_STRIP,$@)
```
jameysharp, line 62: I haven't touched makefiles in a while, but the [GNU make manual](https://www.gnu.org/software/make/manual/html_node/Errors.html) says execution within a rule stops at the first line that exits with a non-zero status. So combining two lines with `&&` is unnecessary and I think makes maintenance a little harder.

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -3,36 +3,41 @@ WIZER ?= wizer
...
 
 compiler_flags:
-	echo '$(CXX_OPT)' | cmp -s - $@ || echo '$(CXX_OPT)' > $@
+	echo '$(CXX_OPT) $(CXX_FLAGS)' | cmp -s - $@ || echo '$(CXX_OPT) $(CXX_FLAGS)' > $@
 
+WASM_STRIP = true
 ifeq (,$(findstring g,$(CXX_OPT)))
 ifneq (,$(shell which wasm-opt))
-WASM_STRIP = wasm-opt --strip-debug -o $@ $@
+WASM_STRIP = wasm-opt --strip-debug -o $1 $1
```
jameysharp, line 40: Does `$1` work here? I see you've changed the usage of this variable to a `$(call)`, but given that my reading of the GNU make manual ([call function](https://www.gnu.org/software/make/manual/html_node/Call-Function.html), [variable references](https://www.gnu.org/software/make/manual/html_node/Reference.html)) is that you'd need to use `$(1)` rather than `$1` here.

I'm not sure either change is necessary anyway: `$@` means the target of the current rule, which should be the right thing here, shouldn't it?

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -53,8 +58,33 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
...
 initialized-js-compute-runtime.wasm: js-compute-runtime.wasm $(INIT_JS)
 	cat $(INIT_JS) | $(WIZER) --allow-wasi --dir=. -r _start=wizer.resume -o $@ $<
+
+clean:
+	$(RM) compile_commands.json $(FSM_OBJ)
+
+distclean: clean
+	$(RM) $(FSM_DEP) compiler_flags
+
+# technically this depends on the value of the FSM_CPP variable
```
jameysharp, line 73: I think you could safely add `$(FSM_CPP)` to the dependencies of this rule to express what this comment says.

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -53,8 +58,33 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
...
+		for file in $(FSM_CPP); do \
+			echo "$$sep"; \
+			sep=","; \
+			echo "{ \"directory\": \"$(FSM_SRC)\","; \
+			echo "  \"command\": \"$(WASI_CXX) $(CXX_FLAGS) $(DEFINES) -I $(SM_SRC)include\","; \
+			echo -n "  \"file\": \"$$(basename $$file)\"}"; \
+		done; \
+		echo; \
+		echo ']' \
+	) > "$@"
```
jameysharp, line 86: All I know about `compile_commands.json` is that when I tried to generate one once I got confused and gave up. Is this the best way to do it? I thought I remembered clang having an option to generate it for you, maybe?

jameysharp (changes requested):
Lots of good improvements here! I'm not immediately convinced that the changes to `WASM_STRIP` are correct, but all my other notes are quibbles you can feel free to ignore.

c-dependencies/js-compute-runtime/Makefile
elliottt: For variables with only a single character name, it's fine to use them without parens. This makefile behaves the same for calls to both `ECHO` and `ECHO2`:

```make
ECHO = echo $1                                                                       
ECHO2 = echo $(1)                                                                    
                                                                                     
foo: ci/*.sh                                                                         
        $(call ECHO,$<)                                                              
        $(call ECHO2,$<) 
```

I was a little bit uneasy leaving it as `$@` because it makes assumptions about where the `WASM_STRIP` macro will be  used (always in a rule context), while being explicit about it being a macro and using `$1` felt like it encoded the intent a little better.

I sort of wish that make didn't allow macros to be introduced with `=` :(

c-dependencies/js-compute-runtime/Makefile
elliottt: This avoids the overhead of invoking the shell once per line, but that's probably not really something worth worrying about here. I'll switch it back because I agree that it's more readable on a separate line :+1: 

There is the `.ONESHELL` directive, but that brings back the opportunity for a command to fail and not halt rule execution.

c-dependencies/js-compute-runtime/Makefile
elliottt: I didn't want it to depend on the files, only the state of that variable. I could pipe them into the `compiler_flags` file as well though...

c-dependencies/js-compute-runtime/Makefile
elliottt: If clang can do it automatically, that would be better. This version doesn't feel ideal, but does produce a working compile_commands.json :)

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -3,36 +3,41 @@ WIZER ?= wizer
...
 ifdef DEBUG
-  MODE=debug
-  CARGO_FLAG=
+  MODE := debug
+  CARGO_FLAG :=
   CXX_OPT := $(CXX_OPT) -DDEBUG -DJS_DEBUG
+  Q :=
+else
+  MODE=release
+  CARGO_FLAG=--release
```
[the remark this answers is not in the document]
elliottt: Good catch, they definitely should be :)

.gitignore
```diff
@@ -1,3 +1,13 @@
...
+
+# clangd state
+c-dependencies/js-compute-runtime/compile_commands.json
+c-dependencies/js-compute-runtime/.cache/
+
+c-dependencies/js-compute-runtime/rusturl/
+c-dependencies/js-compute-runtime/compiler_flags
+c-dependencies/js-compute-runtime/*.o
+c-dependencies/js-compute-runtime/*.d
+c-dependencies/js-compute-runtime/*.wasm
```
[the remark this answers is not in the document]
elliottt: I like that too, but have seen preferences elsewhere for one global `.gitignore`. Why don't we move these into `c-dependencies/js-compute-runtime/.gitignore` and we can migrate rules like `*.o` out if we start accumulating more c++ source outside of that tree?

c-dependencies/js-compute-runtime/Makefile
jameysharp: Oh right, I get it. Well, maybe the rule runs fast enough that you could mark it as `.PHONY` and not worry about its dependencies?

c-dependencies/js-compute-runtime/Makefile
elliottt: A quick googling didn't turn up anything that automatically generates compile_commands.json, and my experience with compile_commands.json from bazel and cmake is that it's generated by the build system, not clang/clangd. How do you feel about using this hacky version for now, and replacing it with something more robust when we find it?

c-dependencies/js-compute-runtime/Makefile
elliottt: Entirely reasonable! I'll make that change.

c-dependencies/js-compute-runtime/Makefile
jameysharp: Ah, I missed the last paragraph of the variable references section. Got it.

I wonder, could the conditional be shoved into the rule directly in a way that's more clear? I dunno, maybe like this:

```sh
	test -z "$(findstring g,$(CXX_OPT))" && which wasm-opt >/dev/null && wasm-opt --strip-debug -o $@ $@
```

I'm not convinced that looking for the single letter "g" makes sense here either, but maybe it's best to leave that alone. I suppose it works if you pass either `-Og` or `-g`?

c-dependencies/js-compute-runtime/Makefile
jameysharp: _If_ we find it. It's entirely possible I gave up the last time because there isn't a nice option if your build system doesn't do it for you; I don't remember. Yeah, I'm satisfied, ship it!

Maybe we should discuss changing the build system here to something more capable. That's certainly not blocking this PR though.

jameysharp (approved):

c-dependencies/js-compute-runtime/Makefile
elliottt: It'll definitely for unrelated occurrences of `g` in `CXX_OPT`. We could search for explicitly `-g`, but that will still match substrings.

PR #93, opened 2022-06-01
Add a format checking job and clang-format config

elliottt:
Add a `.clang-format` config in the project root with mostly default settings (line length is set to 100 instead of 80). Also add a ci job to run `ci/clang-format.sh` on all tracked c/c++ source.


ci/clang-format.sh
```diff
@@ -0,0 +1,68 @@
...
+
+failure=
+for file in $(git ls-files | grep '\.\(cpp\|h\)$'); do
+  if grep -F -x "$file" ci/clang-format.ignore > /dev/null; then
+    continue
+  fi
+
+  formatted="${file}.formatted"
+  clang-format "$file" > "$formatted"
+  if ! diff "$file" "$formatted" > /dev/null; then
```
jameysharp, line 52: Can I suggest `if ! cmp -s "$file" "$formatted"; then` instead? It should be faster than `diff` since you don't care what the differences are, only if there are any. Of note, the `-s` flag to `cmp` is standardized in POSIX, but the `--quiet` and `--silent` synonyms are GNU extensions.

jameysharp (changes requested):
I'm sad that this changes so much code, but checking and canonicalizing formatting is so valuable that it's worth doing anyway. I spent a little while trying to see if there were settings that might reduce the amount of code churn here, but the source tree wasn't really internally consistent so nevermind.

I do think it's important to not touch `wizer.h` though since that comes from another project. `xqd.h` might be in the same situation, but I haven't yet found where the original might be if there is one.

ci/clang-format.sh
elliottt: So much better, thank you!

ci/clang-format.ignore
```diff
@@ -0,0 +1 @@
+c-dependencies/js-compute-runtime/rust-url/rust-url.h
```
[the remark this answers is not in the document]
elliottt: I've excluded it and `xqd.h`, thanks for catching this!

jameysharp (approved):

PR #95, opened 2022-06-09
Do not include debug information when compiling rust-url in release mode

JakeChampion:
Currently we are including the full debug information inside the binary.
This patch changes the release profile to use the default debug value of `false` which will include no debug information inside the binary.

References:
- The default release profile values: https://doc.rust-lang.org/cargo/reference/profiles.html#release
- What the debug value is used for: https://doc.rust-lang.org/cargo/reference/profiles.html#debug


elliottt (approved):
This seems very reasonable to me. Given that this was explicitly enabled before, do we have a strategy for getting a debug build to someone that's debugging runtime issues?

PR #96, opened 2022-06-09
Fix `make compile_commands.json` and add to vscode project settings

JakeChampion:
Because no specific shell was set in the Makefile it defaults to /bin/sh which has a built-in version of echo that does not accept the -n option and instead will write -n to stdout.

Setting the shell to /bin/bash means we will use bash and bashs's built-n echo which does support the -n option, which means we will start generating a valid JSON file for compile_commands.json


elliottt (approved):
Awesome, thanks for fixing this!

PR #98, opened 2022-06-09
Inline the spidermonkey-was-embedding repo

elliottt:
Remove the git submodule for https://github.com/tschneidereit/spidermonkey-wasi-embedding by inlining the content of that repository in `c-dependencies/spidermonkey`. This lets us modify the scripts more easily, and decrements the number of submodules by 1. As we still pull builds from that repo when building locally, a good next step would be to cache the builds on this repo as well, and pull down the spidermonkey build from the most recent commit to `main`.


elliottt:
I'm going to rework this PR into one that's a little easier to review.

PR #100, opened 2022-06-15
Cache crate dependencies on branches

elliottt:
To speed up builds, cache compiled dependencies in the build step using https://github.com/Swatinem/rust-cache. Always rebuild on the `main` branch and for release tags.

As a follow-on from this, we'll need to change our `binary-compatible-builds` action to expose the `~/.cargo` directory outside of the centos container it runs in, as that directory currently disappears with the centos container that runs the linux build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -194,6 +194,10 @@ jobs:
       with:
         name: engine-release
 
+    - name: Cache build
+      uses: Swatinem/rust-cache@v1.4.0
+      if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/v')
+
     # Build `js-compute-runtime`
     - run: PREBUILT_ENGINE=engine-release/js-compute-runtime.wasm $CENTOS cargo build --release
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -194,6 +194,10 @@ jobs:
       with:
         name: engine-release
 
+    - name: Cache build
+      uses: Swatinem/rust-cache@v1.4.0
```
JakeChampion, line 198: Note: Nice! This will be the first time that we use a third-party action in our workflows which makes me think of a few questions about this change
Do we know what our policy is on using third-party actions?
We do have an [.github/actions](https://github.com/fastly/js-compute-runtime/tree/main/.github/actions) folder where we stored our own custom actions, should we be vendoring this third-party action into that folder?
Or maybe we can use the github cache action and configure it for rust, [following the example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo)?

elliottt: These are good points! I think I'll use the github cache action like you mentioned, as that'll avoid any questions about policy or vendoring for now :)

jameysharp (approved):
Looks good to me!

PR #102, opened 2022-06-15
add initial implementation of Blob built-in

JakeChampion:
This is not yet ready for a full review, I've pushed this up early for feedback on my approach and to show what I think we need to do in order to add a new set of web-platform-tests to run against the project


JakeChampion:
These are the commands I use in order to run the web-platform-tests:
```
mkdir dist && cd dist
make -f ../c-dependencies/js-compute-runtime/Makefile && bash ../tests/wpt-harness/build-wpt-runtime.sh && node ../tests/wpt-harness/run-wpt.mjs -vv --update-expectations
```

PR #103, opened 2022-06-15
Bump wizer to 1.4.0

elliottt:
- Bump wizer to 1.4.0
- Don't discard the result in the use of `Wizer::allow_wasi`


jameysharp (approved):
It passes the tests and looks sensible, so, ship it!

PR #107, opened 2022-06-15
Only invoke `docker exec` once during build setup

jameysharp:
This avoids some delay from entering the container repeatedly, and more importantly, only runs `yum` dependency resolution once.


.github/actions/binary-compatible-builds/action.yml
```diff
@@ -2,5 +2,54 @@ name: 'Set up a CentOS 6 container to build releases in'
 description: 'Set up a CentOS 6 container to build releases in'
 
 runs:
-  using: node12
-  main: 'main.js'
+  using: "composite"
+  steps:
```
elliottt, line 6: I love this change!

PR #108, opened 2022-06-16
remove packing ability and requirement for fastly.toml file to exist when using the js-compute-runtime cli

JakeChampion:
The ability to create a tarball was useful before the official fastly cli had the ability to package and deploy c@e projects. As these pieces of functionaliy are now handled by the [fastly cli](https://developer.fastly.com/reference/cli/compute/) itself ([specifically this function is used](https://github.com/fastly/cli/blob/1e3dce3d612766c9f2abeed916f6d49503f3b51d/pkg/commands/compute/build.go#L305-L354))  - we can remove them from the js-compute-cli project.


jameysharp (approved):
Is there documentation somewhere that we need to update for the removal of the `skip-pkg` flag? At the least I guess this change should be mentioned in `CHANGELOG.md`.

Other than that, this looks good to me!

PR #109, opened 2022-06-16
enable running the sdk-tests locally 

JakeChampion:
I wanted to get the tests working locally and I couldn't get the `act` to work correctly, which is what run-local-integration-tests.sh relied on.

I thought we could perhaps make it so that the sdk-tests github action does not rely on anything github action specific and then we can run the action's script ourselves via `node .github/actions/compute-sdk-test/main.js`.

To do this I've chosen to hard-code the test configuration path as we only have one configuration file, this removed the need for the github action specific `@actions/core` npm package. We also had in the code the `@actions/github` package but were never using it, so that has also been removed.

The final change needed to make running the tests locally was to update the `npm run build:test` command to look for the `js-compute-runtime` cli within `./target/release/`.

To run the integration tests locally we can now do this:
`cargo build --release && node .github/actions/compute-sdk-test/main.js`

Note: During the combining of the different repositories we brought in a git pre-commit hook which is running a code-formatter on any modified files, this is why the commits contain lots of format changes.


.github/actions/compute-sdk-test/main.js
```diff
@@ -1,13 +1,12 @@
...
-require('better-logging')(console, {
-  format: ctx => {
-
-    const tag = chalk.bold(`[compute-sdk-test]`); 
+const fs = require("fs");
+const path = require("path");
+const childProcess = require("child_process");
+const fetch = require("node-fetch");
+const chalk = require("chalk");
+require("better-logging")(console, {
```
JakeChampion, line 7: If we can replace these dependencies with nodejs built-in packages such as `http`, then we will no longer need to have a build step nor need to commit the `dist` directory as we could run this file directly with GitHub Actions

elliottt (approved):
This looks great to me, thanks!

Re the formatting pre-commit hook: let's remove the npm dependencies that set that up before merging this pr, so that we can minimize the diff.

.github/actions/compute-sdk-test/main.js
elliottt: I'm all for it!

elliottt: I'm all for it!

PR #110, opened 2022-06-16
Implement ObjectStore and ObjectStoreEntry classes for interacting with Fastly ObjectStore

JakeChampion:
The interfaces for these new classes is as follows:
```ts
declare class ObjectStore {
    constructor(name: string);
    lookup(key: string): Promise<ObjectStoreEntry | null>;
    put(key: string, value: BodyInit): Promise<undefined>;
}

declare class ObjectStoreEntry {
  get body(): ReadableStream;
  get bodyUsed(): boolean;
  text(): Promise<string>;
  json(): Promise<object>;
  arrayBuffer(): Promise<ArrayBuffer>;
  // And eventually formData and blob once we support them on Request and Response, too.
}
```

ObjectStore is a class we expose to application code for interacting with Fastly ObjectStores and ObjectStoreEntry is not exposed to application code. ObjectStoreEntry is only used as the return result from ObjectStore.lookup for successful lookups.

ObjectStoreEntry has an interface very similar to Response, which is intentional as this is an interface JavaScript developers are familar with and it also can be used as a stream via ObjectStoreEntry.body(), which is useful for applications which want to improve their performance and reduce meory usage.

Below is an example js application which uses the new feature:
```js
    event.respondWith(app(event))
})

async function app(event) {
    const store = new ObjectStore('example-store')
    await store.put('hello', 'world')
    let hello = await store.lookup('hello')
    let hellotext = await hello.text()
    return new Response(hellotext)
}
```


JakeChampion:
You can run the integration tests with the below commands:

```bash
cd integration-tests/js-compute/fixtures/object-store
npm ci
node test.js
```

If the last commands has a 0 exit code, then all tests have passed (it will also print all the tests results into stdout)

JakeChampion:
The viceroy tests will fail for now as viceroy does not yet implement the objectstore host calls

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1652,8 +1675,9 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
     // `responseDone`.
     if (Response::is_instance(body_owner)) {
       FetchEvent::set_state(FetchEvent::instance(), FetchEvent::State::responseDone);
+      return HANDLE_RESULT(cx, xqd_body_close(body_handle));
     }
-    return HANDLE_RESULT(cx, xqd_body_close(body_handle));
+    return true;
```
JakeChampion, line 1680:
```suggestion
    return HANDLE_RESULT(cx, xqd_body_close(body_handle));
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
 }
 } // namespace TextDecoder
 
+#ifdef ENABLE_OBJECT_STORE
```
JakeChampion, line 5477: The implementation is currently behind a flag because viceroy does not yet support the required hostcalls. If we compiled with the ObjectStore implementation then when viceroy tries to run the application it will crash due to hostcalls being called which don't exist.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+namespace ObjectStoreEntry {
+namespace Slots {
+enum {
+  Body = RequestOrResponse::Slots::Body,
+  BodyStream = RequestOrResponse::Slots::BodyStream,
+  HasBody = RequestOrResponse::Slots::HasBody,
+  BodyUsed = RequestOrResponse::Slots::BodyUsed,
```
JakeChampion, line 5484: These slots are used in the methods text/json/arraybuffer and the getters body/bodyUsed

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+
+const JSPropertySpec properties[] = {JS_PSG("body", body_get, JSPROP_ENUMERATE),
+                                     JS_PSG("bodyUsed", bodyUsed_get, JSPROP_ENUMERATE), JS_PS_END};
+
+bool constructor(JSContext *cx, unsigned argc, Value *vp) {
+  JS_ReportErrorUTF8(cx, "ObjectStoreEntry can't be instantiated directly");
+  return false;
+}
+
+CLASS_BOILERPLATE(ObjectStoreEntry)
```
JakeChampion, line 5522: The reason we do not use CLASS_BOILERPLATE_NO_CTOR is because that macro will also remove the constructor from the application environment, which makes `entry instanceof ObjectStoreEntry` not be possible because `ObjectStoreEntry` would be undefined

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+  }
+
+  if (strcmp(key_chars, ".") == 0) {
+    JS_ReportErrorASCII(cx, "ObjectStore key can not be \".\"");
+    return false;
+  }
+  if (strcmp(key_chars, "..") == 0) {
+    JS_ReportErrorASCII(cx, "ObjectStore key can not be \"..\"");
+    return false;
+  }
```
JakeChampion, line 5629: These checks exist within the ObjectStore implementation at https://github.com/fastly/object-store
The reason we have them implemented in the JS runtime is to be able to throw a JS Errors that have really nice error message for each scenario instead of the generic "Invalid argument. - Fastly error code 2" message that HANDLE_RESULT would create

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+  // - URLSearchParams
+  // - ReadableStream
+  // After the other other options are checked explicitly, all other inputs are
+  // encoded to a UTF8 string to be treated as a USV string.
+  // TODO: Support the other possible inputs to Body.
+
+  RootedObject body_obj(cx, body_val.isObject() ? &body_val.toObject() : nullptr);
+
+  if (body_obj && JS::IsReadableStream(body_obj)) {
+    JS_ReportErrorLatin1(cx, "Streaming into an ObjectStore is not yet supported");
```
JakeChampion, line 5749: After speaking with @acfoltzer - we think that the hostcalls do support streaming from a BodyHandle which came from the host such as a client request's BodyHandle. Any BodyHandle that came from the guest can not yet be streamed into ObjectStore such as a ReadableStream or a TransformStream.

JakeChampion: There's also an extra challenge here is we wanted it. ObjectStore entries can only be a max of 30MB. We could count the size of the stream as we pass it through to ObjectStore and if it exceeds 30MB we could throw a JS Error which states "ObjectStore value can not be more than 30 Megabytes in size"

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+    } else if (body_obj && URLSearchParams::is_instance(body_obj)) {
+      SpecSlice slice = URLSearchParams::serialize(cx, body_obj);
+      buf = (char *)slice.data;
+      length = slice.len;
+    } else {
+      // Convert into a String following https://tc39.es/ecma262/#sec-tostring
+      text = encode(cx, body_val, &length);
+      // 30MB in bytes is the max size allowed for ObjectStore.
+      if (length > 30 * 1024 * 1024) {
+        JS_ReportErrorLatin1(cx, "ObjectStore value can not be more than 30 Megaytes in size.");
```
JakeChampion, line 5778:
```suggestion
        JS_ReportErrorLatin1(cx, "ObjectStore value can not be more than 30 Megabytes in size.");
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5451,6 +5474,437 @@ bool constructor(JSContext *cx, unsigned argc, Value *vp) {
...
+    if (!HANDLE_RESULT(cx, result)) {
+      return false;
+    }
+
+    size_t inserted = 0;
+    // max-age is ignored for Object Store
+    uint32_t max_age = 0;
+
+    int status = fastly_kv_insert(object_store_handle(self), key_chars, key_len, body_handle,
+                                  max_age, &inserted);
```
JakeChampion, line 5813: This is the now an outdated hostcall, once the new ones are released I can update this pull-request to use the new ones

c-dependencies/xqd.h
```diff
@@ -294,6 +298,18 @@ WASM_IMPORT("fastly_dictionary", "get")
...
+#ifdef ENABLE_OBJECT_STORE
+WASM_IMPORT("fastly_kv", "open")
+int fastly_kv_open(const char *name, size_t name_len, ObjectStoreHandle *object_store_handle_out);
+WASM_IMPORT("fastly_kv", "lookup")
+int fastly_kv_lookup(ObjectStoreHandle object_store_handle, const char *key, size_t key_len,
+                     BodyHandle *opt_body_handle_out);
+WASM_IMPORT("fastly_kv", "insert")
+int fastly_kv_insert(ObjectStoreHandle object_store_handle, const char *key, size_t key_len,
+                     BodyHandle body_handle, uint32_t max_age, size_t *inserted);
+#endif
```
JakeChampion, line 311: These are the now outdated hostcalls, once the new ones are released I can update this pull-request to use the new ones

integration-tests/js-compute/fixtures/extend-from-builtins/extend-from-builtins.ts
```diff
@@ -27,8 +30,9 @@ const builtins = [
               super("gzip")
               break;
             }
-            case "Dictionary": {
-              super("example")
+            case "Dictionary":
+            case "ObjectStore": {
+              super("hello")
```
JakeChampion, line 35: This would actually throw an error due to no ObjectStore named 'hello' existing and being linked to this application. When Viceroy has ObjectStore implemented, we can get this test to pass

integration-tests/js-compute/package.json
```diff
@@ -14,5 +14,8 @@
   },
   "dependencies": {
     "@fastly/js-compute": "file:../../sdk/js-compute"
+  },
+  "volta": {
+    "node": "16.15.1"
```
JakeChampion, line 19: This is useful to me as I use https://volta.sh/ to pick which version of NodeJS and/or npm that each folder/project should use.

integration-tests/js-compute/fixtures/extend-from-builtins/extend-from-builtins.ts
```diff
@@ -1,4 +1,6 @@
 /// <reference types="@fastly/js-compute" />
+declare const CompressionStream: any;
+declare const ObjectStore: any;
```
elliottt, line 3: Thanks for fixing this! It might be worth pulling this out into its own PR just to cut down on the noise during testing.

integration-tests/js-compute/package.json
elliottt: Should we standardize on using this for development?

integration-tests/js-compute/fixtures/object-store/app.js
```diff
@@ -0,0 +1,1242 @@
...
+            return pass()
+        });
+
+        routes.set("/object-store/put/value-parameter-string-over-30mb", async () => {
+            const string = 'x'.repeat(35 * 1024 * 1024) + 'x'
+            const store = createValidStore()
+            let error = assertThrows(() => {
+                let result = store.put('string-over-30mb', string)
+                store.put("test")
+            }, Error, `ObjectStore value can not be more than 30 Megaytes in size.`)
```
JakeChampion, line 388:
```suggestion
            }, Error, `ObjectStore value can not be more than 30 Megabytes in size.`)
```

JakeChampion:
@elliottt this is now ready for a review - the tests won't pass until Viceroy implements the object-store host-calls but the code in this pull-request is now in a stable place for reviewing

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+
+  BodyHandle body_handle = {INVALID_HANDLE};
+  int status =
+      fastly_object_store_lookup(object_store_handle(self), key_chars, key_len, &body_handle);
+  if (!HANDLE_RESULT(cx, status)) {
+    return false;
+  }
+
+  // If the handle is invalid it means no entry was found
+  // When no entry is found, we are going to resolve the Promise with `null`.
```
elliottt, line 180: Should we reject instead?

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+  }
+
+  args.rval().setObject(*result_promise);
+  return true;
+}
+
+bool put(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(2)
+
+  JS::RootedObject result_promise(cx, JS::NewPromiseObject(cx, nullptr));
```
elliottt, line 202: There's a lot of duplicated validation logic between `put` and `lookup`, what do you think about extracting a function do do that part?

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+  } else {
+    mozilla::Maybe<JS::AutoCheckCannotGC> maybeNoGC;
+    JS::UniqueChars text;
+    char *buf;
+    size_t length;
+
+    if (body_obj && JS_IsArrayBufferViewObject(body_obj)) {
+      // Short typed arrays have inline data which can move on GC, so assert
+      // that no GC happens. (Which it doesn't, because we're not allocating
+      // before `buf` goes out of scope.)
```
elliottt, line 330: It might be worth being explicit about `maybeNoGC` needing to be populated for the lifetime of `buf` in this case, otherwise it might be tempting to have it clean itself up at the end of this block.

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+    args.rval().setObject(*result_promise);
+
+    return true;
+  }
+
+  return false;
+}
+
+const JSFunctionSpec methods[] = {JS_FN("lookup", lookup, 1, JSPROP_ENUMERATE),
+                                  JS_FN("put", put, 1, JSPROP_ENUMERATE), JS_FS_END};
```
elliottt, line 398: Why not `get` and `put`?

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+  int status = fastly_object_store_open(name_chars, name_len, &object_store_handle);
+  if (status == 2) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_OBJECT_STORE_DOES_NOT_EXIST,
+                              name_chars);
+    return false;
+  }
+
+  if (!HANDLE_RESULT(cx, fastly_object_store_open(name_chars, name_len, &object_store_handle))) {
+    return false;
+  }
```
elliottt, line 453: Do we need to worry about closing the object store again, or is it just open for the duration of the request?

elliottt (changes requested):
This is looking great, and the test suite is impressive!

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+  auto acme_challenge = ".well-known/acme-challenge/";
+  if (strncmp(key_chars, acme_challenge, strlen(acme_challenge)) == 0) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_OBJECT_STORE_KEY_ACME);
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
+
+  if (strcmp(key_chars, ".") == 0 || strcmp(key_chars, "..") == 0) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_OBJECT_STORE_KEY_RELATIVE);
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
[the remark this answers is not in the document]
JakeChampion: It is meant to check that the entire string is not named `.` or `..`, which would both cause errors from within the object-store service
https://github.com/fastly/object-store/blob/41050556f53e2836dede11869e921ce172215173/api/src/api/key.rs#L78-L83

c-dependencies/js-compute-runtime/builtins/object-store.cpp
JakeChampion: From reading the advice of the W3C TAG (Technical Architecture Group) document on [Writing Promise-Using Specifications](https://www.w3.org/2001/tag/doc/promises-guide#rejections-should-be-exceptional) I thought we should resolve the promise in this situation as the I/O has succeeded but the value was not found, which is a specific example they have about when to resolve and not reject

c-dependencies/js-compute-runtime/builtins/object-store.cpp
JakeChampion: 👍 I like your suggestion as it already exists as the counterpart for `put` in JavaScript `Map` objects - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

c-dependencies/js-compute-runtime/builtins/object-store.cpp
JakeChampion: I believe we do not need to worry about closing the object-store, which is the same approach taken for opening a dictionary. I checked and there are no close methods in the [ABI for object-store](https://github.com/fastly/ExecuteD/blob/main/wasm_abi/rust/fastly/src/object_store/handle.rs) or dictionary which makes me quite confident that we don't require closing them

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+  if (strchr(key_chars, '\n') != NULL) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr,
+                              JSMSG_OBJECT_STORE_KEY_INVALID_CHARACTER, "newline");
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
+  if (strchr(key_chars, '\r') != NULL) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr,
+                              JSMSG_OBJECT_STORE_KEY_INVALID_CHARACTER, "carriage return");
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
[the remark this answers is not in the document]
JakeChampion: nice! I've refactored the logic for validating keys into it's own function now 👍

c-dependencies/js-compute-runtime/builtins/object-store.cpp
JakeChampion: definitely, this commit is the one which has refactored the logic -> [d8328dd83871719549492a7af9920f0914fa28c1](https://github.com/fastly/js-compute-runtime/pull/110/commits/d8328dd83871719549492a7af9920f0914fa28c1)

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -0,0 +1,464 @@
...
+#include "host_call.h"
+#include "js-compute-builtins.h"
+
+namespace ObjectStoreEntry {
+namespace Slots {
+enum {
+  Body = RequestOrResponse::Slots::Body,
+  BodyStream = RequestOrResponse::Slots::BodyStream,
+  HasBody = RequestOrResponse::Slots::HasBody,
+  BodyUsed = RequestOrResponse::Slots::BodyUsed,
```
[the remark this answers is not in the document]
JakeChampion: I'll go with the simpler refactor you suggest of swapping to `namespace Slots = RequestOrResponse::Slots` and open an issue about the larger refactor for RequestOrResponse

JakeChampion:
This is ready for another round of reviews now @elliottt 🙇

elliottt (approved):
This looks good to me!

PR #111, opened 2022-06-16
Remove husky from the integration tests package.json

elliottt:
This was installing pre-commit hooks to run prettier, which is injecting a lot of noise into commits. Let's hold off on git hooks for now, and discuss enforcing js/yaml/etc formatting with a ci job instead.


jameysharp (approved):
Looks good to me!

PR #112, opened 2022-06-16
Cache more in CI

jameysharp:
Here are more experiments in caching various steps that run in GitHub Actions.


.github/actions/binary-compatible-builds/run-linux
```diff
@@ -1,3 +1,8 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
-exec docker exec --env PREBUILT_ENGINE=${PREBUILT_ENGINE:-} -w $PWD -i centos "$@"
+set -euo pipefail
+exec docker run -i \
+  --env PREBUILT_ENGINE=$PREBUILT_ENGINE \
+  --env PATH="/opt/rh/devtoolset-8/root/usr/bin:$PATH" \
+  -v $HOME:$HOME --workdir $PWD \
```
elliottt, line 7:
```suggestion
  -v "$HOME:$HOME" --workdir "$PWD" \
```

.github/workflows/main.yml
```diff
@@ -60,24 +59,24 @@ jobs:
...
+    - name: Cache Rust dependencies
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          c-dependencies/js-compute-runtime/rusturl
+        key: ${{ runner.os }}-cargo-${{ hashFiles('c-dependencies/js-compute-runtime/**/Cargo.*') }}
```
elliottt, line 71: Is it enough to only hash the `Cargo.lock` files?

```suggestion
        key: ${{ runner.os }}-cargo-${{ hashFiles('c-dependencies/js-compute-runtime/**/Cargo.lock') }}
```

.github/workflows/main.yml
```diff
@@ -60,24 +59,24 @@ jobs:
...
 
     - name: "Build JS runtime (release)"
-      run: |
-        set -x
-        mkdir dist
-        cd dist
-        make -f ../c-dependencies/js-compute-runtime/Makefile
-        rm *.{d,o}
-        cd ..
+      run: make -C c-dependencies/js-compute-runtime DESTDIR=$PWD install
```
elliottt, line 79: :tada:

elliottt (approved):
This looks great to me, thanks for improving build times!

.github/workflows/main.yml
jameysharp: There is no `Cargo.lock` for rust-url. :disappointed: That's probably a bug, but I didn't want to make that decision in this PR.

.github/actions/binary-compatible-builds/run-linux
jameysharp: In that case I may as well quote `"$PREBUILT_ENGINE"` too. Done!

elliottt: Good catch!

PR #115, opened 2022-06-17
Rework the sdk-tests

elliottt:
Rework the integration tests to require less information in `sdk-test-config.json,` and to use a local viceroy instance in place of a long-running c@e instance.


integration-tests/js-compute/build-one.sh
```diff
@@ -0,0 +1,18 @@
...
+
+test="$1"
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+npm install -s
+npm run -s build:test --test="$test"
+
+cd fixtures/"$test"
+fastly compute pack --verbose --wasm-binary "./${test}.wasm"
```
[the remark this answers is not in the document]
elliottt: Great catch! I'll remove that step.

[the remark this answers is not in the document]
elliottt: I just remembered that we'll probably need this for when we're running tests on `c@e` infrastructure, so I might leave it in for now if it's not too noisy.

integration-tests/js-compute/fixtures/byte-repeater/fastly.toml
```diff
@@ -13,4 +13,4 @@ service_id = ""
   [local_server.backends]
 
     [local_server.backends.TheOrigin]
-      url = "https://compute-sdk-test-backend.edgecompute.app/"
+      url = "JS_COMPUTE_TEST_BACKEND/"
```
elliottt, line 16: The `JS_COMPUTE_TEST_BACKEND` value gets replaced before the test is run, which allows us to re-target the test to a local backend.

integration-tests/js-compute/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "install:ci": "npm ci",
-    "build:test": "base=$npm_config_test; webpack \"./fixtures/$base/$base.ts\" --output-path \"./fixtures/$base\" --output-filename \"$base.js\"; ../../target/release/js-compute-runtime \"./fixtures/$base/$base.js\" \"./fixtures/$base/$base.wasm\"",
+    "build:test": "base=$npm_config_test; webpack \"./fixtures/$base/$base.ts\" --output-path \"./fixtures/$base\" --output-filename \"$base.js\"; true",
```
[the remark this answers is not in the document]
jameysharp: Two questions:

1. Why not move the webpack call to the shell script too? I'm not sure there's any value in the `build:test` command if most of the work is done in a wrapper anyway.

2. What are the "webpack failures" in question here? I see this was in a commit related to CompressionStream which is currently a draft API we've omitted from `index.d.ts`. In general I think it'd be good to know if our tests are badly typed; that's likely either a bug in the test or in the types. So if this is a question of letting webpack fail on type errors, maybe we should do that and fix our handling of CompressionStream another way. But that can be a separate issue/PR; it shouldn't block this one.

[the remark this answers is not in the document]
elliottt: > 1. Why not move the webpack call to the shell script too? I'm not sure there's any value in the `build:test` command if most of the work is done in a wrapper anyway.

Currently `webpack` is managed by npm, which is why it's run via the script in  `package.json`. We could expose a new `webpack` script that would forward all args to the executable managed by npm, which would make it easier to call directly from the `build-one.sh` script.

> 2. What are the "webpack failures" in question here? I see this was in a commit related to CompressionStream which is currently a draft API we've omitted from `index.d.ts`. In general I think it'd be good to know if our tests are badly typed; that's likely either a bug in the test or in the types. So if this is a question of letting webpack fail on type errors, maybe we should do that and fix our handling of CompressionStream another way. But that can be a separate issue/PR; it shouldn't block this one.

The webpack failures are because we're referring to a class that's not defined: `CompressionStream`. Currently the reference in `extend-from-builtins` causes all tests to fail to build unless we ignore failures from the webpack script. The existing script in `package.json` was inadvertently ignoring errors from webpack by separating the calls to `webpack` and `js-compute-runtime` with `;` instead of `&&`, which is why this isn't causing problems on the main branch.

I think the right fix here is to add types for `CompressionStream`, but I agree that we should do that in a separate PR.

integration-tests/js-compute/replace-host.sh
```diff
@@ -0,0 +1,21 @@
...
+EOF
+  exit 1
+fi
+
+fixtures_dir="$(dirname "${BASH_SOURCE[0]}")/fixtures"
+
+fastly_toml="$fixtures_dir/$1/fastly.toml"
+override_host="$2"
+
+sed -i'' -e "s|JS_COMPUTE_TEST_BACKEND|$override_host|" "$fastly_toml"
```
jameysharp, line 21: I think it's probably necessary to rename all the `fastly.toml` files to something like `fastly.toml.in`, rather than rewriting them in-place. Eventually we'll run most tests at least twice in CI (once on Viceroy, once on C@E), and I also want people to be able to run the tests locally as many times as they need to. So the act of running the test needs to leave its inputs unchanged so they can be used again.

After doing that, I'd also want `fastly.toml` listed in `integration-tests/js-compute/fixtures/.gitignore`.

If you do rewrite them in-place, a minor nit: `sed -i''` is the same as `sed -i`; the quotes are removed by the shell.

integration-tests/js-compute/build-one.sh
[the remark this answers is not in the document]
jameysharp: Based on Jake's #108, I gather we don't need the `pack` command for deploying to C@E either, because e.g. `fastly compute deploy` should do it automatically.

Even if I'm mis-understanding that: as far as I can tell, doing it here doesn't serve a purpose for us right now, so I think we should drop it until we need it.

.github/actions/compute-sdk-test/main.js
```diff
@@ -80,19 +117,25 @@ const mainAsyncTask = async () => {
 
   // Iterate through the module tests, and run the Viceroy tests
   for (const moduleKey of moduleKeys) {
+    const testBase = `${fixtureBase}/${moduleKey}`;
+
+    // created/used by ./integration-tests/js-compute/build-one.sh
+    const fastlyTomlPath = `${testBase}/fastly.toml`;
+    const wasmPath = `${testBase}/${moduleKey}.wasm`;
+    const pkgPath = `${testBase}/pkg/${moduleKey}.tar.gz`;
```
jameysharp, line 125: `pkgPath` appears to be unused; I assume it was only for running tests live in C@E. So I think it should be removed here until we need it.

.github/actions/compute-sdk-test/main.js
```diff
@@ -80,19 +117,25 @@ const mainAsyncTask = async () => {
...
     const moduleTestKeys = Object.keys(module.tests);
     console.info(`Running tests for the module: ${moduleKey} ...`);
 
     // Spawn a new viceroy instance for the module
     viceroy = new Viceroy();
     const viceroyAddr = '127.0.0.1:8080';
-    await viceroy.spawn(module.wasm_path, {
-      config: module.fastly_toml_path,
+    await viceroy.spawn(wasmPath, {
+      config: fastlyTomlPath,
```
jameysharp, line 135: Can't this use the same `spawnViceroy` function as the backend does? That function already needs to match up with `build-one.sh` so I think you should be able to get rid of the `fastlyTomlPath` and `wasmPath` declarations here.

.github/actions/compute-sdk-test/main.js
```diff
@@ -176,14 +222,21 @@ const mainAsyncTask = async () => {
     try {
       await viceroy.kill();
     } catch(e) {
-      console.error('Could not kill Viceory. Error Below:');
+      console.error('Could not kill test Viceory instance. Error Below:');
```
jameysharp, line 225: Totally unimportant nit but, as long as you're touching this line anyway, could you fix the spelling of "Viceroy"? :laughing:

jameysharp (changes requested):
I think it's important to ensure that tests can be run twice or more, and I'd like the other small changes I've mentioned in review comments too. But in the big picture, this looks great to me!

.github/actions/compute-sdk-test/main.js
elliottt: Didn't even notice that, will do!

integration-tests/js-compute/replace-host.sh
elliottt: Yes, good point. I like explicitly rewriting `fastly.toml.in` :+1:

jameysharp (approved):

Issue #113, opened 2022-06-17
Can't call functions of classes derived from built-in classes 


harmony7:
Consider the following code that works as expected:
```javascript
class Base {
  fn() { return 'fn'; }
}
class Derived extends Base {
  constructor() {
    super();
    this.foo = () => { return 'foo'; };
  }
  bar() { return 'bar'; }
  get baz() { return 'baz'; }
}

const obj = new Derived();
console.log(Object.getOwnPropertyNames(Base.prototype));    // constructor,fn
console.log(Object.getOwnPropertyNames(Derived.prototype)); // constructor,bar,baz
console.log(Object.getOwnPropertyNames(obj));               // foo
console.log(obj.fn());                                      // fn
console.log(obj.foo());                                     // foo
console.log(obj.bar());                                     // bar
console.log(obj.baz);                                       // baz
console.log(obj.foo === undefined);                         // false
console.log(obj.bar === undefined);                         // false
console.log(obj.baz === undefined);                         // false
```

When you try to do the same thing by extending builtins:
```javascript
class MyRequest extends Request {
  constructor(input, init) {
    super(input, init);
    this.foo = () => { return 'foo'; };
  }
  bar() { return 'bar'; }
  get baz() { return 'baz'; }
}

const request = new MyRequest('https://www.google.com/');
console.log(Object.getOwnPropertyNames(Request.prototype));   // constructor,method,url,version,headers,body,bodyUsed,arrayBuffer,json,text,setCacheOverride
console.log(Object.getOwnPropertyNames(MyRequest.prototype)); // ** constructor,bar,baz
console.log(Object.getOwnPropertyNames(request));             // foo
console.log(request.url);                                     // https://www.google.com/
console.log(request.foo());                                   // foo
console.log(request.bar());                                   // ** Error: request.bar is not a function
console.log(request.baz);                                     // ** undefined
console.log(request.foo === undefined);                       // false
console.log(request.bar === undefined);                       // ** true
console.log(request.baz === undefined);                       // ** true
```

Look at the logs relating to `bar` and `baz` (marked with asterisks above). You can see that the two methods (`bar` method and `baz` getter) added through the class definition are added to the prototype but have undefined values. Note that the property added in the constructor (`foo`) does work, however.

The same problem happens whether we are extending `Request`, `Response`, `Headers`, `URL`, `URLSearchParams`, etc. so I believe it is just for the builtins, and I believe it's a bug.

JakeChampion:
[...]

PR #116, opened 2022-06-20
Allow classes to extend from Request

JakeChampion:
This is part of the solution for https://github.com/fastly/js-compute-runtime/issues/113 - that issue shows that all our built-in classes are not extendable within applications' JavaScript.

This pull-request changes our Request to use `JS_NewObjectForConstructor` when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account `extends` in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from `Request`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,16 @@
 
 # Integration Tests build output
 integration-tests/**/fixtures/**/*.tar.gz
-integration-tests/**/fixtures/**/fastly.toml
+integration-tests/**/fixtures/**/pkg/**/fastly.toml
 integration-tests/**/fixtures/**/*.wasm
 integration-tests/**/fixtures/**/*.js
 integration-tests/**/fixtures/**/*.js.map
+integration-tests/**/fixtures/**/compiler_flags
+integration-tests/**/fixtures/**/js-compute-builtins.d
+integration-tests/**/fixtures/**/js-compute-builtins.o
+integration-tests/**/fixtures/**/js-compute-runtime.d
+integration-tests/**/fixtures/**/js-compute-runtime.o
+integration-tests/**/fixtures/**/rusturl
 
 # Ignore Reference Docs
 # TODO
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -4086,19 +4086,6 @@ bool apply_cache_override(JSContext *cx, HandleObject self) {
                                                          sk_chars.get(), sk_len));
 }
 
-JSObject *create(JSContext *cx, HandleValue input, HandleValue init);
-
-bool constructor(JSContext *cx, unsigned argc, Value *vp) {
-  REQUEST_HANDLER_ONLY("The Request builtin");
-  CTOR_HEADER("Request", 1);
-  RootedObject request(cx, create(cx, args[0], args.get(1)));
-  if (!request)
-    return false;
-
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -2,10 +2,16 @@
 
 # Integration Tests build output
 integration-tests/**/fixtures/**/*.tar.gz
-integration-tests/**/fixtures/**/fastly.toml
+integration-tests/**/fixtures/**/pkg/**/fastly.toml
 integration-tests/**/fixtures/**/*.wasm
 integration-tests/**/fixtures/**/*.js
 integration-tests/**/fixtures/**/*.js.map
+integration-tests/**/fixtures/**/compiler_flags
```
JakeChampion, line 9: These files are creating when compiling an integration test manually via make, which is something I've been doing a lot recently as it is a faster feedback loop when developing against a specific test:
```
INIT_JS=extend-from-builtins.ts make -f ../../../../c-dependencies/js-compute-runtime/Makefile initialized-js-compute-runtime.wasm
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4198,23 +4202,39 @@ bool init_class(JSContext *cx, HandleObject global) {
...
-  JS::SetReservedSlot(request, Slots::Headers, JS::NullValue());
-  JS::SetReservedSlot(request, Slots::Body, JS::Int32Value(body_handle.handle));
-  JS::SetReservedSlot(request, Slots::BodyStream, JS::NullValue());
-  JS::SetReservedSlot(request, Slots::HasBody, JS::FalseValue());
-  JS::SetReservedSlot(request, Slots::BodyUsed, JS::FalseValue());
-  JS::SetReservedSlot(request, Slots::Method, JS::StringValue(GET_atom));
-  JS::SetReservedSlot(request, Slots::CacheOverride, JS::NullValue());
-  JS::SetReservedSlot(request, Slots::IsDownstream, JS::BooleanValue(is_downstream));
+  return Request::create(cx, requestInstance, request_handle, body_handle, is_downstream);
+}
```
JakeChampion, line 4215: This method exists for use within the `prepare_downstream_request` function, which should always use the runtime's Request implementation and not one which could be supplied by the applications' JS

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4198,23 +4202,39 @@ bool init_class(JSContext *cx, HandleObject global) {
...
+  return requestInstance;
+}
+
+JSObject *create(JSContext *cx, HandleValue input, HandleValue init) {
+  RootedObject requestInstance(cx, JS_NewObjectWithGivenProto(cx, &class_, proto_obj));
+  if (!requestInstance)
+    return nullptr;
+
+  return Request::create(cx, requestInstance, input, init);
 }
```
elliottt, line 4238: Is this `create` method still used? It looks like the same signature as the one that was changed above, but it's hard to tell what might still be using it.

elliottt (approved):
This looks great! I had one question about the new `create` overload introduced, but feel free to reply and merge if it's benign.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: Yes, this is used in our fetch implementation - I've changed the code now to not make use of this create method

PR #121, opened 2022-06-21
Allow CacheOverride to be used as a base class to extend from within application javascript

JakeChampion:

This pull-request changes our CacheOverride implementation to use JS_NewObjectForConstructor when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account extends in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from CacheOverride, the test is written in a way that makes it simpler to add the same test for any other builtin classes

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -3899,14 +3899,24 @@ template <auto accessor_fn> bool accessor_set(JSContext *cx, unsigned argc, Valu
 
 const unsigned ctor_length = 1;
 
-JSObject *create(JSContext *cx);
+const JSFunctionSpec methods[] = {JS_FS_END};
+
+const JSPropertySpec properties[] = {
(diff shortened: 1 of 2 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3899,14 +3899,24 @@ template <auto accessor_fn> bool accessor_set(JSContext *cx, unsigned argc, Valu
...
+bool constructor(JSContext *cx, unsigned argc, Value *vp);
+CLASS_BOILERPLATE(CacheOverride)
 
 bool constructor(JSContext *cx, unsigned argc, Value *vp) {
   CTOR_HEADER("CacheOverride", 1);
 
-  RootedObject self(cx, create(cx));
-  if (!self)
-    return false;
+  RootedObject self(cx, JS_NewObjectForConstructor(cx, &class_, args));
```
elliottt, line 3919: Nice!

elliottt (approved):
Awesome!

PR #122, opened 2022-06-21
Allow TextDecoder to be used as a base class to extend from within application javascript

JakeChampion:

This pull-request changes our TextDecoder implementation to use JS_NewObjectForConstructor when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account extends in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from TextDecoder, the test is written in a way that makes it simpler to add the same test for any other builtin classes


elliottt (approved):
:+1:

PR #123, opened 2022-06-21
Allow TextEncoder to be used as a base class to extend from within application javascript

JakeChampion:

This pull-request changes our TextEncoder implementation to use JS_NewObjectForConstructor when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account extends in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from TextEncoder, the test is written in a way that makes it simpler to add the same test for any other builtin classes


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -5115,10 +5102,17 @@ bool encoding_get(JSContext *cx, unsigned argc, Value *vp) {
...
 const JSPropertySpec properties[] = {JS_PSG("encoding", encoding_get, JSPROP_ENUMERATE), JS_PS_END};
-
+bool constructor(JSContext *cx, unsigned argc, Value *vp);
 CLASS_BOILERPLATE(TextEncoder)
 
-JSObject *create(JSContext *cx) { return JS_NewObjectWithGivenProto(cx, &class_, proto_obj); }
+bool constructor(JSContext *cx, unsigned argc, Value *vp) {
+  CTOR_HEADER("TextEncoder", 0);
+
+  RootedObject self(cx, JS_NewObjectForConstructor(cx, &class_, args));
```
JakeChampion, line 5111:
```suggestion
  RootedObject self(cx, JS_NewObjectForConstructor(cx, &class_, args));
  if(!self)
      return false;
```

elliottt (approved):
Looks good!

PR #124, opened 2022-06-21
Allow URLSearchParams to be used as a base class to extend from within application javascript

JakeChampion:

This pull-request changes our URLSearchParams implementation to use JS_NewObjectForConstructor when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account extends in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from URLSearchParams, the test is written in a way that makes it simpler to add the same test for any other builtin classes

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -6742,11 +6742,11 @@ bool searchParams_get(JSContext *cx, unsigned argc, Value *vp) {
   RootedObject params(cx);
   if (params_val.isNullOrUndefined()) {
     JSUrl *url = (JSUrl *)JS::GetReservedSlot(self, Slots::Url).toPrivate();
-    RootedObject url_search_params_intance(
+    RootedObject url_search_params_instance(
         cx, JS_NewObjectWithGivenProto(cx, &URLSearchParams::class_, URLSearchParams::proto_obj));
-    if (!self)
+    if (!url_search_params_instance)
       return false;
-    params = URLSearchParams::create(cx, url_search_params_intance, url);
+    params = URLSearchParams::create(cx, url_search_params_instance, url);
     if (!params)
       return false;
     JS::SetReservedSlot(self, Slots::Params, JS::ObjectValue(*params));
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -954,11 +954,63 @@ bool SetBackpressure(JSContext *cx, HandleObject stream, bool backpressure);
 bool Error(JSContext *cx, HandleObject stream, HandleValue error);
 } // namespace TransformStream
 
+#define ITERTYPE_ENTRIES 0
+#define ITERTYPE_KEYS 1
+#define ITERTYPE_VALUES 2
 namespace URLSearchParams {
-bool is_instance(JSObject *obj);
-JSObject *create(JSContext *cx, jsurl::JSUrl *url);
+
+namespace Slots {
+enum { Url, Params, Count };
+};
+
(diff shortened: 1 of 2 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -954,11 +954,63 @@ bool SetBackpressure(JSContext *cx, HandleObject stream, bool backpressure);
...
+} // namespace detail
+
 SpecSlice serialize(JSContext *cx, HandleObject self);
+
+bool check_receiver(JSContext *cx, HandleValue receiver, const char *method_name);
+
+const unsigned ctor_length = 1;
+
+bool append(JSContext *cx, unsigned argc, Value *vp);
+
```
elliottt, line 979: I was wondering where the implementations for these prototypes were, and it's down on line 6935 on main -- the `URLSearchParams` namespace is reopened three times throughout `js-compute-builtins.cpp`. We should do some reorganization soon so that we don't need prototypes so far from the definitions :)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6690,7 +6742,11 @@ bool searchParams_get(JSContext *cx, unsigned argc, Value *vp) {
   RootedObject params(cx);
   if (params_val.isNullOrUndefined()) {
     JSUrl *url = (JSUrl *)JS::GetReservedSlot(self, Slots::Url).toPrivate();
-    params = URLSearchParams::create(cx, url);
+    RootedObject url_search_params_intance(
+        cx, JS_NewObjectWithGivenProto(cx, &URLSearchParams::class_, URLSearchParams::proto_obj));
+    if (!self)
+      return false;
```
elliottt, line 6748: I'm not sure this is necessary here -- line 6744 is using `JS::GetReservedSlot` on `self`.

elliottt (changes requested):
Looks great, I just had one question about testing `self` in `searchParams_get`.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6690,7 +6742,11 @@ bool searchParams_get(JSContext *cx, unsigned argc, Value *vp) {
   RootedObject params(cx);
   if (params_val.isNullOrUndefined()) {
     JSUrl *url = (JSUrl *)JS::GetReservedSlot(self, Slots::Url).toPrivate();
-    params = URLSearchParams::create(cx, url);
+    RootedObject url_search_params_intance(
+        cx, JS_NewObjectWithGivenProto(cx, &URLSearchParams::class_, URLSearchParams::proto_obj));
+    if (!self)
```
JakeChampion, line 6747:
```suggestion
    if (!url_search_params_intance)
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6690,7 +6742,11 @@ bool searchParams_get(JSContext *cx, unsigned argc, Value *vp) {
...
+      return false;
+    params = URLSearchParams::create(cx, url_search_params_intance, url);
```
JakeChampion, line 6749:
```suggestion
    RootedObject url_search_params_instance(
        cx, JS_NewObjectWithGivenProto(cx, &URLSearchParams::class_, URLSearchParams::proto_obj));
    if (!url_search_params_instance)
      return false;
    params = URLSearchParams::create(cx, url_search_params_instance, url);
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: good spot, this was meant to be `url_search_params_instance`

elliottt (approved):

PR #125, opened 2022-06-22
Allow URL to be used as a base class to extend from within application javascript

JakeChampion:

This pull-request changes our URL implementation to use JS_NewObjectForConstructor when the construction has come from the applications' JavaScript, which assigns the correct prototype to the instance being constructed (taking into account extends in JavaScript)

I've also added a test which confirms that the correct prototype has been assigned when extending from URL, the test is written in a way that makes it simpler to add the same test for any other builtin classes


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6767,32 +6803,21 @@ bool toJSON(JSContext *cx, unsigned argc, Value *vp) {
...
-
-CLASS_BOILERPLATE(URL)
+bool constructor(JSContext *cx, unsigned argc, Value *vp) {
+  CTOR_HEADER("URL", 1);
 
-JSObject *create(JSContext *cx, SpecString url_str, const JSUrl *base) {
-  RootedObject self(cx, JS_NewObjectWithGivenProto(cx, &class_, proto_obj));
+  RootedObject urlInstance(cx, JS_NewObjectForConstructor(cx, &class_, args));
+  RootedObject self(cx, create(cx, urlInstance, args.get(0), args.get(1)));
   if (!self)
```
elliottt, line 6813: Should this check that `urlInstance` is not `nullptr`?

elliottt (changes requested):
Looks good, just one question about checking an allocated object.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -6767,32 +6803,21 @@ bool toJSON(JSContext *cx, unsigned argc, Value *vp) {
...
-    JS_PSGS("username", username_get, username_set, JSPROP_ENUMERATE),
-    JS_PS_END};
-
-CLASS_BOILERPLATE(URL)
+bool constructor(JSContext *cx, unsigned argc, Value *vp) {
+  CTOR_HEADER("URL", 1);
 
-JSObject *create(JSContext *cx, SpecString url_str, const JSUrl *base) {
-  RootedObject self(cx, JS_NewObjectWithGivenProto(cx, &class_, proto_obj));
+  RootedObject urlInstance(cx, JS_NewObjectForConstructor(cx, &class_, args));
```
JakeChampion, line 6811:
```suggestion
  RootedObject urlInstance(cx, JS_NewObjectForConstructor(cx, &class_, args));
  if (!urlInstance)
    return false;
```

elliottt (approved):

PR #129, opened 2022-06-22
Allow the compute-sdk tests to run without building locally

elliottt:
This change allows us to run `node .github/actions/compute-sdk-test/main.js` without first running `npm build` in that directory. We still need to run `npm install` there, but this simplifies iteration when working locally.

I also updated the `install-rust` action to `node16` while I was at it.


elliottt:
> I haven't really reviewed this yet, but a quick thought regarding the install-rust action specifically: it's almost empty now, it should probably be a shell fragment in a composite action instead, or even just be inlined into the only place it's used in the main workflow.

Seems reasonable to me! I switched it over to a composite action in the third commit.

.github/actions/install-rust/action.yml
```diff
@@ -8,5 +8,20 @@ inputs:
...
+        rustup show
+
+        # Save disk space by avoiding incremental compilation.
+        echo 'CARGO_INCREMENTAL=0' >> "$GITHUB_ENV"
+
+        # Turn down debuginfo from 2 to 1 to help save disk space
+        echo 'CARGO_PROFILE_DEV_DEBUG=1' >> "$GITHUB_ENV"
+        echo 'CARGO_PROFILE_TEST_DEBUG=1' >> "$GITHUB_ENV"
+
+    - if: ${{ ! startsWith(inputs.toolchain, 'nightly') }}
```
jameysharp, line 24: Oh yeah, this reminds me: I broke the "toolchain" input when I switched this over to just reading `rust-toolchain.toml`. That input no longer controls the toolchain that's getting installed. So this test doesn't do what any reasonable person would think it's doing. I definitely agree this commit accurately translates `main.js` as it stands now into a composite action though!

integration-tests/js-compute/package-lock.json
```diff
@@ -1,2404 +1,6 @@
...
-      "dev": true,
-      "engines": {
-        "node": ">=10"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/sindresorhus"
-      }
-    }
-  },
+  "lockfileVersion": 1,
```
JakeChampion, line 3: the lockfile version that's been generated here is an older version than we were using previously, this is likely due to using an older version of npm during development on this branch.

It's not a blocker but it does cause for a larger diff than necessary on this file.

JakeChampion (approved):

.github/actions/install-rust/action.yml
elliottt: We should figure out if we'd like to keep this. It seems nice to have a reusable place to setup the environment variables we're relying on, but the action name seems a bit misleading now :)

jameysharp: `rustup show` _does_ install the toolchain though, as a side-effect. :laughing: Maybe I should have commented to that effect when I made that change... It just has the nice bonus of recording in the CI log exactly what toolchain it ran with.

PR #132, opened 2022-06-28
Add a script to run all integration tests

elliottt:
Make it easier to run the integration tests locally by adding a script to wrap running the github action.

### DIFF
--- a/integration-tests/js-compute/run-all.sh
+++ b/integration-tests/js-compute/run-all.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


jameysharp (approved):
I suppose you could `cd "$(dirname "${BASH_SOURCE[0]}")"/../..` at the beginning of the script and then use relative paths rather than using `${root}` anywhere, but it doesn't matter. This seems fine, and having a script to easily run these tests locally is great. I also appreciate that you check for so many things that might be a surprise for a new contributor.

elliottt:
I think your suggestion to change directory at the beginning of the script makes it more clear, so I've made that change.

PR #133, opened 2022-06-28
Refactor the build-one.sh script

elliottt:
- Switch to `npm ci`, which will not update the package-lock.json file
- Move the call to webpack into `build-one.sh`


jameysharp (approved):
Looks great! Would there be any additional value in making `npm build:test` still work, as a trivial wrapper around `build-one.sh`? I don't know if that's a common npm idiom for local development or something.

elliottt:
> Looks great! Would there be any additional value in making `npm build:test` still work, as a trivial wrapper around `build-one.sh`? I don't know if that's a common npm idiom for local development or something.

Possibly! I've been only using the `build-one.sh` script directly, but we can always add the script entry back in to the package.json later :+1:

PR #134, opened 2022-06-28
Move test specification into fixture dirs

elliottt:
Remove the single authoritative `sdk-test-config.json` in favor of moving the individual cases into the fixture directory for the test it describes. This allows new tests to be created by adding the appropriate files into a sub-directory of `integration-tests/js-compute` only.

This will also avoid a common source of merge conflicts, as multiple PRs adding tests will no longer interact through the `sdk-test-config.json` file.


jameysharp (approved):
I love it. :+1:

PR #135, opened 2022-06-29
Add a class to manage integration test configuration

elliottt:
In order to make the compute-sdk-test action reusable for other sdks, and to wrangle its current complexity, this PR adds a class to track configuration information and plumbs it through the tests. The next step will be to break up the test running logic in `main.js` so that it's possible to run one test at a time.


jameysharp (approved):
On the one hand, my inclination would be to take this in a different direction, namely:
- put as little logic in the GitHub Action as possible
- make each test responsible for all of its own setup; ideally each test would just be an arbitrary executable script, whether that's JS or shell or whatever's convenient
- factor out all the common stuff into importable or sourceable libraries and let the tests invoke whichever parts they need

On the other hand, I think this is an improvement over what's already there, so, :+1:

PR #141, opened 2022-06-30
test: Set a timeout when waiting for viceroy to start

### LINKED ISSUE #140: Timeout test interactions with viceroy
Tests will hang forever when viceroy doesn't implement an imported host call. Let's audit the testing infrastructure to make sure that anywhere we're waiting on viceroy to initialize, we also have a timeout so that we can detect these problems. Here's an example build that was killed after hanging: https://github.com/fastly/js-compute-runtime/actions/runs/2589670789

elliottt:
When starting viceroy during tests, race with a timeout promise to ensure that we don't wait forever if the viceroy instance crashed.

Fixes #140

### DIFF
--- a/tests/wpt-harness/run-wpt.mjs
+++ b/tests/wpt-harness/run-wpt.mjs
@@ -325,8 +325,8 @@ async function startViceroy(runtime, config, logLevel) {
     });
   }
 
-  // give viceroy 10 seconds to become available
-  const VICEROY_READY_TIMEOUT = 10000;
+  // give viceroy 20 seconds to become available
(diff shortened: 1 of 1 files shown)


jameysharp (approved):
Seems like an important improvement. Any particular reason that one timeout is 10 seconds and the other is 20?

elliottt:
> Seems like an important improvement. Any particular reason that one timeout is 10 seconds and the other is 20?

I had to bump the wpt tests to 20s to get CI to pass. My guess is that the image being loaded in that case takes longer to load, while the integration test images are usually quite small.

PR #142, opened 2022-06-30
test: Bump the integration test timeout to 20s

elliottt:
Looks like I spoke too soon on #141, as there was a timeout for ci on main.


jameysharp (approved):
Haha, yeah, let's be patient with the poor CI servers.

PR #147, opened 2022-07-01
Reuse Cargo caches across jobs

jameysharp:


.github/actions/install-rust/action.yml
```diff
@@ -2,26 +2,28 @@ name: 'Install Rust toolchain'
...
-
-        # Save disk space by avoiding incremental compilation.
-        echo 'CARGO_INCREMENTAL=0' >> "$GITHUB_ENV"
+        # Save disk space: avoid incremental compilation, and turn down
+        # debuginfo from 2 to 1.
+        cat >> "$GITHUB_ENV" <<EOF
+        CARGO_INCREMENTAL=0
+        CARGO_PROFILE_DEV_DEBUG=1
+        CARGO_PROFILE_TEST_DEBUG=1
+        EOF
```
elliottt, line 21: Will this end up putting a bunch of leading space on these lines? I'm not entirely sure what github actions passes to bash for a multi-line script like this.

.github/workflows/main.yml
```diff
@@ -244,9 +241,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-      with:
-        submodules: true
-    - run: rustup update stable && rustup default stable
```
elliottt, line 249: Nice catch!

elliottt (approved):
This looks great to me! I just had one comment about the use of a heredoc when setting environment variables.

.github/actions/install-rust/action.yml
jameysharp: That's an interesting question that I hadn't thought about. I can find two pieces of evidence that suggest it's fine:

When the runner logs the script it's about to run together with the environment variables available to the script, those are both at the same indentation level: https://github.com/fastly/js-compute-runtime/runs/7157447404?check_suite_focus=true#step:3:10

And when the runner logs the environment before doing the actual build, these variables are indented the same as all the others: https://github.com/fastly/js-compute-runtime/runs/7157447404?check_suite_focus=true#step:8:4

So I think GitHub must be deleting a consistent amount of whitespace from the beginning of all lines of the script. I wouldn't put it past the YAML spec to require parsers to interpret it that way, honestly.

PR #148, opened 2022-07-01
Replace cargo-audit with GitHub's dependency-review

jameysharp:
GitHub's version works across several package ecosystems and also
supports checking the licenses of dependencies.


elliottt (approved):
Nice!

PR #149, opened 2022-07-01
dependency-review-action only works on pull requests

jameysharp:


elliottt (approved):
Should we add it to the list of steps that are required to pass before merging?

jameysharp:
Yeah, I think we should block merge if the dependency review fails, so I've done that.

Issue #23, opened 2021-08-12
Introduce a proper way to detect "execution mode" in runtime code


tschneidereit:
We currently detect whether we're running build-time initialization code or post-deploy event handling code by checking for the existence of a `FetchEvent` instance. As @cfallin rightly [points out](https://github.com/fastly/js-compute-runtime/pull/22#discussion_r688114136), that isn't ideal, so we should have an explicit way to set and query the execution mode we're in.

PR #151, opened 2022-07-05
Use an enum to explicitly state what the execution phase is that the application is currently in

JakeChampion:
resolves https://github.com/fastly/js-compute-runtime/issues/23

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -50,6 +50,16 @@ using JS::MutableHandleValue;
 
 using JS::PersistentRooted;
 
+enum class Mode { PreWizening, PostWizening };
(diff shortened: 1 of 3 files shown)


elliottt (dismissed):
This approach makes sense to me, however what about not exposing the enum across the boundary to js-compute-runtime.cpp? You could instead make a function that would toggle the relevant internal state and expose that, to avoid the possibility of reverting back to the `PreWizening` state.

JakeChampion:
> This approach makes sense to me, however what about not exposing the enum across the boundary to js-compute-runtime.cpp? You could instead make a function that would toggle the relevant internal state and expose that, to avoid the possibility of reverting back to the `PreWizening` state.

Great suggestion, I have refactored to follow your suggestion now 👍

elliottt (approved):
Looks great!

PR #152, opened 2022-07-06
Update to version of spidermoney-wasi-embedding which uses arc4random to seed the prng

JakeChampion:
This changes which commit sha we pull from to one which ensures spidermonkey uses arc4random when compiled to wasi 
By using arc4random, the prng is seeded correctly and will no longer have the first two generated numbers be similar

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "c-dependencies/spidermonkey"]
 	path = c-dependencies/spidermonkey
-	url = https://github.com/tschneidereit/spidermonkey-wasi-embedding.git
+	url = https://github.com/fastly/spidermonkey-wasi-embedding.git
 	shallow = true
 [submodule "tests/wpt-harness/wpt"]
 	path = tests/wpt-harness/wpt
--- a/integration-tests/js-compute/fixtures/random/fastly.toml.in
+++ b/integration-tests/js-compute/fixtures/random/fastly.toml.in
@@ -1,0 +1,9 @@
+# This file describes a Fastly Compute@Edge package. To learn more visit:
+# https://developer.fastly.com/reference/fastly-toml/
+
+authors = ["jchampion@fastly.com"]
+description = ""
+language = "javascript"
+manifest_version = 2
+name = "random"
+service_id = ""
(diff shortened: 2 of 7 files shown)


elliottt (approved):
This seems like a good interim fix, and an excellent test case! As we're the ones that introduced `js::ResetMathRandomSeed` it might be worth seeing if there's something funny about how we're resetting the seed that's causing this problem.

JakeChampion:
> This seems like a good interim fix, and an excellent test case! As we're the ones that introduced `js::ResetMathRandomSeed` it might be worth seeing if there's something funny about how we're resetting the seed that's causing this problem.

There is something funny yes, because there is no hostcall in wasi for getting a random value, spidermonkey fallbacks to using the current timestamp for both seed values (the PRNG takes two seeds) which is likely the reason why the fix two generated numbers are similar as the two seeds are meant to be different values.

two options I have thought of which spidermonkey could do:
- wait for wasi to include a hostcall for getting a random value and then use it when compiling for wasm --> https://github.com/WebAssembly/wasi-random

- don't wait for wasi-random and instead use arc4random from wasi-libc to generate two different seed values, falling back to the timestamp approach currently being used in spidermonkey for environments which don't have arc4random

I think the second option is good and we could do that already in our fork of spidermonkey if we want to.



The current solution in this pull-request should be okay for us right now as the first call to Math.random is done during the wizening phase and so isn't going to occur during execution on c@e which should mean no degradation in performance

JakeChampion:
>     * don't wait for wasi-random and instead use arc4random from wasi-libc to generate two different seed values, falling back to the timestamp approach currently being used in spidermonkey for environments which don't have arc4random
> 
> 
> I think the second option is good and we could do that already in our fork of spidermonkey if we want to.


I've gone ahead and implemented this option and updated this pull-request to use it 👍 
The below two pull-requests are where the implementations exists:
- https://github.com/bytecodealliance/gecko-dev/pull/1
- https://github.com/fastly/spidermonkey-wasi-embedding/pull/1

elliottt (approved):
This all looks good to me!

It might be worth adding a comment that the test might fail sporadically: there's no reason that a correct implementation won't return two sequential results that have the same four digits.

PR #156, opened 2022-07-14
fix: Mark requests as pending after their body stream has closed

### LINKED ISSUE #154: ReadableStream.tee causes fetch to hang
Running the following example:

```js
  fastly.enableDebugLogging(true);                                                 
                                                                                   
  addEventListener("fetch", event => event.respondWith(handleRequest(event.request)));
                                                                                   
  async function handleRequest(req) {                                     
    let [body1, _body2] = req.body.tee();                                          
                                                                                   
    return fetch(new Request(req.url, {                                            
      body: body1,                                                                 
      headers: req.headers,                                                        
      method: req.method,                                                          
      backend: "origin_0"                                                          
    }));                                                                           
  }
```

the response is never delivered, and the runtime hangs in `process_pending_requests` at the host call for `xqd_req_pending_req_select` (`pending_req_select` in viceroy). I've only verified this working locally with viceroy, but it does exhibit the same failure mode in a [fastly fiddle](https://fiddle.fastlydemo.net/fiddle/a9fe0699).

It's worth noting that the following program works fine, and exits the call to `xqd_req_pending_req_select` immediately:

```js
  fastly.enableDebugLogging(true);                                                 
                                                                                   
  addEventListener("fetch", event => event.respondWith(handleRequest(event.request)));
                                                                                   
  async function handleRequest(req) {     
    let body1 = req.body;                                
    // let [body1, _body2] = req.body.tee();                                          
                                                                                   
    return fetch(new Request(req.url, {                                            
      body: body1,                                                                 
      headers: req.headers,                                                        
      method: req.method,                                                          
      backend: "origin_0"                                                          
    }));                                                                           
  }
```

elliottt:
Fixes #154 

When a request's body is produced by a stream that must be read into the js-compute-runtime (such as those created with `tee` from an existing request's body) we would place the request in the same pending queue that we would use for requests whose request body handle was already closed. The effect of this was that once we entered `process_pending_requests`, we would deadlock if the only request available for processing was the one waiting on its channel to close.

The solution implemented in this PR is to only queue the request for processing when we know that its body has been closed. This change needed to happen in two places:
1. In the `fetch` implementation we need to only enqueue the request if it's not going to be streamed
2. The `then` handler constructed in `maybe_stream_body` needs to keep track of the original request and add it to the pending queue when its stream closes.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1634,8 +1634,7 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
   // The reader is stored in the catch handler, which we need here as well.
   // So we get that first, then the reader.
   RootedObject catch_handler(cx, &extra.toObject());
-  RootedObject closure(cx, &js::GetFunctionNativeReserved(catch_handler, 1).toObject());
-
+  RootedObject reader(cx, &js::GetFunctionNativeReserved(catch_handler, 1).toObject());
   BodyHandle body_handle = RequestOrResponse::body_handle(body_owner);
 
   // We're guaranteed to work with a native ReadableStreamDefaultReader here,
@@ -1659,16 +1658,8 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
       return false;
     }
 
-    // If the owner is a Request, it's now safe to push it into the pending requests queue as its
-    // body is closed.
-    RootedValue ownerVal{cx};
-    if (!JS_GetProperty(cx, closure, "owner", &ownerVal)) {
-      return false;
-    }
-
-    RootedObject owner{cx, &ownerVal.toObject()};
-    if (Request::is_instance(owner)) {
-      if (!pending_requests->append(owner)) {
+    if (Request::is_instance(body_owner)) {
+      if (!pending_requests->append(body_owner)) {
         return false;
       }
     }
@@ -1706,21 +1697,13 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
   }
 
   // Read the next chunk.
-  RootedValue readerVal{cx};
-  if (!JS_GetProperty(cx, closure, "reader", &readerVal)) {
-    return false;
-  }
-
-  RootedObject reader{cx, &readerVal.toObject()};
   RootedObject promise(cx, JS::ReadableStreamDefaultReaderRead(cx, reader));
-  if (!promise) {
+  if (!promise)
     return false;
-  }
-
   return JS::AddPromiseReactions(cx, promise, then_handler, catch_handler);
 }
 
-bool body_reader_catch_handler(JSContext *cx, HandleObject body_owner, HandleValue closure,
+bool body_reader_catch_handler(JSContext *cx, HandleObject body_owner, HandleValue extra,
                                CallArgs args) {
   // TODO: check if this should create a rejected promise instead, so an
   // in-content handler for unhandled rejections could deal with it. The body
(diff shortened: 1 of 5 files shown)


integration-tests/js-compute/fixtures/backend/backend.js
```diff
@@ -113,6 +113,13 @@ function handleRequest(event) {
     return new Response("Compute SDK Test Backend");
   }
 
+  // tee
+
+  if (method == "POST" && url.pathname == "/tee") {
+    const text = await event.request.text();
+    return new Response(text);
+  }
```
elliottt, line 121: Adding a new testing endpoint that can return a post body for use with the `tee` test.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -7482,8 +7520,12 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
...
     return ReturnPromiseRejectedWithPendingError(cx, args);
 
-  if (!pending_requests->append(request))
-    return ReturnPromiseRejectedWithPendingError(cx, args);
+  // If the request body is streamed, we need to wait for streaming to complete before marking the
+  // request as pending.
+  if (!streaming) {
+    if (!pending_requests->append(request))
+      return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
elliottt, line 7528: This is part 1 of the PR description.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1653,7 +1654,26 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
...
+    if (!JS_GetProperty(cx, closure, "owner", &ownerVal)) {
+      return false;
+    }
+
+    RootedObject owner{cx, &ownerVal.toObject()};
+    if (Request::is_instance(owner)) {
+      if (!pending_requests->append(owner)) {
+        return false;
+      }
+    }
```
elliottt, line 1674: This is part 2 from the PR description. Note that since `body_reader_then_handler` is used for both request and response body processing, we only enqueue when the owner of the stream is a `Request`.

jameysharp: Is `owner` different than `body_owner`? I'm not following the types real well here but I would have guessed that if `RequestOrResponse::body_handle` is valid to call on `body_owner`, and so is `Response::is_instance`, then surely `Request::is_instance` is too?

If so then I'd expect to see something more like this here:

```c++
    if (Response::is_instance(body_owner)) {
      FetchEvent::set_state(FetchEvent::instance(), FetchEvent::State::responseDone);
    } else {
      if (!pending_requests->append(body_owner)) {
        return false;
      }
    }
```

(Optionally, with an `assert(Request::is_instance(body_owner))` in the `else` branch.)

And in that case, does that mean you can undo the introduction of the `closure` object and just go back to passing the `reader` around?

jameysharp (changes requested):
If my wild guesses pan out about already having available the objects you need, then I think it's important to not plumb them through a second time. If I'm misreading things, on the other hand, I think this is fine. Either way I think you're close: the tests look good and the general idea of the fix feels right.

integration-tests/js-compute/fixtures/backend/backend.js
```diff
@@ -2,7 +2,7 @@
 
 let spinLoopString = '';
 
-function handleRequest(event) {
+async function handleRequest(event) {
```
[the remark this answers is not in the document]
elliottt: I added this because there were uses of `await` in the function body: I think it was an oversight that it wasn't there before.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
elliottt: You're totally right, `body_owner` will be baked in as the receiver for the handler, and will be either the `Response` or `Request` objects. I'll revert the introduction of the `closure`, thanks for catching this!

jameysharp (approved):

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1692,10 +1703,8 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
   return JS::AddPromiseReactions(cx, promise, then_handler, catch_handler);
 }
 
-bool body_reader_catch_handler(JSContext *cx, HandleObject body_owner, HandleValue reader_val,
+bool body_reader_catch_handler(JSContext *cx, HandleObject body_owner, HandleValue extra,
                                CallArgs args) {
-  RootedObject reader(cx, &reader_val.toObject());
-
```
elliottt, line 1698: `reader` wasn't used in the body of this function, and as `reader_val` is already a `HandleValue` that implies that it's rooted elsewhere.

PR #159, opened 2022-07-19
chore: When calling make from build.rs, pass `-j $NUM_JOBS`

elliottt:
Beginning to split up js-compute-builtins.cpp means that we'll benefit more from running make with a `-j` argument. This PR adds that flag to the make invocation in `build.rs`, and also ensures that the `builtins` output directory exists when building outside of the `c-dependencies/js-compute-runtime` tree.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -34,6 +34,8 @@ fn build_engine(src: &PathBuf, out_dir: &Path) {
     let makefile_path = src.join("js-compute-runtime").join("Makefile");
     let status = Command::new("make")
         .current_dir(out_dir)
+        .arg("-j")
+        .arg("2")
         .arg("-f")
         .arg(makefile_path)
         .status()
--- a/build.rs
+++ b/build.rs
@@ -35,7 +35,7 @@ fn build_engine(src: &PathBuf, out_dir: &Path) {
     let status = Command::new("make")
         .current_dir(out_dir)
         .arg("-j")
-        .arg("2")
+        .arg(var_os("NUM_JOBS").unwrap())
         .arg("-f")
         .arg(makefile_path)
         .status()
--- a/c-dependencies/js-compute-runtime/Makefile
+++ b/c-dependencies/js-compute-runtime/Makefile
@@ -59,7 +59,10 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
 %.o: $(FSM_SRC)%.cpp $(FSM_SRC)Makefile $(RUST_URL_LIB) compiler_flags
 	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) -I $(SM_SRC)include -MMD -MP -c -o $@ $<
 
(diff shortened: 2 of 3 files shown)


c-dependencies/js-compute-runtime/Makefile
```diff
@@ -59,7 +59,10 @@ $(RUST_URL_LIB): $(RUST_URL_RS_FILES) $(RUST_URL_SRC)/Cargo.toml $(RUST_URL_SRC)
 %.o: $(FSM_SRC)%.cpp $(FSM_SRC)Makefile $(RUST_URL_LIB) compiler_flags
 	$(WASI_CXX) $(CXX_FLAGS) $(CXX_OPT) $(DEFINES) -I $(SM_SRC)include -MMD -MP -c -o $@ $<
 
-builtins/%.o: $(FSM_SRC)builtins/%.cpp $(FSM_SRC)Makefile $(RUST_URL_LIB) compiler_flags
+builtins:
+	$Q mkdir -p builtins
+
+builtins/%.o: $(FSM_SRC)builtins/%.cpp $(FSM_SRC)Makefile $(RUST_URL_LIB) compiler_flags | builtins
```
elliottt, line 65: This fixes local builds, as when building in ci we pull down the wasm from a previous step and use that instead.

build.rs
```diff
@@ -34,6 +34,8 @@ fn build_engine(src: &PathBuf, out_dir: &Path) {
     let makefile_path = src.join("js-compute-runtime").join("Makefile");
     let status = Command::new("make")
         .current_dir(out_dir)
+        .arg("-j")
+        .arg("2")
```
elliottt, line 38: I'd like to pass `-j ${num_cpus}` here, and can do that using the `num_cpus` crate. Cargo already tracks this via the `cargo.jobs` config variable, but it's not clear to me how to get the value of that variable from the context of a build script.

elliottt: The `NUM_JOBS` env var is what I was looking for here.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

c-dependencies/js-compute-runtime/Makefile
jameysharp: Is it desirable to build in `$OUT_DIR` when building locally? Or is this a sign that `build.rs` should change? I don't have an opinion either way, I just wonder about things sometimes.

jameysharp (approved):
I'm a little surprised to not see any addition of `-j` for the CI builds at the same time, but regardless, I'm happy to sign off on this.

I think we've discussed replacing this makefile with a Meson-generated Ninja build script, which would have the nice property of using parallel builds by default without any command-line flags. But this patch is fine too :laughing:

c-dependencies/js-compute-runtime/Makefile
elliottt: That's a point, we could use the presence of `$OUT_DIR` to detect when we're building from build.rs, or perhaps have a default that sets it to the path of the makefile when it's missing. That seems like a reasonable change, but probably we should just take the time to switch this over to meson as you mentioned :)

Issue #32, opened 2021-08-25
Missing SharedArrayBuffer implementation


JakeChampion:
When building a project which makes use of `SharedArrayBuffer`, this error is thrown:
>Exception while evaluating JS: (new ReferenceError("SharedArrayBuffer is not defined", "<stdin>", 622))

I've worked around the issue in my project with a very hacky patch (`globalThis.SharedArrayBuffer = ArrayBuffer;`)

I think we should add a `SharedArrayBuffer` implementation as it is used within some packages such as the `data-urls` package which is used in reference/example code on the Fastly Docs website -- https://developer.fastly.com/learning/compute/migrate/#build-an-image-response

tschneidereit:
This one's a bit tricky: we don't have a way to actually implement `SharedArrayBuffer`, and pretending that we support it while we don't means that any code that feature-detects them to actually use them would end up doing the wrong thing.

The issue you mention is due to the webidl-conversions library [assuming SAB support](https://github.com/jsdom/webidl-conversions/issues/31#issuecomment-629668652). While that'd be trivial to fix in webidl-conversions, they made clear that they're not interested, as per the comment I linked.

(One way to fix it would be to change the relevant lines to something like
```js
const sabByteLengthGetter =
    typeof SharedArrayBuffer == "undefined" ? null : Object.getOwnPropertyDescriptor(SharedArrayBuffer.prototype, "byteLength").get;
```

That'd cause `isSharedArrayBuffer` to always return false, because trying to call a `null` value throws.)

guybedford:
Moving to https://github.com/bytecodealliance/StarlingMonkey/issues/115.

PR #160, opened 2022-07-20
Add implementation of DecompressionStream

JakeChampion:
This is an implementation of the WICG DecompressionStream builtin from https://wicg.github.io/compression/#decompression-stream

We already have CompressionStream, this is the counterpart to that.

An example c@e application which uses the feature would be:
```js
async function app(event) {
    const res = await fetch('https://httpbin.org/deflate', {backend: "httpbin"})
    return new Response(res.body.pipeThrough(new DecompressionStream('deflate')))
}

    event.respondWith(app(event))
})
```

Running the above program in viceroy and sending it a request should respond with something like this:
```
❯ http http://127.0.0.1:7878/
HTTP/1.1 200 OK
Date: Wed, 20 Jul 2022 16:15:54 GMT
Transfer-Encoding: chunked

{
  "deflated": true,
  "headers": {
    "Content-Length": "0",
    "Host": "httpbin.org",
    "X-Amzn-Trace-Id": "Root=1-62d82a3a-2cd522527318b751322fcc8d"
  },
  "method": "GET",
  "origin": "51.6.241.60"
}
```

Tests for this will be added in a future commit


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3322,6 +3322,340 @@ JSObject *create_rs_proxy(JSContext *cx, HandleObject input_readable) {
...
+
+  RootedObject controller(cx, TransformStream::controller(transform(self)));
+
+  if (finished) {
+    // Step 2 of flush:
+    // 2.  If the end of the compressed input has not been reached, then throw a TypeError.
+    if (zstream->avail_out != 0) {
+      JS_ReportErrorNumberUTF8(cx, GetErrorMessage, nullptr, JSMSG_DECOMPRESSING_ERROR);
+    }
+  }
```
JakeChampion, line 3420:
```suggestion
    zstream->avail_in = 0;
    zstream->next_in = nullptr;
    
    // Step 2 of flush:
    // 2.  If the end of the compressed input has not been reached, then throw a TypeError.
    if (zstream->avail_out != 0) {
      JS_ReportErrorNumberUTF8(cx, GetErrorMessage, nullptr, JSMSG_DECOMPRESSING_ERROR);
    }
  }
  
    RootedObject controller(cx, TransformStream::controller(transform(self)));
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3322,6 +3322,341 @@ JSObject *create_rs_proxy(JSContext *cx, HandleObject input_readable) {
...
+
+  RootedObject controller(cx, TransformStream::controller(transform(self)));
+
+  // Steps 3-5 of transform are identical to steps 3-5 of flush, so numbers
+  // below refer to the former for those. Also, the compression happens in
+  // potentially smaller chunks in the `do` loop below, so the three steps are
+  // reordered and somewhat intertwined with each other.
+
+  uint8_t *buffer = output_buffer(self);
```
elliottt, line 3426: Are private values not managed by the spidermonkey gc, and as a result don't need to be rooted?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3322,6 +3322,341 @@ JSObject *create_rs_proxy(JSContext *cx, HandleObject input_readable) {
...
+
+enum class Format {
+  GZIP,
+  Deflate,
+  DeflateRaw,
+};
+
+// Using the same fixed encoding buffer size as Chromium, see
+// https://chromium.googlesource.com/chromium/src/+/457f48d3d8635c8bca077232471228d75290cc29/third_party/blink/renderer/modules/compression/deflate_transformer.cc#29
+const size_t BUFFER_SIZE = 16384;
```
elliottt, line 3344: Might as well make this a `constexpr` :)

```suggestion
constexpr size_t BUFFER_SIZE = 16384;
```

elliottt (approved):
This looks good to me! Two questions:

1. We talked about moving the implementation of `DecompressionStream` into the `builtins` directory. Do you want to do that on this PR, or do that refactoring in a separate one?
2. It looks like some of the new expectations show failures, are those expected?

JakeChampion:
> * We talked about moving the implementation of `DecompressionStream` into the `builtins` directory. Do you want to do that on this PR, or do that refactoring in a separate one?

I had automatic merging enabled on this pull-request which meant it got merged as soon an the review approval happened. I've opened subsequent pull-requests to extract into the builtins directory 👍 

> * It looks like some of the new expectations show failures, are those expected?

Any tests which are to do with SharedArrayBuffers or any other shared typed-array we expect to fail due to not having an implementation of SharedArrayBuffer (https://github.com/fastly/js-compute-runtime/issues/32)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: Nice, I've done that in a different pull-request: https://github.com/fastly/js-compute-runtime/pull/176/files#diff-c73cfed8b701bf02df88d204faa554410b6c531dac7a5b9c2109dc66506d3398R33

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I assume they are not managed, my assumption is based upon the fact that our CompressionStream implementation does the same thing - https://github.com/fastly/js-compute-runtime/blob/890b51709bedadd857d3e7e332d73061a7363fb1/c-dependencies/js-compute-runtime/js-compute-builtins.cpp#L3762

PR #164, opened 2022-07-21
Throw TypeError instead of Error if constructor called as regular function

JakeChampion:
Section 3.7.1 of the WebIDL (https://webidl.spec.whatwg.org/#interface-object) specifies in step 1.2 that if a WebIDL interface is called with `NewTarget` undefined (i.e. not as a constructor) then to throw a TypeError. Our implementation was throwing an Error and not a TypeError.

This commit updates our implementation to throw a TypeError with the same message format that is used within Firefox.


elliottt (approved):
Nice!

PR #165, opened 2022-07-21
Throw a TypeError instead of an Error when value_to_buffer is handed a value which is not a buffer

JakeChampion:
Step 1 of the [compress and enqueue a chunk algorithm](https://wicg.github.io/compression/#compress-and-enqueue-a-chunk) states:
> If chunk is not a [BufferSource](https://webidl.spec.whatwg.org/#BufferSource) type, then throw a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

We currently throw an Error and not a TypeError. This pull-request updates our implementation to throw a TypeError.


elliottt (approved):
Looks great!

PR #167, opened 2022-07-22
Fix the error code from the wpt runner

elliottt:
This PR does two things:

* Fixes a bug in the wpt test runner that was considering crashes to be successful runs
* Fixes a bug exposed by #156 that manifested as a crash when a `ReadableStream` didn't yield a chunk that was a byte array

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1374,14 +1374,8 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
   if (!JS_GetProperty(cx, chunk_obj, "value", &val))
     return false;
 
+  // The read operation returned something that's not a Uint8Array
   if (!val.isObject() || !JS_IsUint8Array(&val.toObject())) {
-    // TODO: check if this should create a rejected promise instead, so an
-    // in-content handler for unhandled rejections could deal with it. The read
-    // operation returned a chunk that's not a Uint8Array.
-    fprintf(stderr, "Error: read operation on body ReadableStream didn't respond with a "
-                    "Uint8Array. Received value: ");
-    dump_value(cx, val, stderr);
-
     // reject the request promise
     if (Request::is_instance(body_owner)) {
       RootedObject response_promise(cx, Request::response_promise(body_owner));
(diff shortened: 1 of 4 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1377,6 +1381,22 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
...
+      RootedValue exn(cx);
+
+      // TODO: this should be a TypeError, but I'm not sure how to make that work
+      JS_ReportErrorUTF8(cx, "TypeError");
+      if (!JS_GetPendingException(cx, &exn)) {
+        return false;
+      }
+      JS_ClearPendingException(cx);
+
+      return JS::RejectPromise(cx, response_promise, exn);
```
elliottt, line 1397: @JakeChampion I'm not sure if this is the best way to construct the error, and the wpt test suggests that we should be raising a `TypeError` at this point. Do you happen to know the right error construction function to make a `TypeError`, and is it possible to construct an error without moving it out of the pending exception slot?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1377,6 +1381,22 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
     fprintf(stderr, "Error: read operation on body ReadableStream didn't respond with a "
```
jameysharp, line 1381: I guess you should edit this comment while you're at it.

elliottt: I changed the comment to reflect the open question about handling Responses that have failed to read a `Uint8Array`, and moved the error printing to below the promise rejection for a failed request.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: To raise a different error class, the JS_ReportErrorNumber* functions need to be used (https://searchfox.org/mozilla-central/source/js/src/jsapi.cpp#3544-3627)

let's change this in another pull-request to typeerror as we have a separate pull-request which creates our own error definitions already -> https://github.com/fastly/js-compute-runtime/pull/164

tests/wpt-harness/run-wpt.mjs
```diff
@@ -213,7 +216,7 @@ async function run() {
 
     if (config.tests.updateExpectations) {
       console.log(`Expectations updated: ${expectationsUpdated}`);
-    } else if (stats.unexpectedFail + stats.unexpectedPass != 0) {
+    } else if (stats.unexpectedFail + stats.unexpectedPass != 0 || unexpectedFailure) {
```
JakeChampion, line 219: nice!

JakeChampion (approved):

PR #168, opened 2022-07-22
Revert "fix: Mark requests as pending after their body stream has closed"

elliottt:
Reverts fastly/js-compute-runtime#156

#156 caused one of the WPT tests to start failing. #167 will prevent CI from passing with these errors present in the future.


elliottt:
I found the problem, and have a candidate fix on #167

PR #169, opened 2022-07-22
Set the host-provided response statusText based upon the provided status code

JakeChampion:
Currently there is no way to retrieve the reason-phrase (status text) for the host-provided response. Until such a way exists, we can use the definition mapping from IANA https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

Previously we were setting the statusText to `""` - with this change, the statusText will be a standard constant from IANA if one exists and if not, it will then be `""`.


elliottt (approved):
Awesome!

PR #175, opened 2022-07-27
 extract transform-stream and transform-stream-default-controller into their own files

JakeChampion:
This is a stacked PR - stacking on top of https://github.com/fastly/js-compute-runtime/pull/174


elliottt (approved):
Awesome!

PR #177, opened 2022-07-27
extract compression-stream into its own files

JakeChampion:
This is a stacked PR - stacking on top of https://github.com/fastly/js-compute-runtime/pull/176


elliottt (approved):
Awesome, thank you for extracting this!

PR #182, opened 2022-07-29
convert test fixtures from typescript to javascript

JakeChampion:
Currently nearly all of the test fixtures are written in typescript - which relies upon the assumption that the typescript compiler emits code which is implemented in a similar way to how we have written it in the fixtures. It also means we need to download webpack, ts-loader, and typescript and run them on each fixture.

If we convert the fixtures from typescript to javascript,  we can know for certain that the code being executed is the same code which is written in the fixtures. It also means we can remove the webpack/typescript step for compiling fixtures as the fixtures are now in javascript and do not require a compiler, which should speed up our CI runs a bit ☺️


elliottt (approved):
Seems reasonable to me, though do we have anything that exercises the types we export in sdk/js-compute now?

JakeChampion:
> Seems reasonable to me, though do we have anything that exercises the types we export in sdk/js-compute now?

We won't for a moment but I do plan to add a specific test file which exercises all of our types with https://github.com/SamVerschueren/tsd or similar

PR #183, opened 2022-07-29
Make each fixture in charge of how it gets built by the js-compute-runtime cli

JakeChampion:
this will help us in the future if/when we decide to add flags to the cli which would enable/disable features within the runtime such as experimental built-ins or early stage proposals from ecma262/whatwg/w3c


.github/workflows/main.yml
```diff
@@ -303,11 +303,10 @@ jobs:
...
 
-    - name: Install Fastly CLI
-      run: |
-        echo "Install Fastly CLI 1.7.0"
-        wget https://github.com/fastly/cli/releases/download/v1.7.0/fastly_1.7.0_linux_amd64.deb
-        sudo apt install ./fastly_1.7.0_linux_amd64.deb
+    - name: Set up Fastly CLI
+      uses: fastly/compute-actions/setup@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
elliottt, line 309: Nice!

elliottt (approved):
Looks great!

Issue #184, opened 2022-07-29
Exercises the types we export in sdk/js-compute/index.d.ts


JakeChampion:
We should add a specific test file which exercises all of our types with https://github.com/SamVerschueren/tsd or similar

_Originally posted by @JakeChampion in https://github.com/fastly/js-compute-runtime/issues/182#issuecomment-1199644630_

PR #186, opened 2022-08-04
update our published types and add tests for the types

JakeChampion:
Resolves https://github.com/fastly/js-compute-runtime/issues/184
I've added types for:
- `onfetch`
- `fastly.baseURL`
- `fastly.defaultBackend`
- `CompressionStream`
- `DecompressionStream`
- `console.debug`

I've removed the types for `console.trace` as the method no longer exists.

I've also updated the types to the latest versions from https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts and commented out all the parts that we have not yet implemented such as `Response.redirect()` etc

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -11,5 +11,5 @@ jobs:
     steps:
     - uses: actions/dependency-review-action@v2
       with:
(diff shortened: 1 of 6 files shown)


elliottt (approved):
This looks great! Thank you for bringing this all up to date :D

PR #188, opened 2022-08-08
 allow arbitrary amount of parameters passed to the console methods 

JakeChampion:
This gets us closer to following the spec at https://console.spec.whatwg.org/#logger


c-dependencies/js-compute-runtime/builtins/console.cpp
```diff
@@ -4,12 +4,31 @@ namespace Console {
...
+
+  printf("%s: %s\n", prefix, message.c_str());
+
+
+
+
+
+
+
+
```
JakeChampion, line 31:
```suggestion
```

PR #200, opened 2022-08-10
Move URL related namespaces out of js-compute-builtins.cpp 

JakeChampion:


elliottt (changes requested):
This looks good to me, but before we merge it do you know why those tests started passing?

JakeChampion:
> This looks good to me, but before we merge it do you know why those tests started passing?

They actually started failing - I've opened a pull-request to make the test results a bit more readable by using 'PASS' and 'FAIL' instead of '0' and '1'.
They are failing because in this bit of work I handled a conflict incorrectly and lost the newer version of `ThrowIfNotConstructing` which threw a TypeError. I'm rebasing this pull-request now to include the correct version of `ThrowIfNotConstructing` 👍

elliottt (approved):

PR #203, opened 2022-08-15
make the wpt test results say 'pass' and 'fail' instead of 0 and 1

JakeChampion:
I keep misremembering whether 0 or 1 meant 'pass' when I am reading the test results . Instead of trying to remember that I've changed the test runner to emit 'PASS' and 'FAIL' instead


elliottt (approved):
Awesome! :tada:

PR #204, opened 2022-08-15
feat: Improve console output for all types

JakeChampion:
printing an instance of any of the above classes would output '[object Object]' or similar, which is not ideal

by using `JS_ValueToSource` we instead print out a more usable output such as `{a:1, b:2, c:3}` for the object `{a:1, b:2, c:3}

Remaining classes to improve:
- [x] Set
- [x] Map
- [x] WeakSet
- [x] WeakMap
- [x] Proxy
- [x] Int8Array
- [x] Uint8Array
- [x] Uint8ClampedArray
- [x] Int16Array
- [x] Uint16Array
- [x] Int32Array
- [x] Uint32Array
- [x] Float32Array
- [x] Float64Array
- [x] BigInt64Array
- [x] BigUint64Array
- [x] Promise
- [x] Response
- [x] Request
- [x] ReadableStream
- [x] WritableStream
- [x] TransformStream


c-dependencies/js-compute-runtime/builtins/console.cpp
```diff
@@ -1,26 +1,330 @@
...
+  case JS::PromiseState::Pending: {
+    message += "<pending> }";
+    break;
+  }
+  case JS::PromiseState::Fulfilled: {
+    JS::RootedValue value(cx, JS::GetPromiseResult(obj));
+    auto value_source = ToSource(cx, value, visitedObjects);
+    if (value_source.isErr()) {
+      return JS::Result<std::string>(JS::Error());
+    }
```
elliottt, line 28: What do you think about using `MOZ_TRY_VAR` here instead?
```suggestion
    auto value_source = MOZ_TRY_VAR(ToSource(cx, value, visitedObjects));
```

elliottt (approved):
This looks great!

How would you feel about passing a `std::string &` argument through the various `ToSource` functions and appending to that, instead of returning new strings for each step?

c-dependencies/js-compute-runtime/builtins/console.cpp
JakeChampion: I first moved to MOZ_TRY_VAR but when I did the performance optimisation you suggested for appending to an already existing string, it involved changing these to MOZ_TRY instead of MOZ_TRY_VAR as the functions no longer return strings and instead append in-place

JakeChampion:
> This looks great!
> 
> How would you feel about passing a `std::string &` argument through the various `ToSource` functions and appending to that, instead of returning new strings for each step?

I like it, I have done this in [`aff304a` (#204)](https://github.com/fastly/js-compute-runtime/pull/204/commits/aff304a882621d91fd8aa98a2b0cab3dcb3b7f34) and it is working ☺️

Issue #201, opened 2022-08-10
JS_GC trap issue.


wilsonwang371:
Hi,

I am using this repo and execute the runtime on wasmtime. 

What I observed is that 1 out of 3 times if I call JS_GC at the end of the function that I called, js-compute-runtime can trigger a HeapOutOfBound trap in wasmtime.

Is there any suggestions on this why this is happening?

Thanks

alekhanetweb:
@wilsonwang371 sorry to comment on an old issue ...but how does one execute the runtime on wasmtime ? im looking to try and do the same

PR #208, opened 2022-08-19
fix potential memleak in reactor mode

wilsonwang371:
In reactor mode, the main function can get called multiple times. `__wasilibc_initialize_environ()` will call malloc each time without verifying if environment has been initialized.

We need to call `__wasilibc_ensure_environ()` instead.

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-runtime.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-runtime.cpp
@@ -478,7 +478,7 @@ int main(int argc, const char *argv[]) {
   double total_compute = 0;
   auto start = system_clock::now();
 
-  __wasilibc_initialize_environ();
+  __wasilibc_ensure_environ();
 
   if (debug_logging_enabled()) {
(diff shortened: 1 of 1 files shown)


wilsonwang371:
this fixed the issue I reported in #201

wilsonwang371:
@elliottt can you give some suggestions?

elliottt:
This isn't a use-case that Fastly is currently trying to support for the js-compute-runtime; we never call the main function more than once and always re-initialize for new requests.

That said, it's not clear to me that `__wasilibc_ensure_environ` will interact well with wizer. Currently we run wizer when producing the runnable wasm blob which will run the `main` function. As this will take a snapshot of the memory, it might be updating the global that `__wasilibc_ensure_environ` uses to check that initialization has been done already. A way that you could validate this hypothesis would be to add your own logic to conditionally call `__wasilibc_initialize_environ` if `main` is being called for the first request and `isWizening` returns false. If that continues to crash during the SDK test part of CI, there's another problem, but my guess is that `__wasilibc_ensure_environ` is never calling `__wasilibc_initialize_environ` after wizening.

PR #227, opened 2022-08-24
feat: add btoa and atob native implementations

JakeChampion:
`atob` and `btoa` are global base64 utility methods which are useful to have included in the runtime as projects such as PlnaetScale's Serverless Driver for JavaScript make use of these methods and a native implementation should be more performant than a JavaScript implementation (a polyfill)

Note: I haven't figured out how to raise an exception with a custom error class, which is required for full adherence with the specification (and to make all the web-platform-tests pass) as the function should throw InvalidCharacterError instances but we throw Error instances instead.

https://html.spec.whatwg.org/multipage/webappapis.html#atob

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1,9 +1,11 @@
 #include <arpa/inet.h>
 #include <iostream>
+#include <regex> // std::regex
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include <vector>
 
 #include "js-compute-builtins.h"
 #include "rust-url/rust-url.h"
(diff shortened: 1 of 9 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3702,6 +3704,382 @@ bool init_class(JSContext *cx, HandleObject global) {
...
+  while (inputLength >= 4) {
+    auto out_result = base64Decode4to3(input);
+    if (out_result.isErr()) {
+      return JS::Result<std::string>(JS::Error());
+    }
+    output += out_result.unwrap();
+
+    input.erase(0, 4);
+    inputLength -= 4;
+  }
```
elliottt, line 3915: What do you think about restructuring this loop to operate on a `std::string_view` instead? This way you won't incur any cost for dropping data off of the front of the buffer like you might with `erase`, and you can ensure that none of the decoding functions will modify your buffer. You'll also avoid copying the input buffer when calling the decoding functions, as `std::string_view` is very cheap to copy around.
```suggestion
  std::string_view buffer = input;                                                   
  while (buffer.size() >= 4) {                                                       
    auto out_result = base64Decode4to3(input.substr(0, 4));                          
    if (out_result.isErr()) {                                                        
      return JS::Result<std::string>(JS::Error());                                   
    }                                                                                
    output += out_result.unwrap();                                                   
                                                                                     
    buffer = buffer.substr(5);                                                       
  } 
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3702,6 +3704,382 @@ bool init_class(JSContext *cx, HandleObject global) {
...
+// forgiving-base64 encode for symmetry with forgiving-base64 decode, which is different from the
+// RFC as it defines error handling for certain inputs.
+std::string forgivingBase64Encode(std::string data) {
+  int length = data.length();
+  std::string out = "";
+  while (length >= 3) {
+    out += base64Encode3to4(data);
+    data.erase(0, 3);
+    length -= 3;
+  }
```
elliottt, line 4031: Again, I'd refactor this to use a `std::string_view` so that you're not copying the buffer around while encoding it, and not paying any cost for moving the pointer at the front of `input` around.

elliottt (changes requested):
This is looking great, and that test suite is amazing!

I had some suggestions about how you can avoid copying the strings around in the encode/decode functions, which I think will help ease allocation a bit :)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: Great suggestion - this makes complete sense, string_view is the perfect type for this ☺️ 
I made this change in a single commit alongside updating all the decode functions to use string_view - [`54c2eec` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/54c2eecdd20cfd2d1e2fb43b8efbfb2552f6fe05)

I'll do the same change for the encoding functions as well 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3702,6 +3704,382 @@ bool init_class(JSContext *cx, HandleObject global) {
...
+  /* 96 */ 255, /* a - z */ 26, 27, 28, 29, 30, 31, 32,
+  /* 104 */ 33, 34, 35, 36, 37, 38, 39, 40,
+  /* 112 */ 41, 42, 43, 44, 45, 46, 47, 48,
+  /* 120 */ 49, 50, 51, 255, 255, 255, 255, 255,
+};
+// clang-format on
+
+bool base64CharacterToValue(char character, uint8_t *value) {
+  static const size_t mask = 0x7f;
+  size_t index = static_cast<uint8_t>(character);
```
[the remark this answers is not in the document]
JakeChampion: also done in [`4d54091` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/4d5409173d40758bdd56890dfa2f521a5de29da4) :D

JakeChampion:
@elliottt this is ready for another round of reviews 👍  I really like where this has ended up implementation-wise ☺️

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1,16 +1,19 @@
 #include <arpa/inet.h>
 #include <iostream>
+#include <regex> // std::regex
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include <vector>
```
elliottt, line 8: Looks like this header isn't used?
```suggestion
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1,16 +1,19 @@
 #include <arpa/inet.h>
 #include <iostream>
+#include <regex> // std::regex
```
elliottt, line 3: Looks like this header isn't used?
```suggestion
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+  /* 80 */ 15, 16, 17, 18, 19, 20, 21, 22,
+  /* 88 */ 23, 24, 25, 255, 255, 255, 255, 255,
+  /* 96 */ 255, /* a - z */ 26, 27, 28, 29, 30, 31, 32,
+  /* 104 */ 33, 34, 35, 36, 37, 38, 39, 40,
+  /* 112 */ 41, 42, 43, 44, 45, 46, 47, 48,
+  /* 120 */ 49, 50, 51, 255, 255, 255, 255, 255,
+};
+// clang-format on
+
+bool base64CharacterToValue(char character, uint8_t *value) {
```
elliottt, line 3719: You could make this a reference instead, and skip the `*` when assigning to it on line 3726, as well as the `&` in the uses later on.
```suggestion
bool base64CharacterToValue(char character, uint8_t &value) {
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+}
+
+// https://infra.spec.whatwg.org/#forgiving-base64-decode
+JS::Result<std::string> forgivingBase64Decode(std::string_view data) {
+  // 1. Remove all ASCII whitespace from data.
+  // ASCII whitespace is U+0009 TAB, U+000A LF, U+000C FF, U+000D CR, or U+0020 SPACE.
+  auto hasWhitespace = std::find_if(data.begin(), data.end(), &isAsciiWhitespace);
+  std::string dataWithoutAsciiWhitespace;
+
+  if (hasWhitespace) {
```
elliottt, line 3803: `hasWhitespace` will be an iterator, so you'll want to compare explicitly against `data.end()` to check if there's whitespace.
```suggestion
  if (hasWhitespace != data.end()) {
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+    dataWithoutAsciiWhitespace = data;
+    dataWithoutAsciiWhitespace.erase(std::remove_if(dataWithoutAsciiWhitespace.begin(),
+                                                    dataWithoutAsciiWhitespace.end(),
+                                                    &isAsciiWhitespace),
+                                     dataWithoutAsciiWhitespace.end());
```
elliottt, line 3808: Since you already know how far in the whitespace was inside the string, you can reuse the offset into the copy. Also, this will remove from the first occurrence of whitespace to the end of the string, is that correct?
```suggestion
    dataWithoutAsciiWhitespace.erase(dataWithoutAsciiWhitespace.begin() + std::distance(data.begin(), hasWhitespace),
                                     dataWithoutAsciiWhitespace.end());
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+    data = dataWithoutAsciiWhitespace;
+  }
+  std::string_view data_view(data);
```
elliottt, line 3811: Is `data_view` necessary? Since you've already updated `data` to point at your temporary copy, it seems like you could just use that everywhere instead.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+
+  // 7. Let position be a position variable for data, initially pointing at the start of data.
+
+  // We don't use a position variable, instead we remove_prefix from the `data` each time we have
+  // dealt with some characters.
+
+  while (data_view.length() >= 4) {
+    MOZ_TRY(base64Decode4to3(data_view, output));
+    data_view.remove_prefix(4);
+  }
```
elliottt, line 3857: Nice!

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3625,6 +3630,388 @@ bool response_started(JSObject *self) {
...
+
+  for (i = 0; i < 4; ++i) {
+    output += base[(uint32_t)((b32 >> j) & 0x3F)];
+    j -= 6;
+  }
+}
+
+inline void base64Encode2to4(std::string_view data, std::string &output) {
+  uint8_t src0 = CharTo8Bit(data[0]);
```
elliottt, line 3930: I like `static_cast` with an `auto` result for cases like these, what do you think?
```suggestion
  auto src0 = static_cast<uint8_t>(data[0]);
```

elliottt (commented):
Just a few suggestions about iterator usage, I think we can probably get this merged in Friday :tada:

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I somehow lost that as part of a refactor - I previously had it as part of the definition of `hasWhitespace` itself so that it was in fact a boolean. Seeing your other suggestion on line 3805 where we can use the position of the first found whitespace to skip iterating the elements prior to it is a great reason to keep the iteration position around

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: This should be only removing elements which return true from the `&isAsciiWhitespace` function that we pass in to `remove_if`

`remove_if` does not actually remove elements but instead moves all the elements that return false for the `&isAsciiWhitespace` function to the front of the range, leaving all the elements which return true for `&isAsciiWhitespace` at the end of the range, this is done in a single pass which is great. remote_if will then return an iterator which points to the first element that returned true for `&isAsciiWhitespace`. We then pass that returned iterator as the first parameter to `erase`, and `erase` removes all the ascii-whitespace elements in one call.

When we move to c++20 we can change from using this ['erase-remove idiom'](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) to using std::remove_if instead 😄

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I like it, I applied this suggestion as part of [`ae81dca` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/ae81dca453df75f519a3af68046348497954ab0e)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: It is not necessary! I applied this suggestion as part of [`ae81dca` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/ae81dca453df75f519a3af68046348497954ab0e)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I applied this suggestion as part of [`ae81dca` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/ae81dca453df75f519a3af68046348497954ab0e)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: nice, yes! I applied this suggestion as part of [`ae81dca` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/ae81dca453df75f519a3af68046348497954ab0e)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I removed this as part of [`ae81dca` (#227)](https://github.com/fastly/js-compute-runtime/pull/227/commits/ae81dca453df75f519a3af68046348497954ab0e)

integration-tests/js-compute/fixtures/btoa/bin/index.js
```diff
@@ -0,0 +1,352 @@
...
+      error = assert(btoa(42), btoa('42'));
+      if (error) { return error }
+      error = assert(btoa(null), btoa('null'));
+      if (error) { return error }
+      error = assert(btoa({ x: 1 }), btoa('[object Object]'));
+      if (error) { return error }
+
+      error = assertThrows(() => btoa(), TypeError);
+      if (error) { return error }
+      error = assertThrows(() => btoa('🐱'));
```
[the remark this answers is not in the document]
JakeChampion: I'll use 😿 instead ha
```suggestion
      error = assertThrows(() => btoa('😿'));
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
elliottt: Sorry, I totally missed the erase/remove_if combination there, and thought it was just erasing to the end of the string!

elliottt (approved):
This looks great!

PR #228, opened 2022-08-24
docs: Add api documentation for the ObjectStore and ObjectStoreEntry implementations

JakeChampion:


sdk/js-compute/index.d.ts
```diff
@@ -306,6 +306,76 @@ export declare class Env {
...
+  constructor(name: string);
+  /**
+   * Gets the value associated with the key `key` in the Object-store.
+   * When the key is present, a resolved Promise containing an ObjectStoreEntry will be returned which contains the associated value.
+   * When the key is absent, a resolved Promise containing null is returned.
+   * @param key The key to retrieve from within the Object-store. A key cannot:
+   * - Be any of the strings "", ".", or ".."
+   * - Start with the string ".well-known/acme-challenge/""
+   * - Contain any of the characters "#?*[]\n\r"
+   * - Be longer than 1024 characters
```
elliottt, line 331: Is this still true? The comment in object-store.cpp says that they must be less than 255 chars.

elliottt (approved):
Looks good to me, just a question about the maximum length of names.

sdk/js-compute/index.d.ts
JakeChampion: This is still true as it is for a store key and not the store name - https://github.com/fastly/js-compute-runtime/blob/a6ab7ac951753864a922a48a8ec89ecd4f852a4f/c-dependencies/js-compute-runtime/builtins/object-store.cpp#L125

PR #229, opened 2022-08-24
update our api website theme

JakeChampion:
The version of typedoc we are using no longer has the minimal theme as it is very out of date and buggy.

I've installed a new theme which is similar to minimal but offers some extra features such as search.

I've also fixed some broken internal linking we had in the api docs


elliottt (approved):
:tada:

PR #250, opened 2022-09-01
feat: Add Dynamic Backends support

JakeChampion:
feat: Add Dynamic Backends support

Dynamic Backends allows JavaScript projects to dynamically create new backend server definitions without having to deploy a new version of their Fastly Service. These backends function exactly as existing backends, and will _eventually_ be configurable in all the same ways as existing backends can via the Fastly Service defintion. 

The way I've implemented this feature is to have it only enabled when the property `fastly.allowDynamicBackends` is set to `true`. The reason for that is to try to make it a bit more obvious that Dynamic Backends is not the default approach for a fetch.  (It's not the default approach as it can be a potential avenue for third-party JavaScript code within a project to send requests, potentially including sensitive/secret data, off to destinations that the JavaScript project was not intending, which could be a security issue.) The implementation of the property `fastly.allowDynamicBackends` doesn’t _yet_ gain any security advantages because a third-party dependency has access to the `fastly` global variable and can set `fastly.allowDynamicBackends = true` — there would be a security advantage when we have support for ES Modules and can then limit this property to only be writable from the top-level script and readable from other ES Modules.

There are two ways to make use of Dynamic Backends within JavaScript projects.
The first way is by omitting the `backend` property definition on the Request instance. The JavaScript Runtime will then create a Dynamic Backend definition using default configuration options. This approach is useful for JavaScript applications as it means that a standard `fetch` call will now be possible, which means libraries that use `fetch` should just work 🥳.

E.G:
```js
// Enable dynamic backends -- warning, this is potentially dangerous as third-party dependencies could make requests to their own backends, potentially including your sensitive/secret data
fastly.allowDynamicBackends = true;

// For any request, return the fastly homepage -- without defining a backend!
  event.respondWith(fetch('https://www.fastly.com/'));
});
```

The second way is by creating a new Dynamic Backend using the JavaScript `Backend` class. This approach is useful for JavaScript applications that want to have control over the configuration of the new backend defintion, such as only allowing TLS 1.3 and disallowing older versions of TLS. The `Backend` class has a `toString` implementation which will return it's name, this means it can be attached to a Request instance.
E.G.
```js
// Enable dynamic backends -- warning, this is potentially dangerous as third-party dependencies could make requests to their own backends, potentially including your sensitive/secret data
fastly.allowDynamicBackends = true;

// For any request, return the fastly homepage -- without defining a backend!
  // We are defining all the possible fields here but any number of fields can be defined - the ones which are not defined will use their default value instead.
  const backend = new Backend({
    name: 'fastly',
    target: 'fastly.com',
    hostOverride: "www.fastly.com",
    connectTimeout: 1000,
    firstByteTimeout: 15000,
    betweenBytesTimeout: 10000,
    useSSL: true,
    sslMinVersion: 1.3,
    sslMaxVersion: 1.3,
    certificateHostname: "www.fastly.com",
    checkCertificate: true,
    caCertificate: `-----BEGIN CERTIFICATE-----
MIIGgjCCBWqgAwIBAgIQASW9vtgUNIzM07bVg9HdEzANBgkqhkiG9w0BAQsFADBY
MQswCQYDVQQGEwJCRTEZMBcGA1UEChMQR2xvYmFsU2lnbiBudi1zYTEuMCwGA1UE
AxMlR2xvYmFsU2lnbiBBdGxhcyBSMyBEViBUTFMgQ0EgMjAyMiBRMTAeFw0yMjAz
MzExNDI4NTBaFw0yMzA1MDIxNDI4NDlaMBkxFzAVBgNVBAMMDnd3dy5mYXN0bHku
Y29tMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAmtLSpTiRzJaFiAEi
Eua36KwkZBm647y7cEkYEu6GgStXAWtzDF4ttzI8q9iRpqlljFDjQF2EeWLXN2X2
MEQPxKOOkM+OuJny0pXDCkKiSGS6MwVzNqCNW5tnxkJMMUAq5ciyQFwF72Z+ymlo
NGCl3CIDAJbHcudm4CiVR5ip7KYFH7uaBRLPXtMLX2lmQ1q2TbU/GXrbDhz5OXCD
H9SKAUSHLqwLplj8mojxWayda9zy3bqbTVXsqRVW0Xar62T33Uis0fm+xW4hJbxf
bya5I0aQbAjtxvEjfUa1kSalUqPwfHOXMvvXwoqUIKk1ndZScWl9TGLH84izTCPw
bOVsTQIDAQABo4IDhTCCA4EwOwYDVR0RBDQwMoIOd3d3LmZhc3RseS5jb22CFGRl
dmVsb3Blci5mYXN0bHkuY29tggpmYXN0bHkuY29tMA4GA1UdDwEB/wQEAwIFoDAd
BgNVHSUEFjAUBggrBgEFBQcDAQYIKwYBBQUHAwIwHQYDVR0OBBYEFCYC4rPzDtR8
EI3XULw/wr8lfnv2MFcGA1UdIARQME4wCAYGZ4EMAQIBMEIGCisGAQQBoDIKAQMw
NDAyBggrBgEFBQcCARYmaHR0cHM6Ly93d3cuZ2xvYmFsc2lnbi5jb20vcmVwb3Np
dG9yeS8wDAYDVR0TAQH/BAIwADCBngYIKwYBBQUHAQEEgZEwgY4wQAYIKwYBBQUH
MAGGNGh0dHA6Ly9vY3NwLmdsb2JhbHNpZ24uY29tL2NhL2dzYXRsYXNyM2R2dGxz
Y2EyMDIycTEwSgYIKwYBBQUHMAKGPmh0dHA6Ly9zZWN1cmUuZ2xvYmFsc2lnbi5j
b20vY2FjZXJ0L2dzYXRsYXNyM2R2dGxzY2EyMDIycTEuY3J0MB8GA1UdIwQYMBaA
FJwqZ9C0gzFbgMspllV1CTc4uDylMEgGA1UdHwRBMD8wPaA7oDmGN2h0dHA6Ly9j
cmwuZ2xvYmFsc2lnbi5jb20vY2EvZ3NhdGxhc3IzZHZ0bHNjYTIwMjJxMS5jcmww
ggF/BgorBgEEAdZ5AgQCBIIBbwSCAWsBaQB3AK33vvp8/xDIi509nB4+GGq0Zyld
z7EMJMqFhjTr3IKKAAABf+BhC9UAAAQDAEgwRgIhAI/TiJpulEywYHMuudQ6fYcK
dl18wDnJtAD+3JqFkrXuAiEA+i6ryPZTeENZJ6wEgRnndsggk8blsfch13e4s76u
WngAdQB6MoxU2LcttiDqOOBSHumEFnAyE4VNO9IrwTpXo1LrUgAAAX/gYQwqAAAE
AwBGMEQCIEldTkse3joAWr1llSoi9EOle0K0hz086GrCjL5YDhXkAiARW5dUit2q
Tq/mD+aN+XHfupYiYC7htPvHMWM6iUHPFgB3ALNzdwfhhFD4Y4bWBancEQlKeS2x
ZwwLh9zwAw55NqWaAAABf+BhDFQAAAQDAEgwRgIhAIQ1nuPDm08OuXDmLBkreA7L
LdGLmdhnJOdzOW0n7PDGAiEA4cEZYQ+uFrAoIHezQXBe05/ovXLacu5SVoInRJIK
gBYwDQYJKoZIhvcNAQELBQADggEBAPH+sQGtGBE6BlC0Zp02rJo/1QCPC+/L1T1+
uErHb0175NKnxXmT0HYoXr3gduMoMFXQr0VkhtQzCq7OaAX89+UdQ/OkfdU6CK0z
H4318J11ZShIHqGUCL+w27JsAgzJw/9UQBfT3FLCto735lwfo/l6HyD7qNqNYwhK
UlkETWIEFB5CUAvG5mVmNbnUI8fnF3CvEB7IJcX9DxIkfsS08+4lMi3jHFMthJqY
N5RGelXh07FAawRDugRPO0gh5bHo+hxZnWYTW9s6+A/D9E3pP2OJcyvhnvtWUrjg
rgzylmjRmAH2KP86mK9jRwFvTdn666JZFKzpubei6XjWVdG/eS8=
-----END CERTIFICATE-----`,
    // Colon-delimited list of permitted SSL Ciphers
    ciphers: "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256:!RC4",
    sniHostname: "www.fastly.com",
  });
  event.respondWith(fetch('https://www.fastly.com/', {
    backend // Here we are configuring this request to use the dynamically define backend from above.
  }));
});
```


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4036,10 +4037,10 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
   }
 
   RootedString backend(cx, Request::backend(request));
-  if (!backend) {
+  if (!backend && builtins::Fastly::allowDynamicBackends == false) {
```
elliottt, line 4040: What do you think about reordering these cases to one top-level check of `allowDynamicBackends`, with the different behavior for when `backend` is missing under those branches?

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,705 @@
...
+bool isLowerAlpha(char character) { return character >= 97 && character <= 122; }
+bool isUpperAlpha(char character) { return character >= 65 && character <= 90; }
+bool isAlphaNumeric(char character) {
+  if (!isDigit(character) && !isLowerAlpha(character) && !isUpperAlpha(character)) {
+    return false;
+  }
+  return true;
+}
+bool isDash(char character) { return character == 45; }
+bool isDot(char character) { return character == 46; }
```
elliottt, line 24: What do you think about switching to the functions in the `<cchar>` header instead? (There are already definitions for things like `std::is_digit` in there.)

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,705 @@
...
+    return true;
+  }
+
+  // check the character before the last . is in the regex [a-zA-Z0-9]
+  if (!isAlphaNumeric(hostname.at(lastDot - 1))) {
+    return false;
+  }
+  // check all other characters before the last . are in the regex [a-zA-Z0-9\-.]
+  auto last = std::next(hostname.begin(), lastDot);
+  auto it = std::find_if(std::next(hostname.begin()), last, isNotAlphaNumericDotOrDash);
```
elliottt, line 66: There's `std::find_if_not`, if it makes the predicates easier to think about. I'd also be tempted to use a lambda here, and let the compiler figure out that they're the same function.

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,705 @@
...
+  // ValidHostRegex =
+  // "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(:[0-9]+)$";
+  auto firstCharacter = host.front();
+  // check first character is in the regex [a-zA-Z0-9]
+  if (!isAlphaNumeric(firstCharacter)) {
+    return false;
+  }
+  // split the hostname from the port
+  int pos = host.find_first_of(':');
+  std::string hostname = host.substr(0, pos);
```
elliottt, line 46: This will allocate a new string, would it be better to parse a `std::string_view` instead?

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,705 @@
...
+  // check all characters after the last . (but not the last character) are in the regex
+  // [a-zA-Z0-9\-]
+  last = std::prev(hostname.end());
+  it = std::find_if(std::next(hostname.begin(), lastDot + 1), last, isNotAlphaNumericOrDash);
+  if (it != last) {
+    return false;
+  }
+  // if there is a port - confirm it is all digits and is between 0 and 65536
+  if (pos != std::string::npos) {
+    std::string port = host.substr(pos + 1);
```
elliottt, line 80: This will probably allocate a string, you could avoid it by using a `std::string_view`.

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,705 @@
...
+      return false;
+    }
+    if (firstByteTimeoutInt >= std::pow(2, 32)) {
+      JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr,
+                                JSMSG_BACKEND_FIRST_BYTE_TIMEOUT_TOO_BIG);
+      return false;
+    }
+    JS::SetReservedSlot(backend, Backend::Slots::FirstByteTimeout,
+                        JS::Int32Value(firstByteTimeoutInt));
+  }
```
elliottt, line 487: Would it be possible to extract this logic to a function? it looks pretty similar to the cases for `betweenBytesTimeout` and `connectTimeout`.

elliottt (commented):
This is looking good, and what a test suite for checking all the different options!

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: oh yes this is nice - I've now moved to using `<cctype>` and `std::isalnum` instead of our own implementation -- [`bb1abe2` (#250)](https://github.com/fastly/js-compute-runtime/pull/250/commits/bb1abe28117aa02c48e73dd6d3f54c32488c7893)

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: Yes it would - thank you for spotting this. I've scanned through this file and found a couple more places which could change from string to string_view and have made all the changes in [5fcb9dfdfa86026529ea933260a5c8936d9e922f](https://github.com/fastly/js-compute-runtime/pull/250/commits/5fcb9dfdfa86026529ea933260a5c8936d9e922f)

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: you spotted this one as well! 😄 This has changed to std::string_view in [`5fcb9df` (#250)](https://github.com/fastly/js-compute-runtime/pull/250/commits/5fcb9dfdfa86026529ea933260a5c8936d9e922f)

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: good idea 👍 This was refactored in [`4d0249f` (#250)](https://github.com/fastly/js-compute-runtime/pull/250/commits/4d0249f36f8b5b9103dd87161c019a201edd76d0)

JakeChampion:
@elliottt this is ready for another review now 🙇

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,855 @@
...
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_BACKEND_TARGET_NOT_SET);
+    return false;
+  }
+
+  auto targetStringSlice = encode(cx, target_val);
+  if (!targetStringSlice.data) {
+    return false;
+  }
+
+  std::string_view targetString((char *)targetStringSlice.data, targetStringSlice.len);
```
elliottt, line 474:
```suggestion
  std::string_view targetString(reinterpret_cast<char*>(targetStringSlice.data), targetStringSlice.len);
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,855 @@
...
+    return nullptr;
+  }
+
+  auto url = jsurl::new_jsurl(&url_string);
+  if (!url) {
+    JS_ReportErrorUTF8(cx, "URL constructor: %s is not a valid URL.", (char *)url_string.data);
+    return nullptr;
+  }
+  const jsurl::SpecSlice slice = jsurl::host(url);
+  std::string aaa((char *)slice.data, slice.len);
```
elliottt, line 512: Is `slice.data` not null-terminated? If it is you can avoid the copy when making the string. If not, I'd change the cast to a `reinterpret_cast` to make it clear that this is a copy.
```suggestion
  const jsurl::SpecSlice slice = jsurl::host(url);
  std::string aaa(reinterpret_cast<const char *>(slice.data), slice.len);
```

elliottt (commented):
This is looking good! I think my biggest suggestion is to avoid c-style casts whenever possible :D

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,855 @@
...
+  JS::UniqueChars nameChars = encode(cx, name, &name_len);
+  auto name_cstr = nameChars.get();
+
+  JS::RootedString target(cx, JS::GetReservedSlot(backend, Backend::Slots::Target).toString());
+  size_t target_len;
+  JS::UniqueChars targetChars = encode(cx, target, &target_len);
+  auto target_cstr = targetChars.get();
+  if (hostOverride.has_value()) {
+    backend_config_mask |= BACKEND_CONFIG_HOST_OVERRIDE;
+    definition.host_override = (char *)hostOverride.value().c_str();
```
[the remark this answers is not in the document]
JakeChampion: I've updated the `xqd.h` definition to be `const char *` 👍

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: I checked and it looks null-terminated to me 👍

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -0,0 +1,857 @@
...
+  size_t target_len;
+  JS::UniqueChars targetChars = encode(cx, target, &target_len);
+  auto target_cstr = targetChars.get();
+  if (hostOverride.has_value()) {
+    backend_config_mask |= BACKEND_CONFIG_HOST_OVERRIDE;
+    definition.host_override = hostOverride.value().c_str();
+    definition.host_override_len = hostOverride.value().length();
+  } else {
+    definition.host_override = nullptr;
+    definition.host_override_len = 0;
```
elliottt, line 341: I think all the else cases that are zeroing out the fields of `definition` should be unnecessary after adding the `memset` at the top.

elliottt (approved):
This looks great to me!

Before you merge, would you mind experimenting with removing the else branches that zero the memory of `definition` now that it's being memset at the function entry? I think that'll streamline the latter half of that function a bit :+1:

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: good catch 👍

Issue #251, opened 2022-09-02
Ambient types/objects not being declared correctly since 0.5


harmony7:
Since 0.5, i’m seeing some problems with the ambient types/objects provided by `index.d.ts` of the `js-compute` npm package.

In my IDE I'm getting no code completion.

When building a library, I'm getting messages such as the following:
```
TS2304: Cannot find name 'FetchEvent'.

8 export function addFetchEventAction(priority: number, fn: (event: FetchEvent) => void) {
                                                                    ~~~~~~~~~~
```
This indicates that the ambient types/objects are not available.
These problems happen in both 0.5.0 and 0.5.1, but not in 0.4.0 and before.

The cause is that `export` has been added to everything that is supposed to be simply `declare` (this was done in https://github.com/fastly/js-compute-runtime/commit/8b28f090fd5475f51d9b6bc433c225b3b975ecf7, part of #186).

By using `/// <reference types="@fastly/js-compute" />`, we are supposed to be “declaring” that these names are going to be available at runtime, _in some way_.  Putting `export` on these is wrong, because that would imply that this package would be exporting them, and that we would be using the types like this:
```
import { fastly, Request, Response, Headers } from '@fastly/js-compute';
```
but this is of course incorrect because the package doesn’t actually export these objects.

As for how to correct this:
Since in #186 you are "exporting" the types in order to write tests against them, I would suggest the following.

* Revert the change that added `export` to everything, and
* In the `index.test-d.ts` file, replace:
```
import {expectError, expectType} from 'tsd';
import { addEventListener, atob, btoa, CacheOverride, CacheOverrideInit, CacheOverrideMode, ClientInfo, CompressionStream, CompressionStreamFormat, Console, console, DecompressionStream, DecompressionStreamFormat, Dictionary, Env, EventMap, fastly, Fastly, FetchEvent, FetchEventListener, Geolocation, Logger, onfetch, ObjectStore, ObjectStoreEntry, ReadableStream, Request, Response, TextDecoder, TextEncoder, URL, URLSearchParams, WritableStream } from ".";
```
with:
```
/// <reference path="./index.d.ts" />
import {expectError, expectType} from 'tsd';
```
(the triple slash directive has to come first)
I tried this myself and the tests pass.

JakeChampion:
This was fixed in #252

PR #252, opened 2022-09-02
fix:  declare ambient types for the fastly/js-compute npm package

JakeChampion:
This fixes an issue in how we make our typescript types visible.

An excellent write-up of the situation can be found in this issue: https://github.com/fastly/js-compute-runtime/issues/251


harmony7 (approved):
LGTM!

PR #253, opened 2022-09-02
fix: explicitly declare void as the return type for functions which return nothing - to allow our package to work with typescript's `strict:true` option

harmony7:
This PR adds explicit `void` return types for functions that left them off.
Leaving them off causes problems when consuming `@fastly/js-compute` from a TypeScript project that has `strict: true` specified in `tsconfig.json`, with errors such as the following:

```
TS7010: 'log', which lacks return-type annotation, implicitly has an 'any' return type.
```

A workaround in the meantime would be to use `skipLibCheck: true`.

This was fixed at one time in https://github.com/fastly/compute-sdk-as-js/pull/126, but got overwritten because the file in this repo seems to be the source of truth (?).

### DIFF
--- a/sdk/js-compute/index.d.ts
+++ b/sdk/js-compute/index.d.ts
@@ -474,11 +474,11 @@ export declare class URLSearchParams {
  * **Note**: Messages are prefixed with the respective logel level, starting with an upper-case letter, e.g. `"Log: "`.
  */
 export declare interface Console {
-  log(message: any);
-  debug(message: any);
-  info(message: any);
-  warn(message: any);
-  error(message: any);
+  log(message: any): void;
+  debug(message: any): void;
+  info(message: any): void;
+  warn(message: any): void;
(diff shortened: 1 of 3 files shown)


JakeChampion:
Tnanks @harmony7

> `strict: true` specified in `tsconfig.json`, with errors such as the following:

Is there a way for this project to also enable `strict: true` as to ensure we don't regress on this functionality in the future?

harmony7:
Yes. Go ahead and change `tsconfig.json` to have `strict: true`, and then that `yarn test` step will give you the error messages I described in the OP
```
TS7010: 'log', which lacks return-type annotation, implicitly has an 'any' return type.
```

Moreover, it's safe to do this because `tsconfig.json` isn't included in the `files` section of package.json, so it isn't included in the npm package anyway.

JakeChampion (approved):

PR #264, opened 2022-09-15
fix: use random_get instead of arc4random as arc4random does not work correctly with wizer

JakeChampion:
wizer causes the seed in arc4random to be the same between executions which is not random


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4402,6 +4403,18 @@ bool process_network_io(JSContext *cx) {
   return true;
 }
 
+bool math_random(JSContext *cx, unsigned argc, Value *vp) {
+  int32_t storage;
+  int32_t *buf = &storage;
+  random_get((int32_t)buf, 8);
+  uint32_t value = std::abs(storage);
+  double newvalue = static_cast<float>(value) / std::powf(2.0, 32.0);
```
elliottt, line 4411: I'd probably do this all as a `double` if the result will be used as a double.
```suggestion
  double newvalue = static_cast<double>(value) / std::pow(2.0, 32.0);
```

c-dependencies/js-compute-runtime/builtins/crypto.cpp
```diff
@@ -44,7 +45,7 @@ bool Crypto::get_random_values(JSContext *cx, unsigned argc, JS::Value *vp) {
   JS::AutoCheckCannotGC noGC(cx);
   bool is_shared;
   void *buffer = JS_GetArrayBufferViewData(typed_array, &is_shared, noGC);
-  arc4random_buf(buffer, byte_length);
+  random_get((int32_t)buffer, byte_length);
```
elliottt, line 48: Isn't this going to be pretty slow? Should we pull over a PRNG and seed it from `random_get` when initializing?

tests/wpt-harness/expectations/fetch/api/basic/request-upload.any.js.json
```diff
@@ -48,7 +48,7 @@
     "status": "FAIL"
   },
   "Fetch with POST with text body on 421 response should be retried once on new connection.": {
-    "status": "PASS"
+    "status": "FAIL"
```
elliottt, line 51: Do you know why this test is failing now?

elliottt (commented):
Looking good, just a few questions :)

tests/wpt-harness/expectations/fetch/api/basic/request-upload.any.js.json
JakeChampion: The more surprising result was that it ever reported as passing, our fetch implementation does not include retry logic - which is what this test is attempting to exercise

elliottt: Quite mysterious!

PR #268, opened 2022-09-20
use random_get instead of arc4random

JakeChampion:
arc4random does not work correctly when combined with Wizer as Wizer snapshots the state of the WebAssembly module after initialization any values computed during initialization will be constant across fetch requests, 
including the seed of the CSPRNG used within arc4random. As the seed of the CSPRNG is fixed 
during this initialization process, the sequence of numbers generated become deterministic across 
instantiations of that same WebAssembly module.


elliottt (approved):
:+1:

PR #269, opened 2022-09-21
feat: implement validation for backend cipher definitions

JakeChampion:
Previously a backend could have any string set as the list of ciphers to use, even if the string contained no valid ciphers. This is not great because it would mean when the backend is used in a `fetch` call, the call will reject due to not being able to find an appropriate cipher to use.

By implementing validation on the ciphers being configured, we are able to throw an error at the earliest point possible, which should make it simpler for the user to locate where the error originally stems from. We are also able to provide a more specific custom error message, which should further help the user identify how to resolve the issue.

The validation is done by parsing the given cipher-list string as an OpenSSL CipherList (https://www.openssl.org/docs/manmaster/man1/openssl-ciphers.html#CIPHER-LIST-FORMAT) and then comparing the user supplied ciphers with the list of ciphers that are supported by Fastly. The reason we compare with the list supported by Fastly is because there are lots of ciphers but Fastly support 11 specific ciphers (and by default, only enable 4 specific ciphers) (https://developer.fastly.com/learning/concepts/routing-traffic-to-fastly/#use-a-tls-configuration), if we did not compare with the ones Fastly support then when a user tests their Compute@Edge application using `fastly compute serve` - it could possibly work but when deployed to Fastly's Compute@Edge platform, it would not work. By ensuring the user supplied cipher-list contains at-least one cipher supported by Fastly helps to ensure their application will work when deployed to the Compute@Edge platform.


integration-tests/js-compute/fixtures/dynamic-backend/bin/index.js
```diff
@@ -79,6 +81,31 @@ routes.set('/', () => {
...
+      backend
+    }));
+    if (error) { return error }
+    backend = createValidFastlyBackend();
+    error = await assertResolves(() => fetch('https://www.fastly.com', {
+      backend
+    }));
+    if (error) { return error }
+    return pass()
+  });
```
JakeChampion, line 108:
```suggestion
```

integration-tests/js-compute/fixtures/dynamic-backend/bin/index.js
```diff
@@ -1,4 +1,5 @@
 /* global fastly */
+let isRunningOnComputeAtEdge = false;
```
JakeChampion, line 2:
```suggestion
```

integration-tests/js-compute/fixtures/dynamic-backend/bin/index.js
```diff
@@ -11,6 +11,7 @@ async function app(event) {
     const path = (new URL(event.request.url)).pathname;
     console.log(`path: ${path}`)
     console.log(`FASTLY_SERVICE_VERSION: ${fastly.env.get('FASTLY_SERVICE_VERSION')}`)
+    isRunningOnComputeAtEdge = Boolean(fastly.env.get('FASTLY_SERVICE_VERSION'))
```
JakeChampion, line 14:
```suggestion
```

integration-tests/js-compute/fixtures/dynamic-backend/bin/index.js
```diff
@@ -924,7 +1247,6 @@ routes.set('/', () => {
         return pass()
       });
     }
```
JakeChampion, line 1249:
```suggestion
    }

```

c-dependencies/js-compute-runtime/builtins/backend.h
```diff
@@ -1,8 +1,273 @@
...
+  }
+};
+
+/**
+ * Class in charge with parsing openSSL expressions to define a list of ciphers.
+ */
+class OpenSSLCipherConfigurationParser {
+public:
+  // inline static std::map<std::string, std::vector<Cipher>> aliases;
+  inline static bool initialized = false;
```
elliottt, line 94: If you need to mutate this class variable, it shouldn't be marked `inline`. I would instead initialize it at the top level in backend.cpp as `bool OpenSSLCipherConfigurationParser::initialized = false;`. Another option would be to make all the static methods instance methods instead, and make a single top-level value of `OpenSSLCipherConfigurationParser` defined in an anonymous namespace in backend.cpp. That way you wouldn't need to worry about where to initialize class variables, as any state to be mutated would be an instance variable.
```suggestion
  static bool initialized;
```

c-dependencies/js-compute-runtime/builtins/backend.h
```diff
@@ -1,8 +1,273 @@
...
+  static void remove(std::map<std::string, std::vector<Cipher>> *aliases,
+                     std::vector<Cipher> *ciphers, std::string alias);
+
+  static void strengthSort(std::vector<Cipher> *ciphers);
+
+  /*
+   * See
+   * https://github.com/openssl/openssl/blob/7c96dbcdab959fef74c4caae63cdebaa354ab252/ssl/ssl_ciph.c#L1371
+   */
+  static std::vector<Cipher> defaultSort(std::vector<Cipher> *ciphers);
```
elliottt, line 236: I would make all these vector arguments into references, unless you have a need for `nullptr` arguments.
```suggestion
  static std::vector<Cipher> defaultSort(std::vector<Cipher> &ciphers);
```

elliottt (commented):
What do you think about moving the parser class entirely into `backend.cpp`? You could put the whole thing into the anonymous namespace, and then declare a single instance of the class that you could write wrapper functions to expose its behavior externally. I think this might be a little simpler, as you'll get to hide a lot of the messy details that aren't really important for the external interface, like all he string constants.

c-dependencies/js-compute-runtime/builtins/backend.h
```diff
@@ -1,8 +1,273 @@
...
+};
+
+/**
+ * Class in charge with parsing openSSL expressions to define a list of ciphers.
+ */
+class OpenSSLCipherConfigurationParser {
+public:
+  // inline static std::map<std::string, std::vector<Cipher>> aliases;
+  inline static bool initialized = false;
+  inline static std::string SSL_PROTO_TLSv1_2 = "TLSv1.2";
```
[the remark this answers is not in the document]
JakeChampion: I tried to use std::string_view but the data being referenced inside some of the string_views looked to be incorrect -- I moved to using `static constexpr auto` which expands into `static constexpr inline const char *`

c-dependencies/js-compute-runtime/builtins/backend.h
JakeChampion: I've refactored to using `std::once_flag` and `std::call_once` instead, which seems really nice

c-dependencies/js-compute-runtime/builtins/backend.h
JakeChampion: Definitely agree 👍

JakeChampion:
> What do you think about moving the parser class entirely into `backend.cpp`? You could put the whole thing into the anonymous namespace, and then declare a single instance of the class that you could write wrapper functions to expose its behavior externally. I think this might be a little simpler, as you'll get to hide a lot of the messy details that aren't really important for the external interface, like all he string constants.

@elliottt thanks for the excellent review -- I ended up taking a different approach where I've moved the openssl cipher parser into a header-only file and have a function in the Backend namespace named `isCipherSuiteSupportedByFastly` which is used. Do you think this approach is okay or should I refactor to the suggestion you made about moving the implementation into backend.cpp inside the anonymous namespace?

.vscode/settings.json
```diff
@@ -80,7 +80,10 @@
         "map": "cpp",
         "set": "cpp",
         "unordered_set": "cpp",
-        "numeric": "cpp"
+        "numeric": "cpp",
+        "__tree": "cpp",
+        "charconv": "cpp",
+        "regex": "cpp"
```
JakeChampion, line 86:
```suggestion
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -434,6 +435,32 @@ JSObject *Backend::create(JSContext *cx, JS::HandleObject request) {
...
+  //            c != '7' || c != '8' || c != '9';
+  //   });
+
+  //   if (it == origin.end()) {
+  //     // TODO: Make this an error.
+  //     return nullptr;
+  //   }
+  //   port = std::stoi(origin.substr(found_port, std::distance(origin.begin(), it)));
+  // }
+
```
JakeChampion, line 463:
```suggestion
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -3,10 +3,11 @@
 #include <charconv>
 #include <iostream>
 #include <optional>
+#include <set>
 #include <string>
 #include <string_view>
+#include <unordered_map>
```
JakeChampion, line 9:
```suggestion
#include <string>
#include <string_view>
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -6,7 +6,6 @@
 #include <string>
 #include <string_view>
 #include <vector>
```
JakeChampion, line 8:
```suggestion
#include <vector>

```

.vscode/settings.json
```diff
@@ -83,4 +83,4 @@
         "numeric": "cpp"
     },
     "git.ignoreLimitWarning": true
-}
\ No newline at end of file
+}
```
JakeChampion, line 86:
```suggestion
}

```

c-dependencies/js-compute-runtime/builtins/backend.h
```diff
@@ -1,7 +1,9 @@
 #ifndef JS_COMPUTE_RUNTIME_BACKEND_H
 #define JS_COMPUTE_RUNTIME_BACKEND_H
+#include <iostream>
```
JakeChampion, line 3:
```suggestion
```

JakeChampion:
@elliottt this is ready for another round of reviews 🙏

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
 #include "host_call.h"
 #include "js/Conversions.h"
 
+enum class Authentication {
```
elliottt, line 22: You can make all these `enum class` declarations subtype `uint8_t` to cut ~20 bytes out of the `Cipher` representation.
```suggestion
enum class Authentication : uint8_t {
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+
+class Cipher {
+  std::string openSSLAlias;
+  KeyExchange kx;
+  Authentication au;
+  Encryption enc;
+  MessageDigest mac;
+  Protocol protocol;
+  EncryptionLevel level;
+  int strength_bits;
```
elliottt, line 66: This can be a `uint16_t`, instead of `int`, which cuts some more space out of `Cipher`.

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+           strength_bits == obj.strength_bits;
+  }
+};
+
+/**
+ * Class in charge with parsing openSSL expressions to define a list of ciphers.
+ */
+class OpenSSLCipherConfigurationParser {
+private:
+  std::unordered_map<std::string, std::vector<Cipher>> aliases;
```
elliottt, line 103: This type gets used all over the place, and would benefit from having an alias introduced.

```suggestion
  using AliasMap = std::unordered_map<std::string, std::vector<Cipher>>;
  AliasMap aliases;
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+  // This unordered_map should stay aligned with the canonical list located at:
+  // https://developer.fastly.com/learning/concepts/routing-traffic-to-fastly/#use-a-tls-configuration
+  // The mapping is from OpenSSL cipher names as strings to a the cipher represented as a Cipher
+  // object
+  inline static std::unordered_map<std::string, Cipher> CIPHER{
```
elliottt, line 108: This can be `static constexpr` if the `Cipher` constructor is `constexpr`, but that requires refactoring `Cipher` to not use a `std::string`.
```suggestion
  static constexpr std::unordered_map<std::string, Cipher> CIPHER{
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+      }
+    }
+    return result;
+  }
+
+  std::vector<Cipher> all;
+
+public:
+  OpenSSLCipherConfigurationParser() {
+    all.reserve(CIPHER.size());
```
elliottt, line 469: I tend to like using `this->` on instance variables.
```suggestion
    this->all.reserve(CIPHER.size());
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+    auto toAdd = aliases->at(alias);
+    ciphers.insert(ciphers.end(), toAdd.begin(), toAdd.end());
+  }
+
+  void remove(std::unordered_map<std::string, std::vector<Cipher>> *aliases,
+              std::vector<Cipher> &ciphers, std::string alias) {
+    auto toRemove = aliases->at(alias);
+    ciphers.erase(std::remove_if(ciphers.begin(), ciphers.end(),
+                                 [&](auto x) {
+                                   return find(toRemove.begin(), toRemove.end(), x) !=
```
elliottt, line 297:
```suggestion
                                   return std::find(toRemove.begin(), toRemove.end(), x) !=
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+         std::optional<std::set<Encryption>> enc, std::optional<std::set<EncryptionLevel>> level,
+         std::optional<std::set<MessageDigest>> mac) {
+    std::vector<Cipher> result;
+    for (auto cipher : ciphers) {
+      if (protocols.has_value() && protocols->find(cipher.getProtocol()) != protocols->end()) {
+        result.push_back(cipher);
+      }
+      if (kx.has_value() && kx->find(cipher.getKx()) != kx->end()) {
+        result.push_back(cipher);
+      }
```
elliottt, line 414: Won't this potentially duplicate entries in the result vector? What do you think about ending each conditional with `continue`?

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+                  mac);
+  }
+
+  std::vector<Cipher>
+  filter(std::vector<Cipher> &ciphers, std::optional<std::set<Protocol>> protocols,
+         std::optional<std::set<KeyExchange>> kx, std::optional<std::set<Authentication>> au,
+         std::optional<std::set<Encryption>> enc, std::optional<std::set<EncryptionLevel>> level,
+         std::optional<std::set<MessageDigest>> mac) {
+    std::vector<Cipher> result;
+    for (auto cipher : ciphers) {
```
elliottt, line 408: Referencing the cipher will avoid eagerly copying it.
```suggestion
    for (auto &cipher : ciphers) {
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+  }
+
+  std::vector<Cipher> parse(std::string_view expression) {
+    /**
+     * All ciphers by their openssl alias name.
+     */
+    auto elements = splitCipherSuiteString(std::string(expression));
+    std::vector<Cipher> ciphers;
+    std::vector<Cipher> removedCiphers;
+    for (auto element : elements) {
```
elliottt, line 577: This string is getting copied for every iteration currently.
```suggestion
    for (auto &element : elements) {
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+        if (aliases.find(alias) != aliases.end()) {
+          moveToEnd(&aliases, ciphers, alias);
+        }
+      } else if ("@STRENGTH" == element) {
+        strengthSort(ciphers);
+        break;
+      } else if (aliases.find(element) != aliases.end()) {
+        add(&aliases, ciphers, element);
+      } else if (element.find(AND) != std::string::npos) {
+        auto intersections = split(element, std::string(1, AND).append("\\"));
```
elliottt, line 601: This will allocate the string "+\\" for each iteration that makes it to this branch. As the downstream consumer takes a `std::string_view`, you can instead use a literal to avoid the allocation.
```suggestion
        auto intersections = split(element, "+\\");
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
elliottt: Do you have a sense for how big this map gets? It might be just as efficient to represent it as a vector of pairs, instead of using an unordered map.

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -17,6 +19,617 @@
...
+            }
+          }
+          // Add all of `result` onto `ciphers`
+          ciphers.insert(ciphers.end(), result.begin(), result.end());
+        }
+      }
+    }
+    // Remove all ciphers from `ciphers` which are contained in `removedCiphers`
+    ciphers.erase(std::remove_if(ciphers.begin(), ciphers.end(),
+                                 [removedCiphers](Cipher c) {
```
elliottt, line 624: I'd reference this vector, instead of copying it into the closure.
```suggestion
                                 [&removedCiphers](Cipher c) {
```

elliottt (commented):
Left some comments, this is looking good!

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: It currently is a fixed size of 49 entries, which will change whenever a cipher is added or removed from Fastly's supported cipher list

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -3,20 +3,628 @@
...
+  }
+
+  std::vector<Cipher> filterByMessageDigest(std::vector<Cipher> &ciphers,
+                                            std::set<MessageDigest> mac) {
+    return filter(ciphers, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt,
+                  mac);
+  }
+
+  std::vector<Cipher>
+  filter(std::vector<Cipher> &ciphers, std::optional<std::set<Protocol>> protocols,
```
elliottt, line 403:
```suggestion
  filter(const std::vector<Cipher> &ciphers, std::optional<std::set<Protocol>> protocols,
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,602 @@
...
+      token = s.substr(pos_start, pos_end - pos_start);
+      pos_start = pos_end + delim_len;
+      res.push_back(token);
+    }
+
+    res.push_back(s.substr(pos_start));
+    return res;
+  }
+
+  std::vector<std::string> splitCipherSuiteString(std::string string) {
```
elliottt, line 378: Looking at how the result is used lower down, I think the input here could be a `std::string_view`, returning a vector of string_views.
```suggestion
  std::vector<std::string_view> splitCipherSuiteString(std::string_view string) {
```

elliottt (commented):
Found a few more opportunities to avoid copies :)

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,605 @@
...
+      pos_start = pos_end + delim_len;
+      res.push_back(token);
+    }
+
+    res.push_back(s.substr(pos_start));
+    return res;
+  }
+
+  std::vector<std::string_view> splitCipherSuiteString(std::string_view string) {
+    std::vector<std::string> result;
```
elliottt, line 378:
```suggestion
    std::vector<std::string_view> result;
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,605 @@
...
+  }
+
+  std::function<bool(const Cipher &)> byMessageDigest(MessageDigest val) {
+    return [val](auto &c) { return c.mac == val; };
+  }
+
+  std::vector<std::string_view> split(std::string_view s, std::string_view delimiter) {
+    size_t pos_start = 0, pos_end, delim_len = delimiter.length();
+    std::string token;
+    std::vector<std::string> res;
```
elliottt, line 365:
```suggestion
    std::vector<std::string_view> res;
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,605 @@
...
+    // The content of the default list is determined at compile time and normally corresponds to
+    // ALL:!COMPLEMENTOFDEFAULT:!eNULL.
+    aliases.insert({DEFAULT, parse("ALL:!COMPLEMENTOFDEFAULT:!eNULL")});
+  }
+
+  std::vector<Cipher> parse(std::string_view expression) {
+    /**
+     * All ciphers by their openssl alias name.
+     */
+    auto elements = splitCipherSuiteString(std::string(expression));
```
elliottt, line 544: This is allocating a temporary `std::string` as a copy of `expression` that will be freed after the call finishes. The resulting string views in `elements` will be references to the freed memory.
```suggestion
    auto elements = splitCipherSuiteString(expression);
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,605 @@
...
+          result.push_back(line.substr(prev, pos - prev));
+        }
+        prev = pos + 1;
+      }
+      if (prev < line.length()) {
+        result.push_back(line.substr(prev, std::string::npos));
+      }
+    }
+    return result;
+  }
```
elliottt, line 394: How about refactoring this loop like the following to avoid needing to use the `std::stringstream`?
https://godbolt.org/z/T9qhGM7YG

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,605 @@
...
+             ciphersToMoveToStart.end();
+    });
+  }
+
+  void add(AliasMap &aliases, std::vector<Cipher> &ciphers, std::string alias) {
+    auto toAdd = aliases.at(alias);
+    ciphers.insert(ciphers.end(), toAdd.begin(), toAdd.end());
+  }
+
+  void remove(AliasMap &aliases, std::vector<Cipher> &ciphers, std::string alias) {
```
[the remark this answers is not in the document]
JakeChampion: It can be a string_view 👍

c-dependencies/js-compute-runtime/builtins/backend.cpp
JakeChampion: I like it 😊

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,607 @@
...
+
+    return result;
+  }
+
+  std::vector<std::string_view> splitCipherSuiteString(std::string_view string) {
+    std::vector<std::string_view> result;
+    for (auto part : parts(string)) {
+      result.push_back(part);
+    }
+    return result;
```
elliottt, line 395: It might be worth renaming `parts` to `splitCipherSuiteString` if it's not needed elsewhere.
```suggestion
    return parts(string);
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,607 @@
...
+    auto complementOfDefault(ecdh);
+    complementOfDefault.insert(complementOfDefault.end(), triple_des.begin(), triple_des.end());
+    aliases.insert({COMPLEMENTOFDEFAULT, complementOfDefault});
+
+    // The content of the default list is determined at compile time and normally corresponds to
+    // ALL:!COMPLEMENTOFDEFAULT:!eNULL.
+    aliases.insert({DEFAULT, parse("ALL:!COMPLEMENTOFDEFAULT:!eNULL")});
+  }
+
+  std::vector<Cipher> parse(std::string_view expression) {
```
elliottt, line 542: Can this be marked `const`? That would ensure we're not accidentally mutating `aliases`.
```suggestion
  std::vector<Cipher> parse(std::string_view expression) const {
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,607 @@
...
+  // The ciphers included in ALL, but not enabled by default.
+  static constexpr auto COMPLEMENTOFDEFAULT = "COMPLEMENTOFDEFAULT";
+  static constexpr auto ALL = "ALL";
+
+  void moveToEnd(const AliasMap &aliases, std::vector<Cipher> &ciphers, std::string_view cipher) {
+    moveToEnd(ciphers, aliases.at(cipher));
+  }
+
+  void moveToEnd(std::vector<Cipher> &ciphers, const std::vector<Cipher> &ciphersToMoveToEnd) {
+    std::stable_partition(ciphers.begin(), ciphers.end(), [ciphersToMoveToEnd](auto cipher) {
```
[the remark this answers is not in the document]
elliottt: This one should be applied, without the `&` the `ciphersToMoveToEnd` vector will be copied into the closure, instead of referenced.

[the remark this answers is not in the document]
elliottt: I'd also change the argument to `auto &cipher` so that you don't copy it to use it with the `std::find` call on the next line.

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,606 @@
...
+
+  void add(const AliasMap &aliases, std::vector<Cipher> &ciphers, std::string_view alias) const {
+    auto toAdd = aliases.at(alias);
+    ciphers.insert(ciphers.end(), toAdd.begin(), toAdd.end());
+  }
+
+  void remove(const AliasMap &aliases, std::vector<Cipher> &ciphers, std::string_view alias) const {
+    auto &toRemove = aliases.at(alias);
+    ciphers.erase(std::remove_if(ciphers.begin(), ciphers.end(),
+                                 [&](auto x) {
```
elliottt, line 276: Referencing to avoid copying the cipher argument.
```suggestion
                                 [&](auto &x) {
```

c-dependencies/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1,22 +1,606 @@
...
+            }
+            // Add all of `result` onto `ciphers`
+            ciphers.insert(ciphers.end(), result.begin(), result.end());
+          }
+        }
+      }
+    }
+    // Remove all ciphers from `ciphers` which are contained in `removedCiphers`
+    ciphers.erase(std::remove_if(ciphers.begin(), ciphers.end(),
+                                 [&removedCiphers](Cipher c) {
```
elliottt, line 595: Avoid copying the `c` argument by turning it into a reference.
```suggestion
                                 [&removedCiphers](auto &c) {
```

elliottt (commented):
Found a few more places where a reference would avoid copying, and I think that https://github.com/fastly/js-compute-runtime/pull/269#discussion_r1046119919 might have been resolved accidentally. Once these last ones are addressed, I'm happy with merging :D

PR #270, opened 2022-09-21
feat: Add ConfigStore class

JakeChampion:
ConfigStore will eventually be the new name for Dictionary, but for now we will keep the Dictionary class around to allow customers to be able to slowly migrate to the new class.

### DIFF
--- a/c-dependencies/js-compute-runtime/builtins/config-store.cpp
+++ b/c-dependencies/js-compute-runtime/builtins/config-store.cpp
@@ -1,0 +1,66 @@
+#include "config-store.h"
+#include "host_call.h"
+
+namespace builtins {
+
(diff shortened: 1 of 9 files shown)


c-dependencies/js-compute-runtime/builtins/config-store.cpp
```diff
@@ -0,0 +1,66 @@
+#include "config-store.h"
```
JakeChampion, line 1: @elliottt - this and config-store.h are nearly identical copies of the already existing dictionary.cpp and dictionary.h files but with the builtin renamed to ConfigStore

elliottt (approved):
This looks great, thank you for the pointer about this being so similar to the Dictionary implementation! :bow:

PR #272, opened 2022-09-23
perf: avoid calling fastly_object_store_open twice by using the return status from the first call

JakeChampion:
the second call is not needed and could cause two handles to be created for the object store which is not required


elliottt (approved):
Great find!

PR #273, opened 2022-09-23
fix: implement validation for Dictionary names and keys

JakeChampion:
This commit updates our Dictionary implementation to ensure we throw nice error messages when given invalid dictionary names or keys, such as a name for a non-existant dictionary or a key that is too long.

I've also updated the xqd header to have an enum defined for the return codes of the host-calls as it makes it simpler to understand what each return code stands for.


JakeChampion:
This is currently blocked on this Viceroy PR -- https://github.com/fastly/Viceroy/pull/182

c-dependencies/js-compute-runtime/builtins/dictionary.cpp
```diff
@@ -11,13 +11,33 @@ DictionaryHandle Dictionary::dictionary_handle(JSObject *obj) {
...
+
   size_t name_len;
-  JS::UniqueChars name = encode(cx, args[0], &name_len);
+  // Convert into a String following https://tc39.es/ecma262/#sec-tostring
+  JS::UniqueChars name_chars = encode(cx, name_arg, &name_len);
+  if (!name_chars) {
+    return false;
+  }
+
+  // trim whitespace from key
```
elliottt, line 23: Was this a TODO? It looks like it's only checking `name_len` now.

c-dependencies/js-compute-runtime/builtins/dictionary.cpp
```diff
@@ -45,17 +65,65 @@ bool Dictionary::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  }
+
+  // If the converted string has a length of more than 255 then we throw an Error
+  // because Dictionary names have to be less than 255 characters.
+  if (name_len > 255) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_DICTIONARY_NAME_TOO_LONG);
+    return false;
+  }
+
+  std::string name(name_chars.get(), name_len);
```
elliottt, line 91: This will copy `name_chars`, would a `std::string_view` work?

elliottt (commented):
Nice, much better errors!

c-dependencies/js-compute-runtime/builtins/dictionary.cpp
JakeChampion, line 91:
```suggestion
  std::string_view name(name_chars.get(), name_len);
```

c-dependencies/js-compute-runtime/builtins/dictionary.cpp
JakeChampion: 🙈 this was just a not to myself and not a todo ha
```suggestion
```

elliottt (approved):
Looks good to me!

PR #274, opened 2022-09-23
Add FastlyStatus enum and update all hostcall definitions to return FastlyStatus

JakeChampion:
This should make it simpler to understand what each return code stands for within the codebase

### DIFF
--- a/c-dependencies/js-compute-runtime/builtins/config-store.cpp
+++ b/c-dependencies/js-compute-runtime/builtins/config-store.cpp
@@ -16,10 +16,11 @@ bool ConfigStore::get(JSContext *cx, unsigned argc, JS::Value *vp) {
 
   OwnedHostCallBuffer buffer;
   size_t nwritten = 0;
-  int status = xqd_config_store_get(ConfigStore::config_store_handle(self), name.get(), name_len,
-                                    buffer.get(), CONFIG_STORE_ENTRY_MAX_LEN, &nwritten);
-  // Status code 10 indicates the key wasn't found, so we return null.
-  if (status == 10) {
+  auto status = convert_to_fastly_status(
+      xqd_config_store_get(ConfigStore::config_store_handle(self), name.get(), name_len,
+                           buffer.get(), CONFIG_STORE_ENTRY_MAX_LEN, &nwritten));
+  // FastlyStatus::none indicates the key wasn't found, so we return null.
+  if (status == FastlyStatus::None) {
     args.rval().setNull();
     return true;
   }
--- a/c-dependencies/js-compute-runtime/geo_ip.cpp
+++ b/c-dependencies/js-compute-runtime/geo_ip.cpp
@@ -11,6 +11,8 @@ JSString *get_geo_info(JSContext *cx, JS::HandleString address_str) {
   if (!address)
     return nullptr;
 
+  // TODO: Remove all of this and rely on the host for validation as the hostcall only takes one
+  // user-supplied parameter
   int format = AF_INET;
   size_t octets_len = 4;
   const char *caddress = address.get();
@@ -27,6 +29,7 @@ JSString *get_geo_info(JSContext *cx, JS::HandleString address_str) {
     // While get_geo_info can be invoked through FetchEvent#client.geo, too,
     // that path can't result in an invalid address here, so we can be more
     // specific in the error message.
+    // TODO: Make a TypeError
     JS_ReportErrorLatin1(cx, "Invalid address passed to fastly.getGeolocationForIpAddress");
     return nullptr;
   }
--- a/c-dependencies/js-compute-runtime/host_call.cpp
+++ b/c-dependencies/js-compute-runtime/host_call.cpp
@@ -1,6 +1,45 @@
 
 #include "host_call.h"
 
+bool handle_fastly_result(JSContext *cx, int result, int line, const char *func) {
+  FastlyStatus status = convert_to_fastly_status(result);
+  return handle_fastly_result(cx, status, line, func);
+}
+
+FastlyStatus convert_to_fastly_status(int result) {
+  switch (result) {
+  case 0:
+    return FastlyStatus::Ok;
+  case 1:
+    return FastlyStatus::Error;
+  case 2:
+    return FastlyStatus::Inval;
+  case 3:
+    return FastlyStatus::BadF;
+  case 4:
+    return FastlyStatus::BufLen;
+  case 5:
+    return FastlyStatus::Unsupported;
+  case 6:
+    return FastlyStatus::BadAlign;
+  case 7:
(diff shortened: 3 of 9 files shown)


c-dependencies/js-compute-runtime/xqd.h
```diff
@@ -54,6 +54,49 @@ typedef struct {
 
 #define INVALID_HANDLE (UINT32_MAX - 1)
 
+typedef enum FastlyStatus {
+  // Success value.
+  // This indicates that a hostcall finished successfully.
+  ok = 0,
```
elliottt, line 60: I'd be inclined to capitalize the enum values.
```suggestion
  Ok = 0,
```

c-dependencies/js-compute-runtime/builtins/config-store.cpp
```diff
@@ -16,10 +16,11 @@ bool ConfigStore::get(JSContext *cx, unsigned argc, JS::Value *vp) {
 
   OwnedHostCallBuffer buffer;
   size_t nwritten = 0;
-  int status = xqd_config_store_get(ConfigStore::config_store_handle(self), name.get(), name_len,
-                                    buffer.get(), CONFIG_STORE_ENTRY_MAX_LEN, &nwritten);
-  // Status code 10 indicates the key wasn't found, so we return null.
-  if (status == 10) {
+  FastlyStatus status =
```
elliottt, line 19: Using `auto` here would be a bit more resilient to future refactoring as well.
```suggestion
  auto status =
```

elliottt (commented):
This seems pretty reasonable to me, but I wonder if we could get further by defining the enum outside of `xqd.h` as an `enum class`, and then defining a function to turn an `int` result into a value of that `enum class`. That way we could lean on the compiler to check exhaustiveness when we use `switch` on a return value from any of the host calls.

c-dependencies/js-compute-runtime/xqd.h
```diff
@@ -93,6 +93,49 @@ typedef struct DynamicBackendConfig {
 
 #define INVALID_HANDLE (UINT32_MAX - 1)
 
+typedef enum class FastlyStatus {
```
elliottt, line 96: I would put this definition in a different file, like our `host_call.h` header: `enum class` isn't valid C, and this declaration is inside the `extern "C" {` block started at the top of the header.

Defining it in a different header would have a nice side effect of allowing us to use c++ features a bit more freely :)

c-dependencies/js-compute-runtime/xqd.h
```diff
@@ -108,49 +151,52 @@ typedef enum BodyWriteEnd {
 
 // Module fastly_abi
 WASM_IMPORT("fastly_abi", "init")
-int xqd_init(uint64_t abi_version);
+FastlyStatus xqd_init(uint64_t abi_version);
```
elliottt, line 154: I think it's fine to leave all the host call definitions as returning `int` values, we can build a more typed api on top of them.

elliottt (commented):
Looking good! I think it would be worth trying to modify xqd.h as little as possible, and instead improving the boundary with the functions that header defines.

c-dependencies/js-compute-runtime/xqd.h
JakeChampion: Good point, I've moved it to host_call now 👍

c-dependencies/js-compute-runtime/xqd.h
JakeChampion: I've reverted them back to returning `int` values and have a new function which converts `int` to `FastlyStatus` and asserts the conversion was possible, I.E. will crash if trying to convert an integer which is not included as an enumeration in FastlyStatus

c-dependencies/js-compute-runtime/geo_ip.cpp
```diff
@@ -11,6 +11,8 @@ JSString *get_geo_info(JSContext *cx, JS::HandleString address_str) {
   if (!address)
     return nullptr;
 
+  // TODO: Remove all of this and rely on the host for validation as the hostcall only takes one
+  // user-supplied parameter
```
elliottt, line 15: Should we extract this to an issue?

c-dependencies/js-compute-runtime/geo_ip.cpp
```diff
@@ -27,6 +29,7 @@ JSString *get_geo_info(JSContext *cx, JS::HandleString address_str) {
     // While get_geo_info can be invoked through FetchEvent#client.geo, too,
     // that path can't result in an invalid address here, so we can be more
     // specific in the error message.
+    // TODO: Make a TypeError
```
elliottt, line 32: Should we extract this to an issue?

c-dependencies/js-compute-runtime/host_call.cpp
```diff
@@ -1,6 +1,45 @@
...
+    return FastlyStatus::HttpIncomplete;
+  case 10:
+    return FastlyStatus::None;
+  case 11:
+    return FastlyStatus::HttpHeadTooLarge;
+  case 12:
+    return FastlyStatus::HttpInvalidStatus;
+  default:
+    MOZ_ASSERT_UNREACHABLE("coding error");
+    return FastlyStatus::None;
```
elliottt, line 39: Should we make a `Unknown` enum variant for this case instead of re-using `None`?

c-dependencies/js-compute-runtime/host_call.h
```diff
@@ -1,56 +1,104 @@
...
-#include "js-compute-builtins.h"
+// TODO: remove these once the warnings are fixed
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winvalid-offsetof"
+#include "jsapi.h"
+#pragma clang diagnostic pop
 
 #include "xqd.h"
 
+typedef enum class FastlyStatus {
```
elliottt, line 12: You don't need to typedef this definition.
```suggestion
enum class FastlyStatus {
```

c-dependencies/js-compute-runtime/host_call.h
```diff
@@ -1,56 +1,104 @@
...
+  HttpIncomplete = 9,
+  // A `None` error.
+  // This status code is used to indicate when an optional value did not exist, as opposed to
+  // an empty value.
+  None = 10,
+  // Message head too large.
+  HttpHeadTooLarge = 11,
+  // Invalid HTTP status.
+  HttpInvalidStatus = 12,
+} FastlyStatus;
```
elliottt, line 53: Removing the other half of the `typedef` :)
```suggestion
};
```

elliottt (commented):
This looks great! The only blocking comments here are the unnecessary `typedef`, and reuse of the `None` enum variant.

elliottt (approved):
Looks great, thank you!

PR #279, opened 2022-09-28
Add note about Dynamic Backends being disabled for Fastly Services by default

JakeChampion:
The reason we disable it by default is because Dynamic Backends open up the potential for a data exfiltration attack to be possible, which is an attack vector not previously possible on a Fastly Service.


kailan (approved):
Thank you!

PR #281, opened 2022-09-30
feat: implement Request.prototype.setCacheKey

JakeChampion:
This method is used to set the cache key that will be used when attempting to satisfy this request from a cached response.

It is the equivalent of the VCL property `req.hash`.

The rust sdk version of [this method](https://docs.rs/fastly/latest/src/fastly/experimental.rs.html#183-195)
The header name is from xqd-headers
This [thread](https://fastly.slack.com/archives/CP8H97EDQ/p1664298940822429?thread_ts=1664296201.814199&cid=CP8H97EDQ) is where I learnt that this is the correct approach to take

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1137,7 +1137,7 @@ bool set_cache_key(JSContext *cx, HandleObject self, HandleValue cache_key_val) 
   if (!keyString) {
     return false;
   }
-  std::string key(keyString.get(), key_len);
+  std::string_view key(keyString.get(), key_len);
   std::string hex_str;
   picosha2::hash256_hex_string(key, hex_str);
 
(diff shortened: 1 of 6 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1128,6 +1129,34 @@ JSString *method(JSContext *cx, HandleObject obj) {
...
 
+bool set_cache_key(JSContext *cx, HandleObject self, HandleValue cache_key_val) {
+  MOZ_ASSERT(is_instance(self));
+  size_t key_len;
+  // Convert the key argument into a String following https://tc39.es/ecma262/#sec-tostring
+  JS::UniqueChars keyString = encode(cx, cache_key_val, &key_len);
+  if (!keyString) {
+    return false;
+  }
+  std::string key(keyString.get(), key_len);
```
elliottt, line 1140: Does `key` need to copy the bytes from `keyString`, or would it be possible to make a `std::string_view` and use that with `hash256_hex_string` instead?

elliottt (commented):
This looks good to me, just one question about whether or not we need to copy the key before hashing it.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I've changed to string_view 👍

elliottt (approved):
Looks great!

PR #287, opened 2022-10-11
download all artifacts at once by using actions/download-artifact@v3

JakeChampion:


elliottt (approved):
That's a pretty great cleanup :)

PR #292, opened 2022-10-18
Add more tests for fastly.getGeolocationForIpAddress

JakeChampion:
We previously only tested what happens when the function is given a single valid ipv4 address

This adds tests for:
- The functions descriptors
- The functions parameter length
- Calling the function as a constructor
- Calling the function with no parameters
- Confirming the function calls ToString on the parameter
- Calling the function with an empty string
- Calling with a valid ipv4 address
- Calling with a valid compressed ipv6 address
- Calling with a valid shortened ipv6 address
- Calling with a valid expanded ipv6 address
- ~Calling with a valid dual ipv4/v6 address~ -- This test found a bug in ipll which needs to be fixed first
- Calling as an unbound function


elliottt (approved):
Looks great!

PR #293, opened 2022-10-19
feature: implement setTimeout, setInterval, clearTimeout, and clearInterval

### LINKED ISSUE #63: setTimeout support?
`setTimeout` does not seem to work

JakeChampion:
Using the new fastly_async_io module we can implement JS timers 🥳 

This pull-request implements setTimeout, setInterval, clearTimeout, and clearInterval and adds some interface and basic tests for them. There are also tests in the web-platform-tests that exercise the functions fully, which is why I only needed to write some basic tests.

Currently Viceroy does not have a host implementation for fastly_async_io so for now our tests can only run in production.

Closes #63

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -4014,6 +4014,8 @@ private:
     timers.insert(iter, timer);
   }
 
+  // `repeat_first` must only be called if the `timers` list is not empty
+  // The caller of repeat_first needs to check the `timers` list is not empty
   void repeat_first() {
     Timer *timer = first();
     MOZ_ASSERT(timer);
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -3994,7 +3994,7 @@ public:
     if (std::empty(timers)) {
       return nullptr;
     } else {
-      return timers.front()
+      return timers.front();
     }
   }
 
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1,6 +1,8 @@
 #include <arpa/inet.h>
+#include <chrono>
 #include <cmath>
 #include <iostream>
+#include <list>
 #include <regex> // std::regex
 #include <stdio.h>
 #include <stdlib.h>
@@ -57,6 +59,10 @@
 #include "builtins/worker-location.h"
 
 using namespace std::literals;
+
+using std::chrono::ceil;
+using std::chrono::milliseconds;
+using std::chrono::system_clock;
 
 using JS::CallArgs;
 using JS::CallArgsFromVp;
@@ -128,8 +134,7 @@ static_assert(URI_MAX_LEN < HOSTCALL_BUFFER_LEN);
 using jsurl::SpecSlice, jsurl::SpecString, jsurl::JSUrl, jsurl::JSUrlSearchParams,
     jsurl::JSSearchParam;
 
-static JS::PersistentRootedObjectVector *pending_requests;
-static JS::PersistentRootedObjectVector *pending_body_reads;
+static JS::PersistentRootedObjectVector *pending_async_tasks;
 
 // TODO(performance): introduce a version that writes into an existing buffer, and use that
 // with the hostcall buffer where possible.
@@ -1087,7 +1092,7 @@ bool body_source_pull_algorithm(JSContext *cx, CallArgs args, HandleObject sourc
   // (This deadlock happens in automated tests, but admittedly might not happen
   // in real usage.)
 
-  if (!pending_body_reads->append(source))
+  if (!pending_async_tasks->append(source))
     return false;
 
   args.rval().setUndefined();
@@ -1131,7 +1136,7 @@ bool body_reader_then_handler(JSContext *cx, HandleObject body_owner, HandleValu
     }
 
     if (Request::is_instance(body_owner)) {
-      if (!pending_requests->append(body_owner)) {
+      if (!pending_async_tasks->append(body_owner)) {
         return false;
       }
     }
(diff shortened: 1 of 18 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+        break;
+      }
+    }
+
+    auto inserted = (*timers.insert(iter, timer));
+    return inserted->id;
+  }
+
+  void repeat_first() {
+    Timer *timer = first();
```
elliottt, line 3740: What happens if the `timers` list is empty? If this function won't be called in that case, it would be worth adding a comment here explaining why a null check isn't necessary.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+    JS::TraceEdge(trc, &callback, "Timer callback");
+    for (auto &arg : arguments) {
+      JS::TraceEdge(trc, &arg, "Timer callback arguments");
+    }
+  }
+};
+
+class ScheduledTimers {
+public:
+  Timer *first() { return *timers.begin(); }
```
elliottt, line 3720: This will read arbitrary memory if the list is empty. Is it only called in cases where you know the list is not empty?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+    RootedObject fun(cx, timer->callback);
+    JS::RootedVector<JS::Value> argv(cx);
+    if (!argv.initCapacity(timer->arguments.size())) {
+      JS_ReportOutOfMemory(cx);
+      return false;
+    }
+
+    for (size_t i = 0; i < timer->arguments.size(); i++) {
+      argv.infallibleAppend(timer->arguments[i]);
+    }
```
elliottt, line 3778:
```suggestion
    for (auto &arg : timer->arguments) {
      argv.infallibleAppend(arg);
    }
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+
+    // The timer might be deleted by content while running, so we mustn't access it again here.
+    timer = nullptr;
```
elliottt, line 3781: Another option would be to put the code that access `timer` into a lexical block, so that it's impossible to name it after the point that it's consumed.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+
+    RootedValue rval(cx);
+    if (!JS::Call(cx, JS::NullHandleValue, fun, argv, &rval)) {
+      return false;
+    }
+
+    // Repeat / remove the first timer if it's still the one we just ran.
```
elliottt, line 3788: In what situations would the timer at the head of the queue no longer be the one we just ran? Would it be easier to pop the timer off of the front of the queue and take ownership of it for the duration of the processing? That way you wouldn't have to worry about whether or not something was pushed in the meantime.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4305,13 +4437,83 @@ bool structuredClone(JSContext *cx, unsigned argc, Value *vp) {
...
+  JS::RootedValueVector handler_args(cx);
+  if (args.length() > 2 && !handler_args.initCapacity(args.length() - 2)) {
+    JS_ReportOutOfMemory(cx);
+    return false;
+  }
+  for (size_t i = 2; i < args.length(); i++) {
+    handler_args.infallibleAppend(args[i]);
+  }
+
+  uint32_t id = timers->add_timer(handler, uint32_t(delay), handler_args, repeat);
```
elliottt, line 4476:
```suggestion
  uint32_t id = timers->add_timer(handler, delay, handler_args, repeat);
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4444,17 +4615,69 @@ bool process_next_body_read(JSContext *cx) {
...
 }
 
-bool process_network_io(JSContext *cx) {
-  if (!has_pending_requests())
-    return true;
+bool process_pending_async_tasks(JSContext *cx) {
+  MOZ_ASSERT(has_pending_async_tasks());
+
+  uint32_t timeout = 0;
+  if (!timers->empty()) {
```
elliottt, line 4622: Is it possible for a timer that repeatedly sets another timer with delay `0` to starve the loop that processes `pending_async_tasks`?

elliottt (commented):
This is looking great! I haven't had to use `chrono` much, so this was fun to read through :D

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I don't believe it would starve the event loop, I ran the below in Firefox's console and could run other functions after it:
```js
function starve() {
console.log(2)
setTimeout(starve, 0)
}
starve()
console.log(1)
```

I'll see if I can come up with a test C@E application for this

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
 }
 } // namespace FetchEvent
 
+using std::chrono::system_clock;
```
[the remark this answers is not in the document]
JakeChampion: It was a duplicate of an earlier one, I've removed it 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+        JS::HandleValueVector args, bool repeat)
+      : id(id), callback(callback), delay(delay), deadline(deadline), repeat(repeat) {
+    deadline = system_clock::now() + system_clock::duration(delay * 1000);
+    arguments.reserve(args.length());
+    for (auto arg : args) {
+      arguments.push_back(JS::Heap(arg));
+    }
+  }
+
+  friend bool operator<(Timer const &left, Timer const &right) {
```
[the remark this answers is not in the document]
JakeChampion: Once I moved `Timer` and `ScheduledTimers` into a namespace I then found that this function was not used anywhere and so I've removed it - I did find this surprising though

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+  uint32_t delay;
+  system_clock::time_point deadline;
+  bool repeat;
+
+  Timer(uint32_t id, HandleObject callback, uint32_t delay, system_clock::time_point deadline,
+        JS::HandleValueVector args, bool repeat)
+      : id(id), callback(callback), delay(delay), deadline(deadline), repeat(repeat) {
+    deadline = system_clock::now() + system_clock::duration(delay * 1000);
+    arguments.reserve(args.length());
+    for (auto arg : args) {
```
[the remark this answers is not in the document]
JakeChampion: Nice!

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
...
+  JS::Heap<JSObject *> callback;
+  TimerArgumentsVector arguments;
+  uint32_t delay;
+  system_clock::time_point deadline;
+  bool repeat;
+
+  Timer(uint32_t id, HandleObject callback, uint32_t delay, system_clock::time_point deadline,
+        JS::HandleValueVector args, bool repeat)
+      : id(id), callback(callback), delay(delay), deadline(deadline), repeat(repeat) {
+    deadline = system_clock::now() + system_clock::duration(delay * 1000);
```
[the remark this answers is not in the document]
JakeChampion: good spot, I've removed the use of setting `deadline` via a parameter -- it was leftover from a refactor

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: Good idea! I've moved the code into a lexical block now 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,133 @@ bool response_started(JSObject *self) {
 }
 } // namespace FetchEvent
 
+using std::chrono::system_clock;
+using TimerArgumentsVector = std::vector<JS::Heap<JS::Value>>;
+
+class Timer {
```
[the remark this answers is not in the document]
JakeChampion: They're now wrapped in a namespace 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: It is only called when we know there is an item in the list, however I have also added an assert to this function and also changed it to use `.front()` instead:
```cpp
Timer *first() {
    MOZ_RELEASE_ASSERT(timers.size() > 0);
    return timers.front();
  }
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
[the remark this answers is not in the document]
JakeChampion: Update: the one location where we did a comparison on Timers was updated to instead compare on timer.deadline, which is why this operator was not being used anywhere

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I believe this can happen if the timer being ran adds a new timer which has a deadline that is earlier than the timer being ran. I'll see if that is actually possible to occur within a JS application

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: This explains why setTimeout with a duration of 0 doesn't starve the loop -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#zero_delays

JakeChampion: You can starve the loop with Promises though :D
```js
function starveWithPromises() {
console.log(3)
Promise.resolve().then(starveWithPromises)
}
starveWithPromises()
console.log(9) // this will never run
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I've changed this once more now, it now checks to see if the list is empty and if so, it returns nullptr, otherwise it returns the front of the list

JakeChampion:
@elliottt this is ready for another rounds of reviews -- I also used the version of Viceroy which has implementations for the `fastly_async_io` module so that I could update our web-platform-test results -- it's so good to see them all passing ☺️

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,141 @@ bool response_started(JSObject *self) {
...
+  }
+};
+
+class ScheduledTimers {
+public:
+  Timer *first() {
+    if (std::empty(timers)) {
+      return nullptr;
+    } else {
+      return timers.front()
```
elliottt, line 3719:
```suggestion
      return timers.front();
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
elliottt: A comment would be good here, thanks!

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3676,6 +3681,141 @@ bool response_started(JSObject *self) {
...
+      }
+    }
+
+    timers.insert(iter, timer);
+  }
+
+  void repeat_first() {
+    Timer *timer = first();
+    MOZ_ASSERT(timer);
+    timer->deadline = system_clock::now() + system_clock::duration(timer->delay * 1000);
```
elliottt, line 3742: Does this work? If so, it might be nice to skip having to multiply by 1000 :)
```suggestion
    timer->deadline = system_clock::now() + std::milliseconds(timer->delay);
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
elliottt: Ah I see: if this timer is set to repeat, then it moves the timer back in the queue rather than re-allocate the structure.

elliottt (approved):
This looks good to me!

JakeChampion:
This is currently blocked on a new version of Viceroy which includes the `fastly_async_io` module -- https://github.com/fastly/Viceroy/pull/188

Once that pull-request is merged and released, we can update our version of Viceroy and should then see all the tests pass on this PR 👍

PR #297, opened 2022-10-26
feat: implement support in Response.json/text/arrayBuffer methods for guest-backed streams

JakeChampion:
Currently we do not support guest-backed streams within the Response methods `arrayBuffer`/`json`/`text`, we throw an error, which means applications like the below do not work as expected:

```js
function iteratableToStream(iterable) {
    return new ReadableStream({
        async pull(controller) {
            for await (const value of iterable) {
                controller.enqueue(value);
            }
            controller.close();
        }
    });
}

async function app(event) {
    try {
        let res = new Response(iteratableToStream([new Uint8Array([65, 66])]))
        let text = await res.text()
        return new Response(text)
    } catch (error) {
        return new Response(error?.message || '', { status: 500 })
    }
}
addEventListener("fetch", event => event.respondWith(app(event)));
```

This application should respond with a HTTP 200 status and a body of `AB`

This pull-request implements the ability for the Response methods `arrayBuffer`/`json`/`text` to consume a guest-backed stream, making the above application work as expected 🎉


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,244 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+  MOZ_ASSERT(contentsIsArray);
+#endif
+
+  auto reader_val = js::GetFunctionNativeReserved(catch_handler, 1);
+  MOZ_ASSERT(reader_val.isObject());
+  RootedObject reader(cx, &reader_val.toObject());
+
+  // TODO: Confirm this -- I think it is not true...
+  // We're guaranteed to work with a native ReadableStreamDefaultReader here,
+  // which in turn is guaranteed to vend {done: bool, value: any} objects to
```
elliottt, line 734: Is `vend` a typo?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,244 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+    MOZ_ASSERT(foundBodyParser);
+#endif
+    // Now we can call parse_body on the result
+    RootedValue body_parser(cx);
+    if (!JS_GetElement(cx, catch_handler, 2, &body_parser)) {
+      return false;
+    }
+    auto parse_body = (ParseBodyCB *)body_parser.toPrivate();
+    size_t bytes_read = allBytes.size();
+    UniqueChars buf(reinterpret_cast<char *>(allBytes.data()));
```
elliottt, line 799: I'm not sure if this is safe: `UniqueChars` is a unique pointer holding a pointer to `char []` data, and will free the data when its destructor runs. As the vector still holds the bytes, this will likely be a double-free.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,244 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+
+  {
+    uint32_t contentsLength;
+    if (!JS::GetArrayLength(cx, contents, &contentsLength)) {
+      return false;
+    }
+    std::string index = std::to_string(contentsLength);
+    if (!JS_SetProperty(cx, contents, index.c_str(), val)) {
+      return false;
+    }
```
elliottt, line 826: Will `JS_SetProperty` parse the number out of the string when setting the property?

elliottt (commented):
It's awesome to see so many more passing tests!

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I'll change to `return` instead of `vend` as it seems less confusing 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: `JS_SetProperty` won't parse the number out of the string - we can actually avoid the conversion to a string here by using `JS_SetElement` -- I've now moved to using that instead 👍

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: I believe this is safe as we transfer ownership from the `UniqueChars buf` via the `std::move(buf)` call - is this not the case?

https://searchfox.org/mozilla-central/source/mfbt/UniquePtr.h#63-64

elliottt: The problem is that `allBytes` still owns the memory when `buf` is constructed. When `buf` is destroyed in `parse_body` it will call `free` on the bytes associated with `allBytes.data()`, and when `allBytes` is destroyed when this function returns it will also try to free the bytes.

JakeChampion: After reading through the `parse_body` implementation once more I can see the double free 👍 
I've refactored this to no longer use std::vector and instead use a char* which removes the double free issue

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,287 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+        JS::AutoCheckCannotGC nogc;
+        MOZ_ASSERT(val.isObject());
+        JSObject *array = &val.toObject();
+        MOZ_ASSERT(JS_IsUint8Array(array));
+        bool is_shared;
+        static_assert(CHAR_BIT == 8, "Strange char");
+        char *bytes = reinterpret_cast<char *>(JS_GetUint8ArrayData(array, &is_shared, nogc));
+        size_t length = JS_GetTypedArrayByteLength(array);
+        offset += length;
+        if (length + offset > buf_size) {
```
elliottt, line 795: Should offset be updated before testing against `buf_size`? This will test that there's enough space for `2*length`.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,287 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+          while (length + offset > buf_size) {
+            size_t new_size = buf_size + HANDLE_READ_CHUNK_SIZE;
+            new_buf = static_cast<char *>(JS_realloc(cx, buf, buf_size, new_size));
```
elliottt, line 798: Why not find the next boundary for the allocation in blocks of `HANDLE_READ_CHUNK_SIZE` and do the allocation once? I think `HANDLE_READ_SIZE * (length / HANDLE_READ_CHUNK_SIZE + 1)` should work?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,287 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+            buf_size += HANDLE_READ_CHUNK_SIZE;
+          }
+        }
+        // strcat(buf, *bytes);
+        memcpy(buf, bytes, length);
+        // allBytes.insert(allBytes.end(), bytes, bytes + length);
+      }
+    }
+    new_buf = static_cast<char *>(JS_realloc(cx, buf, buf_size, offset + 1));
+    if (!buf) {
```
elliottt, line 814: Did you mean `new_buf` here?

elliottt (commented):
This is looking good, thank you for fixing the double free! I'm beginning to wonder if we should rework the `parse_body` interface to always copy the buffer, as that would allow us to always pass in a const pointer to the source. Are we mostly hitting the ArrayBuffer case when calling `parse_body`?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,287 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+              JS_free(cx, buf);
+              JS_ReportOutOfMemory(cx);
+              return false;
+            }
+            buf = new_buf;
+            buf_size += HANDLE_READ_CHUNK_SIZE;
+          }
+        }
+        // strcat(buf, *bytes);
+        memcpy(buf, bytes, length);
```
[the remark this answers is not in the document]
JakeChampion: This was necessary as it was the only place in the function where we were putting the bytes into buf.

memcpy was very incorrect to use though as it would replace the buf contents instead of concatenating into buf.

I've updated to use strcpy and strcat instead and will add our own custom tests to ensure this functionality is correctly implemented as it seems web-platform-tests does not cover this

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: what is `HANDLE_READ_SIZE` defined as?
If it is meant to be `HANDLE_READ_CHUNK_SIZE` then I think that simplifies down to `HANDLE_READ_CHUNK_SIZE + length`

I've changed the implementation now so that a single realloc is called, once we know the new alloc size will be large enough -- [`311a645` (#297)](https://github.com/fastly/js-compute-runtime/pull/297/commits/311a645f142b484c2a7da4615cd60b719a5c2cb6)

elliottt: Sorry, that was supposed to be `HANDLE_READ_CHUNK_SIZE`. My point was that you could avoid the loop entirely by rounding up to the nearest multiple of `HANDLE_READ_CHUNK_SIZE` bytes.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,288 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+            buf_size = new_size;
+          }
+
+          static_assert(CHAR_BIT == 8, "Strange char");
+          auto bytes = reinterpret_cast<char *>(JS_GetUint8ArrayData(array, &is_shared, nogc));
+          if (index == 0) {
+            memcpy(buf, bytes, length);
+          } else {
+            memcpy(buf + offset, bytes, length);
+          }
```
elliottt, line 815: I think this conditional is redundant: when `index == 0`, `offset` is also `0`, so the two branches are the same.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,288 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+    // https://github.com/fastly/js-compute-runtime/issues/217
+    auto buf = static_cast<char *>(JS_malloc(cx, buf_size));
+    if (!buf) {
+      JS_ReportOutOfMemory(cx);
+      return false;
+    }
+    // For realloc below.
+    char *new_buf;
+    size_t offset = 0;
+    for (uint32_t index = 0; index < contentsLength; index++) {
```
elliottt, line 779: What do you think about iterating over the javascript array twice? One iteration could sum up the total allocation required, then you could allocate the correct sized buffer, then the second iteration would just memcpy the bytes quickly into the right sized buffer.

elliottt (commented):
I had a thought about how to avoid the reallocation in the reading loop :)

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: >I think that simplifies down to HANDLE_READ_CHUNK_SIZE + length

Silly me for simplifying this in a pure maths world when we are in a cpp world 😅

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: Excellent idea :D

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -694,11 +694,304 @@ bool parse_body(JSContext *cx, HandleObject self, UniqueChars buf, size_t len) {
...
+        MOZ_ASSERT(val.isObject());
+        JSObject *array = &val.toObject();
+        MOZ_ASSERT(JS_IsUint8Array(array));
+        size_t length = JS_GetTypedArrayByteLength(array);
+        if (length) {
+          // if buf is not big enough to fit the next uint8array's bytes then resize
+          if (offset + length > buf_size) {
+            size_t new_size =
+                buf_size + (HANDLE_READ_CHUNK_SIZE * ((length / HANDLE_READ_CHUNK_SIZE) + 1));
+            auto new_buf = static_cast<char *>(JS_realloc(cx, buf, buf_size, new_size));
```
elliottt, line 798: Instead of reallocing in the middle, accumulate `buf_size` and do a single `malloc` between the two loops. That way you only need the one single allocation, and can even avoid the final realloc to `buf_len+1` for the null terminator :)

elliottt (approved):
This is looking good now, thank you for iterating so much on the read/alloc logic!

PR #298, opened 2022-10-28
fix: respond with 500 Internal Server Error when an unhandled error has occured and no response has already been sent to the client

JakeChampion:
Currently the below program will return a 200 OK, which is likely not what the user is expecting:
```js
  throw new Error('uh oh');
});
```

This pull-request makes the above program return a 500 Internal Server Error instead


elliottt (approved):
This looks good to me, thanks for fixing this!

PR #302, opened 2022-11-10
refactor: convert the cli from rust to javascript (NodeJS)

JakeChampion:
This refactor means we no longer need to cross compile the CLI to each platform we want to support, as NodeJS supports all the platforms we want to target (and more!) 

This will also reduce our CI time and simplify our release process as we won't be committing multiple binaries into the npm package anymore

The README.md file in the root has been renamed to DEVELOPMENT.md as to me it looked more like information that useful when working on the js-compute-runtime codebase.

I've moved the npm package from the `sdk/js-compute` folder to the root of the project as it is now the sole CLI. The root README.md file now contains information on how to use the CLI.


src/compileApplicationToWasm.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+  } catch (error) {
+    console.error(
+      `Error: Failed to check whether the \`output\` (${output}) is a file path`,
+      error.message
+    );
+    process.exit(1);
+  }
+
+  let lits = findRegexLiterals(inputContents);
```
guybedford, line 116: Won't this throw for invalid JavaScript? If so, doesn't that already solve the TODO above? Would be worth checking how the tree sitter parse errors get displayed.

JakeChampion: It does not throw for invalid JavaScript unfortunately -- I'm currently adding tests for the CLI to confirm all the expected behavior and am including a test for invalid JavaScript 👍

guybedford (approved):
Did you mean to check in both a yarn.lock and a package-lock.json?

PR #312, opened 2022-11-25
New documentation website and fastly: modules

JakeChampion:
This pull-request updates the documentation website's design. The new design includes the ability to have the documentation be versioned, which will become very useful for customers who are using older versions of the SDK.

https://formally-flexible-anemone.edgecompute.app/dev/index.html

The homepage for the website now contains the contents of our README.md file rather than a list of types. I've also updated the README to contain a small example application, but we should revisit the README contents again at some point to improve it's use for people coming to Fastly C@E for the first time.

The types have also been updated in this pull-request to have the fastly specific features be defined in "fastly:" prefixed modules. This makes it simpler for users to see which APIs are fastly specific and also makes it simpler for js-compute-runtime to move to a Component Model world.

image

image


test-d/fastly:backend.test-d.ts
```diff
@@ -0,0 +1,26 @@
...
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', hostOverride: 'example.com', }))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', connectTimeout: 5000, }))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', firstByteTimeout: 5000, }))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', betweenBytesTimeout: 5000, }))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', useSSL: true,}))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', tlsMinVersion: 1.2,}))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', tlsMaxVersion: 1.2,}))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', certificateHostname: 'example.com',}))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', checkCertificate: false,}))
+    expectType<Backend>(new Backend({name: 'eu', target: 'www.example.com', caCertificate: '',}))
```
guybedford, line 21: Is the empty string a meaningful value for `caCertificate`, or is it the same as providing `undefined`?

types/fastly:backend.d.ts
```diff
@@ -0,0 +1,93 @@
...
+    betweenBytesTimeout?: number;
+    /**
+     * Whether or not to require TLS for connections to this backend.
+     */
+    useSSL?: boolean;
+    /**
+     * Minimum allowed TLS version on SSL connections to this backend.
+     * If the backend server is not able to negotiate a connection meeting this constraint, a 503 response will be presented instead.
+     */
+    tlsMinVersion?: number;
```
guybedford, line 42: Should we use an enum for `tlsMinVersion` instead of a number? Effectively `"1.1" | "1.2" | "1.3"` as a string?

guybedford (approved):
Great! Will `fastly:env` / `fastly:geolocation` / `fastly:logger` exact methods still be filled out here?

types/fastly:backend.d.ts
JakeChampion: > Should we use an enum for `tlsMinVersion` instead of a number? Effectively `"1.1" | "1.2" | "1.3"` as a string?

Good idea, let's do this in a separate PR to this docs one -- I'm happy to move to some form of enum, what do you think about Symbols instead of strings? Like this -- `import { Backend, TLS_1_2 } from "fastly:backend"`

test-d/fastly:backend.test-d.ts
JakeChampion: > Is the empty string a meaningful value for `caCertificate`, or is it the same as providing `undefined`?

Providing an empty string will throw a TypeError at runtime with the message `"Backend constructor: caCertificate can not be an empty string"`

PR #313, opened 2022-11-25
feature: support `import` and expose the fastly features via 'fastly:' namespaced modules 

JakeChampion:
This makes it so that a customer does not have to use a bundler in their own project, instead they can use `js-compute-runtime ./my/cool/app.js` and it will bundle and compile their application as expected.

This pull-request also adds 'fastly:' namespaced modules for features such as Dictionary, deprecating the global definitions. This makes it simpler for customers to see what features are Fastly specific and also helps align js-compute-runtime with the Component Model, which we are moving towards supporting


src/bundle.js
```diff
@@ -0,0 +1,24 @@
...
+          'fastly:dictionary': `export const Dictionary = globalThis.Dictionary`,
+          'fastly:object-store': `export const ObjectStore = globalThis.ObjectStore`,
+        })
+      ]
+    });
+    return await bundle.generate({
+      format: 'iife'
+    })
+  } catch (error) {
+    console.error(error);
```
JakeChampion, line 21: TODO: Need better error handling here

src/bundle.js
```diff
@@ -0,0 +1,28 @@
...
+    const bundle = await rollup({
+      input,
+      plugins: [
+        virtual({
+          'fastly:backend': `export const Backend = globalThis.Backend`,
+          'fastly:cache-override': `export const CacheOverride = globalThis.CacheOverride`,
+          'fastly:config-store': `export const ConfigStore = globalThis.ConfigStore`,
+          'fastly:dictionary': `export const Dictionary = globalThis.Dictionary`,
+          'fastly:env': `export const env = globalThis.fastly.env.get`,
+          'fastly:geolocation': `export const lookup = globalThis.fastly.getGeolocationForIpAddress`,
```
guybedford, line 15: Perhaps keep the name `getGeolocationForIpAddress`? Or something with `geo` in the name like `geoLookup`?

guybedford (approved):

PR #316, opened 2022-11-29
fix: Add types for setTimeout/etc

harmony7:
This PR adds types for setTimeout/clearTimeout/setInterval/clearInterval. If you're using TypeScript, then declarations for these are needed to use them.

```typescript
async function handleRequest(event: FetchEvent) {
  await new Promise<void>(resolve => {
    setTimeout(() => { resolve(); }, 1000);
  });
  return new Response("OK", { status: 200 });
}
```

```
ERROR in /Users/komuro/Developer/learning/c-at-e/ts/src/index.ts
./src/index.ts 8:4-14
[tsl] ERROR in /Users/komuro/Developer/learning/c-at-e/ts/src/index.ts(8,5)
      TS2304: Cannot find name 'setTimeout'.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1187,6 +1187,47 @@ declare function atob(data: string): string;
 declare function btoa(data: string): string;
 
 /**
+ * The setTimeout() method sets a timer which calls a function once the timer expires.
+ *
+ * @param callback A function to be called after the timer expires.
+ * @param delay The time, in milliseconds, that the timer should wait before calling the specified function. Defaults to 0 if not specified.
+ * @param args Additional arguments which are passed through to the callback function.
+ * @returns A numeric, non-zero value which identifies the timer created; this value can be passed to clearTimeout() to cancel the timeout.
+ *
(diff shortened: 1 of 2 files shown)


harmony7:
If `tsconfig.json` is not set up to exclude `node` from `types`, then the definitions for Node's global `setTimeout`/etc can get picked up (from `@types/node`) by TypeScript, and these functions may appear to work without the changes in this PR. However, I strongly recommend excluding `node` from `types` and using `js-compute`'s types file to declare everything we make available.

By the way, `@types/node` happens to always be in the project, because C@E JS apps always use `webpack`, which depends on `terser-webpack-plugin`, which depends on `jest-worker`, which depends on `@types/node`. I would consider this a bug in `jest-worker`, which I would argue should put `@types/node` in its `devDependencies` instead of `dependencies`. (And a ticket does seem to exist in Jest https://github.com/facebook/jest/issues/10907)

JakeChampion (approved):

PR #319, opened 2022-11-29
fix: ensure custom cache keys are uppercased

### LINKED ISSUE #318: Custom Cache Keys Must Use Uppercase HEX Characters
After debugging an invalid argument error with Fastly support we realized that custom cache cache keys must be 64 uppercase HEX characters and right now they are lowercase HEX characters.

https://github.com/fastly/js-compute-runtime/pull/281/files#diff-bca0f8f03afc68c49c0881002ded55236dc27932e34e1f60a7827a9bbe54c58fR92

JakeChampion:
Fixes #318

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.cpp
@@ -1423,6 +1423,8 @@ bool set_cache_key(JSContext *cx, HandleObject self, HandleValue cache_key_val) 
   std::string_view key(keyString.get(), key_len);
   std::string hex_str;
   picosha2::hash256_hex_string(key, hex_str);
+  std::transform(hex_str.begin(), hex_str.end(), hex_str.begin(),
+                 [](unsigned char c) { return std::toupper(c); });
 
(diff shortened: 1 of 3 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1423,6 +1423,8 @@ bool set_cache_key(JSContext *cx, HandleObject self, HandleValue cache_key_val)
   std::string_view key(keyString.get(), key_len);
   std::string hex_str;
   picosha2::hash256_hex_string(key, hex_str);
+  std::transform(hex_str.begin(), hex_str.end(), hex_str.begin(),
+                 [](unsigned char c) { return std::toupper(c); });
```
elliottt, line 1427: Nice. I didn't realize it was acceptable to use the input range beginning as the output range beginning, but the docs call that out explicitly: https://en.cppreference.com/w/cpp/algorithm/transform.

elliottt (approved):
Looks great, thanks!

PR #326, opened 2022-12-02
Compute runtime component integration

guybedford:
This includes the WIT for the compute runtime (in `c-dependencies/js-compute-runtime/xqd.wit`), handling the generation of all host bindings through wit-bindgen via the C generator into `c-dependencies/js-compute-runtime/xqd-world/xqd_world.h` and `xqd_world.c`. Bindings are generated as one-off and committed via `make regnerate-bindings` in `c-dependencies/js-compute-runtime`. All host calls are updated to use these new C bindings interfaces - which means using these WIT-generated handles, and C-style WIT generated types for the interfaces. Ideally these interfaces should be using C++ idioms via a C++ version of wit-bindgen, but that project does not yet exist, so at least having a small C-style component binding layer integrated into the codebase seems like the best approach for now.

The makefile is updated to build two targets now - `js-compute-runtime.wasm` as well as `js-compute-runtime-component.wasm`. This is done so that we don't create two builds of each file, but instead have the component model APIs being called in both builds to create a unified approach. In order to support the component-style host APIs on the normal `js-compute-runtime.wasm` build, a new adapter layer at `xqd-world/xqd_world_adapter.cc` is added which branches between using the bindgen-generated component interface or wrapping the `xqd.h` interfaces with a component interface depending on which build is being done.

This approach should in theory allow merging the component work and supporting two builds while using the `xqd.wit` and `xqd.h` side-by-side hopefully without too much extra overhead.

I haven't been able to properly get the tests running locally, so will be continuing to debug this PR still. The component being built is fully virtualized on the component model down to the WASI integration now, which I plan to work on next week.

The PR _is_ ready for initial review and feeback though which I would be very grateful for @JakeChampion @elliottt if you have a moment.


c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -0,0 +1,510 @@
+world xqd-world {
```
guybedford, line 1: The WIT world can be reviewed here.

c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cc
```diff
@@ -0,0 +1,926 @@
+#include "xqd_world_adapter.h"
+
+JSContext *context = nullptr;
+
+void set_js_context(JSContext *cx) { context = cx; }
+
+#ifndef COMPONENT
```
guybedford, line 7: The non-component build using the component-generated APIs is handled by the wrapper functions for `xqd.h` here.

guybedford:
All tests are finally passing here, ready for review.

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -43,9 +45,10 @@ WASM_STRIP = wasm-opt --strip-debug -o $1 $1
 endif
 endif
 
+OBJ_DIR := obj/
```
guybedford, line 48: This PR also refactors the Makefile to use an explicit object directory.

c-dependencies/js-compute-runtime/Makefile
```diff
@@ -23,9 +23,11 @@ SM_OBJ := $(SM_SRC)lib/*.o
 SM_OBJ += $(SM_SRC)lib/*.a
 FSM_SRC := $(ROOT_SRC)/js-compute-runtime/
 
+WASI_CC ?= /opt/wasi-sdk/bin/clang
 WASI_CXX ?= /opt/wasi-sdk/bin/clang++
+WIT_BINDGEN ?= ~/src/bytecodealliance/wit-bindgen/target/release/wit-bindgen
```
JakeChampion, line 28: Could we update the default path to be similar to where wasi-sdk is stored?
Hmm, another idea... if `WIT_BINDGEN` is not set, we could do this:
```
WIT_BINDGEN := $(shell which wit-bindgen)

ifndef WIT_BINDGEN
     $(error "No wit-bindgen in $PATH, consider doing cargo install --git https://github.com/bytecodealliance/wit-bindgen wit-bindgen-cli")
endif
```

That way we allow the user to store wit-bindgen anywhere they want to and we return a helpful error message which includes how to install wit-bindgen

c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -0,0 +1,510 @@
...
+    /*
+     * Fastly Backend
+     */
+    enum backend-health {
+      unknown,
+      healthy,
+      unhealthy,
+    }
+
+    backend-is-healthy: func(backend: string) -> result<backend-health, error>
```
JakeChampion, line 460: This doesn't have a public ABI definition as of today (It is not in the public wit stored in Viceroy) - which is the deciding factor on whether we include a feature in the public SDKs. Do you know if that deciding factor changes now that we use the component model?

c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -0,0 +1,510 @@
...
+    /*
+     * Fastly KV
+     */
+    type kv-store-handle = u32
+
+    kv-open: func(name: string) -> result<kv-store-handle, error>
+
+    kv-lookup: func(store: kv-store-handle, key: list<u8>) -> result<option<body-handle>, error>
+
+    kv-insert: func(store: kv-store-handle, key: list<u8>, body-handle: body-handle, max-age: u32) -> result<bool, error>
```
JakeChampion, line 419: Same comment here as for backend-is-healthy, these definitions are not in the Viceroy wit files, which makes me think we shouldn't have them in js-compute-runtime yet

c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -0,0 +1,510 @@
...
+      http-incomplete,
+      /// A `None` error.
+      /// This status code is used to indicate when an optional value did not exist, as opposed to
+      /// an empty value.
+      /// Note, this value should no longer be used, as we have explicit optional types now.
+      optional-none,
+      /// Message head too large.
+      http-head-too-large,
+      /// Invalid HTTP status.
+      http-invalid-status,
```
JakeChampion, line 45: This is missing the newer error case for exceeding limits (https://github.com/fastly/Viceroy/commit/8ee547b53cadb8366082bd7ca931be0a3a770411):
```suggestion
      http-invalid-status,
      // Limit exceeded
      // This is returned when an attempt to allocate a resource has exceeded the maximum number of 
      // resources permitted. For example, creating too many response handles.
      limit-exceeded,
```

c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cc
```diff
@@ -0,0 +1,966 @@
...
+  case 8:
+    return FASTLY_ERROR_HTTP_USER;
+  case 9:
+    return FASTLY_ERROR_HTTP_INCOMPLETE;
+  case 10:
+    return FASTLY_ERROR_OPTIONAL_NONE;
+  case 11:
+    return FASTLY_ERROR_HTTP_HEAD_TOO_LARGE;
+  case 12:
+    return FASTLY_ERROR_HTTP_INVALID_STATUS;
```
JakeChampion, line 53: This is missing the newer error case for exceeding limits (https://github.com/fastly/Viceroy/commit/8ee547b53cadb8366082bd7ca931be0a3a770411):
```suggestion
  case 12:
    return FASTLY_ERROR_HTTP_INVALID_STATUS;
  case 13:
    return FASTLY_ERROR_LIMIT_EXCEEDED;
```

c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
```diff
@@ -0,0 +1,392 @@
...
+#define FASTLY_ERROR_BAD_HANDLE 3
+#define FASTLY_ERROR_BUFFER_LEN 4
+#define FASTLY_ERROR_UNSUPPORTED 5
+#define FASTLY_ERROR_BAD_ALIGN 6
+#define FASTLY_ERROR_HTTP_INVALID 7
+#define FASTLY_ERROR_HTTP_USER 8
+#define FASTLY_ERROR_HTTP_INCOMPLETE 9
+#define FASTLY_ERROR_OPTIONAL_NONE 10
+#define FASTLY_ERROR_HTTP_HEAD_TOO_LARGE 11
+#define FASTLY_ERROR_HTTP_INVALID_STATUS 12
```
JakeChampion, line 31: This is missing the newer error case for exceeding limits (https://github.com/fastly/Viceroy/commit/8ee547b53cadb8366082bd7ca931be0a3a770411):
```suggestion
#define FASTLY_ERROR_HTTP_INVALID_STATUS 12
#define FASTLY_ERROR_LIMIT_EXCEEDED 13
```

c-dependencies/js-compute-runtime/builtins/fastly.cpp
```diff
@@ -16,7 +16,7 @@
 
 namespace builtins {
 
-bool Fastly::debug_logging_enabled = false;
+bool Fastly::debug_logging_enabled = true;
```
JakeChampion, line 19: Was this potentially left-over from a debugging session?

```suggestion
bool Fastly::debug_logging_enabled = false;
```

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -99,37 +99,29 @@ enum { ObjectStore, Count };
...
-  if (key_len > 1024) {
+  if (len > 1024) {
     JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_OBJECT_STORE_KEY_TOO_LONG);
-    return std::nullopt;
+    return false;
   }
 
-  std::string key(keyString.get(), key_len);
-  auto key_chars = key.c_str();
+  auto key_chars = key->get();
```
JakeChampion, line 124: nice work removing a copy :D

c-dependencies/js-compute-runtime/host_call.cpp
```diff
@@ -40,6 +45,40 @@ FastlyStatus convert_to_fastly_status(int result) {
...
+  case FASTLY_ERROR_HTTP_USER:
+    return FastlyStatus::HttpUser;
+  case FASTLY_ERROR_HTTP_INCOMPLETE:
+    return FastlyStatus::HttpIncomplete;
+  case FASTLY_ERROR_OPTIONAL_NONE:
+    return FastlyStatus::None;
+  case FASTLY_ERROR_HTTP_HEAD_TOO_LARGE:
+    return FastlyStatus::HttpHeadTooLarge;
+  case FASTLY_ERROR_HTTP_INVALID_STATUS:
+    return FastlyStatus::HttpInvalidStatus;
```
JakeChampion, line 75:
```suggestion
  case FASTLY_ERROR_HTTP_INVALID_STATUS:
    return FastlyStatus::HttpInvalidStatus;
  case FASTLY_ERROR_LIMIT_EXCEEDED:
    return FastlyStatus::LimitExceeded;
```

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -2965,49 +3041,35 @@ static bool retrieve_value_for_header_from_handle(JSContext *cx, HandleObject se
...
+  FastlyStatus result;
+  if (mode == Headers::Mode::ProxyToRequest) {
+    result = convert_to_fastly_status(xqd_fastly_http_req_header_values_get(handle, &str, &ret));
+  } else {
+    result = convert_to_fastly_status(xqd_fastly_http_resp_header_values_get(handle, &str, &ret));
+  }
 
-            if (!append_header_value_to_map(cx, self, name, value))
-              return false;
+  if (!ret.is_some)
```
JakeChampion, line 3058: Do we need to check `result` here?
```suggestion
  if (!HANDLE_RESULT(cx, result)){
    return false;
  }
  if (!ret.is_some)
```

c-dependencies/js-compute-runtime/xqd.wit
guybedford: Sure, added and also extended a new JS error message for this as well.

c-dependencies/js-compute-runtime/builtins/dictionary.cpp
```diff
@@ -3,51 +3,51 @@
...
     return false;
 
-  JS::RootedString text(cx, JS_NewStringCopyUTF8N(cx, JS::UTF8Chars(buffer.get(), nwritten)));
+  if (!ret.is_some) {
+    args.rval().setNull();
+    return true;
+  }
+
+  JS::RootedString text(cx, JS_NewStringCopyUTF8N(cx, JS::UTF8Chars(ret.val.ptr, ret.val.len)));
+  JS_free(cx, ret.val.ptr);
```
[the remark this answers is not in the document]
guybedford: As discussed earlier, we should be able to do this in due course!

c-dependencies/js-compute-runtime/builtins/fastly.cpp
guybedford: Thanks for catching that.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
guybedford: Yes, it seems the clang upgrade caught this bug!

JakeChampion (approved):
This is fantastic work 🎉

c-dependencies/js-compute-runtime/xqd.h
```diff
@@ -224,163 +172,128 @@ int xqd_req_downstream_client_ip_addr_get(char *octets, size_t *nwritten);
...
-                                 const char *backend, size_t backend_len,
-                                 PendingRequestHandle *pending_req_out);
+int xqd_req_send_async_streaming(fastly_request_handle_t req_handle,
+                                 fastly_body_handle_t body_handle, const char *backend,
+                                 size_t backend_len,
+                                 fastly_pending_request_handle_t *pending_req_out);
 
-WASM_IMPORT("fastly_http_req", "pending_req_poll")
-int xqd_req_pending_req_poll(PendingRequestHandle req_handle, uint32_t *is_done_out,
-                             ResponseHandle *resp_handle_out, BodyHandle *resp_body_handle_out);
```
JakeChampion, line 299: I'm not 100% on this but I think this has been removed by mistake, is that correct?

c-dependencies/js-compute-runtime/xqd-world/xqd_world.c
```diff
@@ -0,0 +1,2480 @@
...
+    result.val.err = (int32_t)(*((uint8_t *)(ptr + 4)));
+    break;
+  }
+  }
+  *ret = result.val.ok;
+  *err = result.val.err;
+  return result.is_err ? 1 : 0;
+}
+
+bool fastly_async_io_is_ready(fastly_async_handle_t handle, bool *ret, fastly_error_t *err) {
```
JakeChampion, line 2429: The wpt debug build failed on this function with this error:

> WebAssembly trapped: In func fastly_async_io::is_ready at write is_done: Pointer not aligned to 4: Region { start: 1048374, len: 4 }
> wasm backtrace:
>     0: 0x79b00b - <unknown>!<wasm function 15488>
>     1: 0x25055 - <unknown>!<wasm function 296>
>     2: 0x2dd97 - <unknown>!<wasm function 396>
>     3: 0x809bf1 - <unknown>!<wasm function 18035>
>     4: 0x2d98c - <unknown>!<wasm function 395>

c-dependencies/js-compute-runtime/xqd.h
guybedford: This is unused, so I removed the unused bindings, but I can appreciate we should leave them in!

guybedford: I've added back all removed xqd.h bindings except for the config-store ones which aliased the dictionary.

JakeChampion (approved):

c-dependencies/js-compute-runtime/xqd-world/xqd_world.c
guybedford: Thanks, think I found the issue.

JakeChampion (approved):

Issue #47, opened 2021-10-29
Serialisation issue with Headers and Set-Cookie


williamoverton:
When setting set-cookie headers with `header.append()` there can sometimes be an issue where the set-cookie header is serialised into one header which is comma seperated:
![image](https://user-images.githubusercontent.com/2838108/139453411-263c0418-21cc-42f7-b20d-cdf3f0d7cb56.png)

This is not accepted as multiple cookies by browsers, after some digging I have found that this only happens when using a Header object which is then passed as a HeaderInit object when creating a response.

This works (creates multiple headers):
```javascript
addEventListener("fetch", (event) => {
  let r = new Response("Hello");

  r.headers.append("Set-Cookie", "test=1");
  r.headers.append("Set-Cookie", "test2=2");
  r.headers.append("Set-Cookie", "test3=3");
  
  event.respondWith(r)
});
```

This does not (single comma separated header):
```javascript
addEventListener("fetch", (event) => {
  let h = new Headers();

  h.append("Set-Cookie", "test=1");
  h.append("Set-Cookie", "test2=2");
  h.append("Set-Cookie", "test3=3");
  
  event.respondWith(new Response("Hello", {
    headers: h
  }))
});
```

AndersDJohnson:
@JakeChampion @williamoverton Any updates on this one? It's a really critical feature before we can use Fastly Compute@Edge with JavaScript.

We need to be able to `fetch` to a backend, and get multiple `Set-Cookie` headers from its response.
And then also reconstruct a new `Response` with those multiple `Set-Cookie` headers plus some additional ones.

AndersDJohnson:
I can confirm @williamoverton's observation that `response.headers.append(...)` seems to work in that it adds additional headers of the same name without overwriting existing ones.

But beyond that, we're still blocked. For example, when trying to re-create a new response by cloning one but modifying its body, e.g.:

```js
const newResponse = new Response('new body', {
  headers: originalResponse.headers
});
```

The multiple headers of the same name are collapsed into a single one with comma-separated value.

AndersDJohnson:
For what's it's worth, it does seem to work to pull `Set-Cookie` header from the original response, split it with `set-cookie-parser`'s `splitCookiesString`, then clone it with `new Response('new body', originalResponse)` and call `newResponse.headers.delete('Set-Cookie')` then loop thru the split cookies and call `newResponse.headers.append(...)` for each, and then proceed with any custom logic to append more `Set-Cookie` headers at will.

PR #329, opened 2022-12-06
fix: ensure we do not collapse multiple set-cookies headers into one header when constructing a new Response with an existing Headers instance

JakeChampion:
Fixes: https://github.com/fastly/js-compute-runtime/issues/47

- When appending multiple headers of the same name onto a JavaScript Headers instance, the specification states they must be collapsed into a single value (https://fetch.spec.whatwg.org/#concept-headers-append)
- When constructing a Response instance with a ResponseInit parameter that contains a Headers instance, we call `maybe_consume_sequence_or_record<detail::append_header_value>` which iterates over the sequence and appends the headers onto the new Response instance, this is where the issue arises, as we append the collapsed set-cookie header and we should explode it into multiple headers

I've modified `detail::append_header_value` to be aware of headers named `set-cookie` and we now split the header into multiple values and make sure the host has stored each one as a separate header


elliottt (approved):
This looks good to me!

AndersDJohnson:
Awesome, thanks @JakeChampion! 💯

PR #331, opened 2022-12-07
chore: Fix the tests/wpt-harness/run-wpt.sh script

elliottt:
Build the runtime using `npm run build` in the root instead of running make in `c-dependencies/js-compute-runtime`.


guybedford (approved):
👍  One thing I noticed, is I'm not sure the `npm run build` is running make with parallelism?

elliottt:
> +1 One thing I noticed, is I'm not sure the `npm run build` is running make with parallelism?

It's not and that would be a great change to add in with this PR!

PR #337, opened 2022-12-08
Update to wasi-sdk-16

elliottt:
As wasi-sdk-16 uses a version of llvm that produces uses of `memory.copy`, we need to enable the wizer flag for handling bulk memory.

This PR required bumping our wizer dependency, to get a bug fix for handling data sections https://github.com/bytecodealliance/wizer/pull/61.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,13 +51,13 @@ jobs:
         sudo apt-get update -y
         bash ./build-engine.sh ${{ matrix.profile }}
 
-    - name: "Install wasi-sdk-15 (linux)"
+    - name: "Install wasi-sdk-16 (linux)"
       run: |
         set -x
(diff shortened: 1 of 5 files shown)


JakeChampion (changes requested):
Seeing issues in CI about invalid wasm:
https://github.com/fastly/js-compute-runtime/actions/runs/3650612649/jobs/6167084453

> ERROR: the Fastly API returned 400 Bad Request: Bad request (ERROR: failed to parse WebAssembly module
> 
> Caused by:
>     Invalid input WebAssembly code at offset 6718101: data count and data section have inconsistent lengths).

I'm not sure what would cause this

elliottt:
Let's close this for now. Using wasi-sdk 15 is fine for now, and my guess is that this is related to llvm running `wasm-opt` with options that we have no control over.

elliottt:
After shadowing `wasm-opt` in the path in 321f4f0da482591fcb88b7787dd6fd07a34a89e8, the problems persist. Additionally, it turns out that clang is adding the bulk memory operations, not `wasm-opt`.

JakeChampion (approved):

PR #347, opened 2022-12-13
wit-bindgen updates

guybedford:
This updates the current WIT-bindgen output for the C guest generator, which had the following two PR updates recently:

* https://github.com/bytecodealliance/wit-bindgen/pull/453: Optional params now use nullable pointers. This only affects one API call of ours.
* https://github.com/bytecodealliance/wit-bindgen/pull/452: The boolean result return of functions was changed to return true for success instead of false for success, so we needed to switch every result check. In the process I also streamlined the error handling to use a single `HANDLE_ERROR` function instead of `HANDLE_RESULT` since we have the boolean check on the base component host function, then also to use the direct error code definitions generated by the component and entirely deprecating the `FastlyStatus` enum as well, so the WIT is now a singular source of truth for these definitions.


c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
```diff
@@ -153,21 +153,22 @@ int convert_tag(fastly_http_cache_override_tag_t tag) {
...
-                            h, convert_tag(tag), ttl->is_some ? ttl->val : 0,
-                            stale_while_revalidate->is_some ? stale_while_revalidate->val : 0,
-                            reinterpret_cast<char *>(sk_str.ptr), sk_str.len),
-                        err);
+  return convert_result(
+      xqd_req_cache_override_v2_set(
+          h, convert_tag(tag), maybe_ttl == NULL ? 0 : *maybe_ttl,
+          maybe_stale_while_revalidate == NULL ? 0 : *maybe_stale_while_revalidate,
+          reinterpret_cast<char *>(sk_str.ptr), sk_str.len),
+      err);
```
guybedford, line 171: The new nullable pointers for optional parameters are updated here.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -3030,17 +3050,20 @@ bool get_header_names_from_handle(JSContext *cx, uint32_t handle, Mode mode,
...
+    ok = xqd_fastly_http_req_header_names_get(handle, &ret, &err);
   } else {
-    is_error = xqd_fastly_http_resp_header_names_get(handle, &ret, &err);
+    ok = xqd_fastly_http_resp_header_names_get(handle, &ret, &err);
   }
 
-  if (!HANDLE_RESULT(cx, is_error, err))
+  if (!ok) {
+    HANDLE_ERROR(cx, err);
+    JS_free(cx, buf);
```
guybedford, line 3064: I believe I found a missing free in the process...

elliottt (approved):
This looks good to me!

PR #359, opened 2022-12-23
component: --component option for outputting a component

guybedford:
This adds a new `--component` flag to the runtime CLI, that when provided will output a full component binary for the runtime instead of a core Wasm binary.

The component binary uses the latest version of the wasi-preview2 adapter to adapt the WASI imports, fetched from the releases of https://github.com/bytecodealliance/preview2-prototyping.

Component generation along with WASI 1 -> WASI preview 2 adaption is only after Wizering so that we retain full compat with the current core Wasm Wizer process.

The WASI adapter is applied using a build of wit component provided by `js-component-tools` providing `wasm-tools` for JS, so `js-component-tools` is introduced as a dependency of the runtime in the process (~20MB, no dependencies).

Strictly speaking, this completes the task of generating a js-compute-runtime Wasm component.

Whether this should be merged is a question of if we want to introduce the extra js-component-tools dependency for non component installers. It might not be such an issue, or we could maintain this branch.

Review / feedback / thoughts / comments very welcome.


JakeChampion:
Nice work! This looks good to me 👍 

Is there currently a way to test the component version? If there is, we should add a test and hook it up in our GitHub Workflow. Once Viceroy supports using a component, we can make use of all our existing integration test applications :D

guybedford:
> Is there currently a way to test the component version?

Right now, no there is not, but I'm working on this problem!

PR #366, opened 2023-01-11
feat: implement Request.prototype.clone

JakeChampion:
This commit adds a native implementation for Request.prototype.clone - The clone() method creates a copy of the current Request object.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -189,6 +189,7 @@ jobs:
           - random
           - regex
           - request-cache-key
+          - request-clone
           - request-downstream
           - request-limits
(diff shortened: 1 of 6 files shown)


elliottt (approved):
This looks great!

What do you think about adding a test that awaits the request body for a cloned request, after sending on the original? That would ensure that we're keeping around the correct state after the request has been sent.

JakeChampion:
> This looks great!
> 
> What do you think about adding a test that awaits the request body for a cloned request, after sending on the original? That would ensure that we're keeping around the correct state after the request has been sent.

I'm going to add this test as a separate PR 👍

PR #378, opened 2023-01-20
Introduce allocation wrappers around spidermonkeys allocator api

elliottt:
Introduce a strong symbol for `cabi_realloc` to ensure that we're always calling `JS_realloc`, and some convenience functions for helping to identify when memory is managed by wit-bindgen-generated code.

### DIFF
--- a/c-dependencies/js-compute-runtime/allocator.cpp
+++ b/c-dependencies/js-compute-runtime/allocator.cpp
@@ -1,0 +1,17 @@
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winvalid-offsetof"
+#pragma clang diagnostic ignored "-Wdeprecated-enum-enum-conversion"
+#include "jsapi.h"
+#pragma clang diagnostic pop
+
+JSContext *CONTEXT = nullptr;
+
+extern "C" {
+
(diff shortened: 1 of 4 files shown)


c-dependencies/js-compute-runtime/allocator.cpp
```diff
@@ -0,0 +1,17 @@
...
+#pragma clang diagnostic ignored "-Winvalid-offsetof"
+#pragma clang diagnostic ignored "-Wdeprecated-enum-enum-conversion"
+#include "jsapi.h"
+#pragma clang diagnostic pop
+
+JSContext *CONTEXT = nullptr;
+
+extern "C" {
+
+void *cabi_realloc(void *ptr, size_t orig_size, size_t align, size_t new_size) {
```
guybedford, line 12: Should this be:

```suggestion
__attribute__((weak, export_name("cabi_realloc")))
void *cabi_realloc(void *ptr, size_t orig_size, size_t align, size_t new_size) {
```

Note that it is a `weak` symbol because the separate bindgen file from wit bindgen also defines its own alloc function which is being overridden.

elliottt: I've moved it there, but am intentionally not making it `weak` as I'd like to ensure that the one generated by `wit-bindgen` is overridden by this one during linking.

guybedford (approved):

Issue #217, opened 2022-08-23
Make use of malloc slack when allocating read buffers


elliottt:
`malloc` will sometimes allocate more than requested, and if we can observe this slack we can use it when reading from a handle in `read_from_handle_all`.

PR #379, opened 2023-01-21
Refactor read_from_handle_all

elliottt:
Rework `read_from_handle_all` to avoid realloc'ing for each chunk read. Additionally use unique_ptr to avoid needing any explicit uses of `JS_free`, ensuring that we free memory no matter which path we take when returning.


JakeChampion (approved):
This is really nice work 🙂 this'll resoslve #217

PR #382, opened 2023-01-23
chore: C++ interface to host calls

elliottt:
Start implementing an abstraction layer over the host call interface, so that we can take advantage of smart pointers and other c++ convenience features.


elliottt:
I'm getting a pretty consistent test failure for the e2e `geoip`  test, though it's not clear to me if it's related to my changes.

JakeChampion (approved):

PR #386, opened 2023-01-26
fix: ensure CacheOverride bitflags are the same value as definde in xqd

JakeChampion:
Previously we had defined a `http-cache-override-tag` flag of `none` - which is not a defined flag within xqd. This extra defined flag meant that all our flag values were off-by-one.

This PR removes the `none` flag and also adds `static_asserts` to ensure that our world-adapter values are the same values that we have defined in `cache-override.h`

This PR also removes the `convert_tag` function which was used in the definition of `xqd_fastly_http_req_cache_override_set` when in non-component-mode to convert the provided tag into a single flag value. That behaviour is incorrect as it effectively changes the behavior of the flag into an enum, whereas flags are different to enums because multiple flags can be set at the same time.

I've also gone ahead and written a set of integration tests for the behavior and fixed a few cases where we would cause wasm exceptions to occur (calling the CacheOverride setters/getters directly on the CacheOverride prototype) - we instead throw a JS Error instead of a wasm exception.

The tests can only run correctly on C@E and not on Viceroy, this is because we are observing caching behavior and Viceroy does not implement any caching logic.


c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
```diff
@@ -65,11 +65,10 @@ typedef struct {
 
 typedef uint8_t fastly_http_cache_override_tag_t;
 
-#define FASTLY_HTTP_CACHE_OVERRIDE_TAG_NONE (1 << 0)
```
JakeChampion, line 68: @guybedford @elliottt - this is one of the more important parts of this PR - when we moved to the component model .wit definitions, we accidentally got an off-by-one error for all our cache-override flag values

elliottt: This file is generated by `wit-bindgen`, is there any way we can make the fix in `xqd.wit` instead so that it generates the correct output? Perhaps just removing the `none` variant?

c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
```diff
@@ -157,7 +141,7 @@ bool xqd_fastly_http_req_cache_override_set(fastly_request_handle_t h,
   }
   return convert_result(
       xqd_req_cache_override_v2_set(
-          h, convert_tag(tag), maybe_ttl == NULL ? 0 : *maybe_ttl,
+          h, (int)(tag), maybe_ttl == NULL ? 0 : *maybe_ttl,
```
elliottt, line 144:
```suggestion
          h, static_cast<int>(tag), maybe_ttl == NULL ? 0 : *maybe_ttl,
```

elliottt (commented):
Looks great, nearly ready to merge!

c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
elliottt: Looks like you removed the constructor in xqd.wit already, so this looks good to me!

elliottt (approved):

guybedford (approved):

PR #387, opened 2023-01-26
fix: ensure CacheOverride bitflags are the same value as defined in xqd

JakeChampion:
Previously we had defined a `http-cache-override-tag` flag of `none` - which is not a defined flag within xqd. This extra defined flag meant that all our flag values were off-by-one.

This commit removes the `none` flag and also adds static_asserts to ensure that our world-adapter values are the same values that we have defined in cache-override.h

This commit also remove the convert_tag function which was used in the definition of `xqd_fastly_http_req_cache_override_set` when in non-component-mode to convert the provided tag into a single flag value. That behaviour is incorrect as it effectively changes the behaviour of the flag into an enum, whereas flags are different to enums because multiple flags can be set at the same time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -167,6 +167,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build]
     strategy:
+      fail-fast: false
       matrix:
         fixture: 
           - async-select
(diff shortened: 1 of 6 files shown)


phamann (approved):

guybedford (approved):

guybedford:
CI change should be fixed by https://github.com/fastly/js-compute-runtime/pull/388.

PR #390, opened 2023-01-26
component: reinclude conversion function to avoid world-based binary assumptions

guybedford:
This provides the proper flag-based implementation of `convert_tag`, whereas the previous function which was behind https://github.com/fastly/js-compute-runtime/pull/386 was implemented as if it were an enum object instead.

This avoids assuming flag binary alignment between xqd and the component model.

### DIFF
--- a/c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
+++ b/c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
@@ -127,6 +127,19 @@ bool xqd_fastly_http_req_body_downstream_get(fastly_request_t *ret, fastly_error
   return convert_result(xqd_req_body_downstream_get(&ret->f0, &ret->f1), err);
 }
 
+int convert_tag(fastly_http_cache_override_tag_t tag) {
+  int out_tag = 0;
+  if (tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_PASS > 0)
+    out_tag |= CACHE_OVERRIDE_PASS;
+  if (tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_TTL > 0)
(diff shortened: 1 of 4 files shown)


JakeChampion (approved):

guybedford:
One other thing that would have also helped catch this issue would be proper static typing of the flags and enum types in the component model. Strictly speaking, this was the root cause of the off-by-one as well, which I've included a commit to fix now.

c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
```diff
@@ -127,6 +127,19 @@ bool xqd_fastly_http_req_body_downstream_get(fastly_request_t *ret, fastly_error
...
+int convert_tag(fastly_http_cache_override_tag_t tag) {
+  int out_tag = 0;
+  if ((tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_PASS) > 0)
+    out_tag |= CACHE_OVERRIDE_PASS;
+  if ((tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_TTL) > 0)
+    out_tag |= CACHE_OVERRIDE_TTL;
+  if ((tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_STALE_WHILE_REVALIDATE) > 0)
+    out_tag |= CACHE_OVERRIDE_STALE_WHILE_REVALIDATE;
+  if ((tag & FASTLY_HTTP_CACHE_OVERRIDE_TAG_PCI) > 0)
+    out_tag |= CACHE_OVERRIDE_PCI;
```
elliottt, line 139: Could you add curly braces to all these if statements?

elliottt (approved):
This looks good to me, with the curly braces fixed :)

JakeChampion (approved):

PR #392, opened 2023-01-30
fix: allow lots of the same header name to exist within a request or response

JakeChampion:
Currently a C@E application retrieving multiple headers with the same name will crash due to a wasm out-of-bounds access.

This is due to the amount of memory we allocate to retrieving header values from the host. Looking into our memory allocation, it looks like we forgot to multiply the amount of headers by the size of a `xqd_world_string_t`. I've updated our allocation to account for the size of `xqd_world_string_t` and written a test to confirm this stop the application crashing.

We don't need to worry about headers being greater than this size, as it is the max size supported by c@e/viceroy 👍


elliottt (approved):
Looks good!

guybedford:
Great work tracking this down!

PR #394, opened 2023-01-30
Refactor header_names_get to avoid explicitly freeing memory

elliottt:
Rework `xqd_fastly_http_req_header_names_get` to avoid any need to explicitly free the memory that it manages.

### DIFF
--- a/c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
+++ b/c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
@@ -183,20 +183,20 @@ bool xqd_fastly_http_req_new(fastly_request_handle_t *ret, fastly_error_t *err) 
   return convert_result(xqd_req_new(ret), err);
 }
 
+struct Chunk {
+  JS::UniqueChars buffer;
+  size_t length;
+
+  static Chunk make(std::string_view data) {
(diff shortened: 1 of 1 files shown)


elliottt:
The e2e geoip test is failing, but I think that might be unrelated to this PR.

c-dependencies/js-compute-runtime/xqd-world/xqd_world_adapter.cpp
```diff
@@ -176,50 +185,63 @@ bool xqd_fastly_http_req_new(fastly_request_handle_t *ret, fastly_error_t *err)
...
-      if (buf[i] != '\0')
-        continue;
-      if (str_cnt == str_max) {
-        strs = static_cast<xqd_world_string_t *>(
-            cabi_realloc(strs, str_max * sizeof(xqd_world_string_t), 1,
-                         (str_max + LIST_ALLOC_SIZE) * sizeof(xqd_world_string_t)));
-        str_max += LIST_ALLOC_SIZE;
+  };
+
+  std::vector<Chunk> chunks;
```
guybedford, line 199: Perhaps rather call this `header_names`?

guybedford (approved):
Haven't fully verified every step of the logic, but the overall approach looks great and much safer to me.

guybedford (approved):

JakeChampion (approved):

PR #395, opened 2023-01-30
component: update to latest bindgen

guybedford:
This updates the component model to the latest world bindgen using `js-component-tools@0.3.0-beta.1` and the latest WASI preview2 build.

The latest wit-bindgen also now provides the bug fix for the previous ret area bug we had before as well.


guybedford:
Fixed up the issues, this PR is now just pending the merge and release of https://github.com/bytecodealliance/js-component-tools/pull/28.

src/component.js
```diff
@@ -1,12 +1,10 @@
-import { componentNew } from 'js-component-tools';
+import { componentEmbed, componentNew } from '../../js-component-tools/dist/api.mjs';
```
JakeChampion: This file does not exist inside this codebase - should this be reverted to the package name?
```suggestion
import { componentEmbed, componentNew } from 'js-component-tools';
```

package.json
```diff
@@ -44,7 +44,7 @@
   "dependencies": {
     "@jakechampion/wizer": "^1.6.0",
     "esbuild": "^0.15.16",
-    "js-component-tools": "0.2.2",
+    "js-component-tools": "0.3.0-beta.2",
```
JakeChampion, line 47: question: do we know when this is coming out of beta?

guybedford: It's purely a js-component-tools thing for now, and there's no reason not to already, probably going to give things another couple of weeks to settle.

guybedford:
Not sure what is up with the geoip test here, but this should be good to land I believe.

elliottt:
> Not sure what is up with the geoip test here, but this should be good to land I believe.

Could you rebase on main? I think @JakeChampion has fixed that test.

c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -1,4 +1,4 @@
-world xqd-world {
+default world xqd-world {
```
elliottt, line 1: What's the effect of this change?

elliottt (approved):
This seems reasonable to me!

I read through some of the changes to xqd_world.h, mostly verifying that constants hadn't changed. Do you know why there was so much churn in the output of the header?

JakeChampion (approved):

c-dependencies/js-compute-runtime/xqd.wit
guybedford: The world files were upgraded with a bunch of new semantics including packages and use support. Default worlds and default interfaces are the new way of defining the item in the wit document to use by default since they can contain multiple worlds (otherwise you need to explicitly pass `--world` to specify which world is being generated out of the file).

guybedford:
> Do you know why there was so much churn in the output of the header?

wit-bindgen defines code into source section blocks, and some things were moved around I believe.

PR #401, opened 2023-02-02
component: export-based interface

guybedford:
This returns an exported function for the component build to run the end-to-end component service as a fully encapsulated function call.

### DIFF
--- a/c-dependencies/js-compute-runtime/xqd.wit
+++ b/c-dependencies/js-compute-runtime/xqd.wit
@@ -486,4 +486,8 @@ default world xqd-world {
 
     purge-surrogate-key: func(surrogate-key: string, soft-purge: bool) -> result<purge-result, error>
   }
+
+  export fastly-runtime: interface {
+    serve-sync: func() -> ()
+  }
 }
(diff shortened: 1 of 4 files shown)


c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -486,4 +486,8 @@ default world xqd-world {
 
     purge-surrogate-key: func(surrogate-key: string, soft-purge: bool) -> result<purge-result, error>
   }
+
+  export fastly-runtime: interface {
+    serve-sync: func() -> ()
+  }
```
JakeChampion, line 492: Would it be possible to have this in a different file which imports the xqd world?
The reason I ask is because I think it would be good to keep the xqd definition in a different file to help make it clear that the xqd definitions are not ours (also we should probably name it compute-at-edge instead of xqd 🙈 )

guybedford: Agreed, we could even explicitly name it exerpimental-compute-at-edge or something to indicate it's undergoing change.

guybedford: Agreed, we could even explicitly name it exerpimental-compute-at-edge or something to indicate it's undergoing change.

guybedford:
Will close this out, to rather leave it to @elliottt if they want to implement something along these lines or not.

PR #402, opened 2023-02-02
content-encodings should be a flags

elliottt:
- Content encodings should be a `flags`
- Re-generate the xqd world

### DIFF
--- a/c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
+++ b/c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
@@ -4,13 +4,13 @@
 extern "C" {
 #endif
 
+#include <stdbool.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
-#include <stdint.h>
-#include <stdbool.h>
 
 typedef struct {
-  char*ptr;
+  char *ptr;
   size_t len;
 } xqd_world_string_t;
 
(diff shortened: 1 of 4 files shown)


c-dependencies/js-compute-runtime/xqd-world/xqd_world.h
```diff
@@ -154,7 +154,7 @@ typedef uint32_t fastly_dictionary_handle_t;
 
 typedef uint8_t fastly_content_encodings_t;
 
-#define FASTLY_CONTENT_ENCODINGS_GZIP 0
+#define FASTLY_CONTENT_ENCODINGS_GZIP (1 << 0)
```
elliottt, line 157: Here's the actual change, the rest is churn from wit-bindgen.

c-dependencies/js-compute-runtime/xqd.wit
```diff
@@ -193,7 +193,7 @@ default world xqd-world {
 
     http-req-close: func(h: request-handle) -> result<_, error>
 
-    enum content-encodings {
+    flags content-encodings {
```
guybedford, line 196: If I remember correctly from the `cache_override_tag` case, the first flag in WIT is a 1, while the first flag in WITX was a 0?

Either way though, it would be great if we could ensure we don't assume binary value alignment between the interfaces and actually ensure we're passing the value through a conversion function where it is used.

guybedford (commented):
It seems we have very different C formatters! xqd_world.c diff is likely due to that as I always run it through the formatter.

guybedford (approved):
Just checked the codebase, and this is only used by `http-req-auto-decompress-response-set`, which it appears we don't use at all, therefore that's why we're unaffected by this.

c-dependencies/js-compute-runtime/xqd.wit
elliottt: The first flag will be the 0 bit, which is consistent with witx. Here we're always passing `0` for a content-encodings value, as that's what the first enum value translates to, which corresponds to a value with no flags set in viceroy.

guybedford: If you are sure. I just recall that when we first implemented the cache_tag flags @JakeChampion had to remove the `none` value to get these to align with WITX, which made me think that encoding might have been starting from 0 - https://github.com/fastly/Viceroy/blob/main/lib/compute-at-edge-abi/typenames.witx#L123.

PR #406, opened 2023-02-06
fix: Remove error codes from external error messaging

JakeChampion:
We should not be sending out these error codes as they are not something we guarantee will stay the same and we will possibly move to a different way of modelling internal error in the future


elliottt (approved):
Looks good to me!

PR #407, opened 2023-02-07
component: reactor-based initialization redux

guybedford:
This sets up the world for the component to use an exported interface of the form:

```
export js-compute-runtime: interface {
    type request-handle = u32
    type body-handle = u32
    type request = tuple<request-handle, body-handle>

    /// Serve the given request
    ///
    /// response handle not currently returned, because in the case of a streamed response
    /// send-downstream would still need to be eagerly called, and therefore there would be
    /// no need to return the response, due to the run to completion semantics.
    /// 
    /// serve-specific error codes, outside of HTTP semantics, could be a useful future
    /// addition to this interface, with an empty error placeholder for now (currently always
    /// returns success in the runtime)
    serve: func(req: request) -> result
}
```

Internally, `xqd_req_body_downstream_get` is called immediately on invocation and any errors result in an abort.


elliottt (commented):
I'm a little concerned that the entry-point in the component case isn't calling `_start`, which means that we're potentially avoiding some necessary c runtime setup. It would be good to get on the same page with the folks working on the preview2-prototyping repo before merging this.

guybedford:
I've updated to the latest preview2 reactor build here as well.

elliottt (approved):
This looks good to me!

PR #412, opened 2023-02-08
fix: add custom error message when making a request to a backend which does not exist

JakeChampion:
previously we would return the generic message: "Generic error value. This means that some unexpected error occurred during a hostcall."

Now we will return a TypeError with the message format `Requested backend named '{0}' does not exist` where `{0}` is the name of the backend that was configured for the request.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -314,7 +314,6 @@ jobs:
           - 'error'
           - 'geoip'
           - 'hello-world'
-          - missing-backend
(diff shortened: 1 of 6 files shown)


guybedford (approved):
Is there an XQD issue to track an error code for this?

JakeChampion:
> Is there an XQD issue to track an error code for this?

There is not, I believe the plan is to not introduce new error codes (until we move to components)

PR #426, opened 2023-02-28
fix: modular builtin separation

guybedford:
This fully separates the builtins build by including the `encode` function and also including a custom `GetErrorMessageBuiltin` with separate builtin error message codes so that the error message pipeline for the builtins is completely separated from the Fastly-specific code.

Structuring feedback very welcome, but I can confirm this works for building the console against ComponentizeJS.


JakeChampion (approved):

elliottt (approved):
This looks great, thank you!

PR #430, opened 2023-03-01
chore: Split more builtins out of js-compute-builtins.cpp

elliottt:
Finish splitting out the remaining builtins from `js-compute-builtins.cpp`, and remove the `BOILERPLATE` macros from `builtin.h`.


c-dependencies/js-compute-runtime/builtins/text-encoder.cpp
```diff
@@ -0,0 +1,81 @@
...
+    return false;
+
+  args.rval().setObject(*byte_array);
+  return true;
+}
+
+bool TextEncoder::encoding_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  JS::RootedString str(cx, JS_NewStringCopyN(cx, "utf-8", 5));
```
JakeChampion, line 47: Non-blocking comment: Since this can never change, should we make this a persistent string instead of creating a new string each time? JS Strings are immutable, so a persistent string should work here

c-dependencies/js-compute-runtime/builtins/text-encoder.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace builtins {
+
+bool TextEncoder::encode(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(1)
+
+  // Default to empty string if no input is given.
+  if (args.get(0).isUndefined()) {
+    JS::RootedObject byte_array(cx, JS_NewUint8Array(cx, 0));
+    if (!byte_array)
+      return false;
```
JakeChampion, line 20: Non-Blocking Question/Comment When would this actually occur? Does this only occur if we are out of memory? If so, we should probably report this as a JS InternalError, which is how this is reported in other runtimes. "InternalError: allocation size overflow"

c-dependencies/js-compute-runtime/builtins/text-decoder.cpp
```diff
@@ -0,0 +1,64 @@
...
+    return false;
+
+  args.rval().setString(str);
+  return true;
+}
+
+bool TextDecoder::encoding_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  JS::RootedString str(cx, JS_NewStringCopyN(cx, "utf-8", 5));
```
JakeChampion, line 32: non-blocking: Same here as in text-encoder.cpp, this could be a persistent string instead of a newly created string on each call

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -4712,16 +758,16 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
   }
 
   RootedObject requestInstance(
-      cx, JS_NewObjectWithGivenProto(cx, &Request::class_, Request::proto_obj));
+      cx, JS_NewObjectWithGivenProto(cx, &builtins::Request::class_, builtins::Request::proto_obj));
```
guybedford, line 761: Wondering why we don't just use everything from the `builtins` namespace by default?

guybedford (approved):

c-dependencies/js-compute-runtime/js-compute-builtins.h
```diff
@@ -52,7 +54,7 @@ inline bool ReturnPromiseRejectedWithPendingError(JSContext *cx, const JS::CallA
 std::optional<std::span<uint8_t>> value_to_buffer(JSContext *cx, JS::HandleValue val,
                                                   const char *val_desc);
 
-typedef bool InternalMethod(JSContext *cx, JS::HandleObject receiver, JS::HandleValue extra,
+using InternalMethod = bool(JSContext *cx, JS::HandleObject receiver, JS::HandleValue extra,
```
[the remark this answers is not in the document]
elliottt: I find `typedef` yields really difficult to understand type synonyms, particularly with function pointers. `using` more closely maps how type synonyms work in different languages, and I thought I would make this change while I was touching this file.

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
elliottt: We definitely could. Historically it was to not clash with things defined in the root namespace of `js-compute-builtins.cpp`.

c-dependencies/js-compute-runtime/builtins/text-decoder.cpp
elliottt: I think that's a good change, but let's do it in a follow :+1:

c-dependencies/js-compute-runtime/builtins/text-encoder.cpp
elliottt: I think that's a good change, but let's do it in a follow :+1:

c-dependencies/js-compute-runtime/builtins/text-encoder.cpp
elliottt: I think that's a totally reasonable change. We should take a pass over the builtins now that they're all split out of js-compute-builtins.cpp and look for similar patterns to fix.

JakeChampion (approved):

PR #431, opened 2023-03-01
deps: upgrade to @bytecodeallianace/wizer@1.6.1-beta.4

guybedford:
This includes the fix from https://github.com/bytecodealliance/wizer/pull/71, as well as moving over the bytecodealliance org version on npm.


guybedford:
(if someone can do another check on the output size difference, when I tested it I got a 2 byte difference, but it would help to verify this!)

JakeChampion (approved):

JakeChampion:
```
Before this change : 12,249,849 bytes
After this change  : 12,155,371 bytes

Difference         :    -94,478 bytes
```

PR #432, opened 2023-03-01
builtins: refactor shared console implementation

guybedford:
This refactors the console implementation as a shared builtin, based on providing an implementation for the function:

```
void builtin_impl_console_log(builtins::Console::LogType log_ty, const char *msg);
```

I'm open to suggestions re naming / namespacing of the above, as ideally this is a pattern we can use for builtin implementations going forward for others.


c-dependencies/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -150,46 +150,52 @@ JS::Result<mozilla::Ok> SetToSource(JSContext *cx, std::string &sourceOut, JS::H
...
+    sourceOut += "{ ";
+  }
   bool firstValue = true;
   for (size_t i = 0; i < length; ++i) {
     const auto &id = ids[i];
     if (!JS_GetPropertyById(cx, obj, id, &value)) {
       return JS::Result<mozilla::Ok>(JS::Error());
     }
 
-    if (!value.isObject() || !JS_ObjectIsFunction(&value.toObject())) {
```
guybedford, line 168: It would still be good to understand the reason for this guard originally.

JakeChampion (approved):

PR #438, opened 2023-03-06
builtins: more shared builtins

guybedford:
This extends the shared builtins to `TextEncoder`, `TextDecoder` and also `URL`. I can confirm these all work in ComponentizeJS against this PR.

I tried building a single combined archive, but with cross-platform differences in `ar` and the fact that these archives are all rooted, wasn't able to come up with an easy approach. This seems to work fine though.


elliottt (approved):
Looks great to me!

PR #440, opened 2023-03-08
feat: Add support for SubtleCrypto

JakeChampion:
This adds an implementation of a subset of SubtleCrypto, specifically for JSON Web Token signing and validating.
- SubtleCrypto.prototype.generateKey
- SubtleCrypto.prototype.importKey
- SubtleCrypto.prototype.sign
- SubtleCrypto.prototype.verify with the following algorithms:
- RSASSA_PKCS1_v1_5
- RSA_OAEP and the following digest algorithms:
- SHA_1
- SHA_224
- SHA_256
- SHA_384
- SHA_512

Work in the future will be done to add the remaining algorithms and SubtleCrypto method implementations


c-dependencies/js-compute-runtime/js-compute-builtins.h
```diff
@@ -85,6 +85,107 @@ bool define_fastly_sys(JSContext *cx, JS::HandleObject global);
...
+} // namespace FetchEvent
+
+namespace GlobalProperties {
+// Maps an encoded character to a value in the Base64 alphabet, per
+// RFC 4648, Table 1. Invalid input characters map to UINT8_MAX.
+// https://datatracker.ietf.org/doc/html/rfc4648#section-4
+
+constexpr uint8_t nonAlphabet = 255;
+
+static const uint8_t base64DecodeTable[] = {
```
elliottt, line 130: Could you move this back to `js-compute-builtins.cpp` and make this declaration `extern const uint8_t base64DecodeTable[];` instead?

c-dependencies/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -341,105 +341,27 @@ JS::PersistentRooted<js::UniquePtr<ScheduledTimers>> timers;
...
   }
-  *value = base64DecodeTable[index & mask];
+  *value = decodeTable[index & mask];
 
   return *value != 255;
 }
 
-inline JS::Result<mozilla::Ok> base64Decode4to3(std::string_view input, std::string &output) {
+inline JS::Result<mozilla::Ok> base64Decode4to3(std::string_view input, std::string &output,
+                                                const uint8_t *decodeTable) {
```
elliottt, line 357: What drove making the decode table a parameter here?

tests/wpt-harness/expectations/WebCryptoAPI/algorithm-discards-context.https.window.js.json
```diff
@@ -0,0 +1,41 @@
+{
+  "Context is discarded in generateKey": {
+    "status": "FAIL"
+  },
```
elliottt, line 4: Are these all expected to fail with this pr?

c-dependencies/js-compute-runtime/builtins/crypto-key.h
```diff
@@ -0,0 +1,942 @@
+#pragma once
```
elliottt, line 1: Could you switch to header guards? If we ever change build systems, `#pragma once` could become problematic.

JakeChampion:
Changing this to a draft as I will be breaking the PR up into a bunch of smaller PRs

panva:
If I could lobby for a few more algorithms to be supported

- RSA-PSS
- ECDSA (at least P-256, but ideally all three curves)
- Ed25519 from https://wicg.github.io/webcrypto-secure-curves/

stefan-guggisberg:
@JakeChampion We are using 3rd party dependencies relying on subtle.crypto, e.g. for signing AWS requests.

We're still getting errors like `Supplied algorithm is not yet supported`.

Here's a code fragment from https://github.com/mhart/aws4fetch which seems to trigger this error:

```
async function hmac(key2, string3) {
  const cryptoKey = await crypto.subtle.importKey(
    "raw",
    typeof key2 === "string" ? encoder2.encode(key2) : key2,
    { name: "HMAC", hash: { name: "SHA-256" } },
    false,
    ["sign"]
  );
  return crypto.subtle.sign("HMAC", cryptoKey, encoder2.encode(string3));
}
```

It seems like the lack of HMAC support is causing the error.

This is a blocker for us as it. Without this support we won't be able to use C@E.

Can you give us an ETA when this PR finally lands in production?

mlegenhausen:
For a complete move to compute edge we would need support for the following API

* `crypto.subtle.encrypt`
* `crypto.subtle.decrypt`

Both with support for `AES-CBC` and `AES-GCM`

JakeChampion:
Thanks all -- we've added HMAC and raw importkey support and I have opened issues for the other requests which have been added to this draft pull-request.

I will be closing this pull-request as it is never going to be merged, all the work is happening in other, smaller pull-requests 👍

Issue #446, opened 2023-03-10
bug: regression in 1.5.0


domdomegg:
After upgrading to 1.5.0, we get the error:

```
> fastly-project@1.0.0 prebuild
> webpack

[same webpack output]

webpack 5.75.0 compiled with 3 warnings in 4932 ms

> fastly-project@1.0.0 build
> js-compute-runtime bin/index.js bin/main.wasm

fastly_project/node_modules/regjsparser/parser.js:1571
      throw SyntaxError(message + ' at position ' + from + (details ? ': ' + details : '') + '\n' + context + '\n' + pointer);
            ^

SyntaxError: Expected atom at position 7
    ((?=lg)?[vl]k\-?\
           ^
    at bail (fastly_project/node_modules/regjsparser/parser.js:1571:13)
    at parseTerm (fastly_project/node_modules/regjsparser/parser.js:584:11)
    at parseAlternative (fastly_project/node_modules/regjsparser/parser.js:548:21)
    at parseDisjunction (fastly_project/node_modules/regjsparser/parser.js:528:16)
    at finishGroup (fastly_project/node_modules/regjsparser/parser.js:622:18)
    at parseGroup (fastly_project/node_modules/regjsparser/parser.js:618:14)
    at parseAtomAndExtendedAtom (fastly_project/node_modules/regjsparser/parser.js:786:16)
    at parseTerm (fastly_project/node_modules/regjsparser/parser.js:575:18)
    at parseAlternative (fastly_project/node_modules/regjsparser/parser.js:548:21)
    at parseDisjunction (fastly_project/node_modules/regjsparser/parser.js:528:16)
```

This appears to be coming from interactions with the library `ua-parser-js`, which previously worked on fastly.

Specifically, it seems to be this regex to detect LG devices:
```
/((?=lg)?[vl]k\-?\d{3}) bui| 3\.[-\w; ]{10}lg?-([06cv9]{3,4})/i
```

https://github.com/faisalman/ua-parser-js/blob/3c3c03ceeba9920437533dcf1d72276acac091a3/src/ua-parser.js#L464

domdomegg:
Related: https://github.com/jviereck/regjsparser/issues/130

PR #447, opened 2023-03-10
fix: handle fallthrough of regex parser bugs

guybedford:
Resolves https://github.com/fastly/js-compute-runtime/issues/446.

While the root cause fix is https://github.com/jviereck/regjsparser/issues/130, we shouldn't expose parser bugs, and rather fall through to the engine, as well as in the case of an invalid regex more generally.


elliottt (approved):
Nice!

PR #454, opened 2023-03-14
chore: make geo tests not test the timezone offsets as they change when daylight savings is toggled

JakeChampion:


elliottt (approved):
Thanks for fixing this!

PR #460, opened 2023-03-22
chore: patch random functions in openssl to use arc4random_buf when compiled to wasi

JakeChampion:
without this change, the random functions in openssl do not work when compiled to wasi and instead return an int indicating an error


elliottt (approved):
Looks great!

PR #462, opened 2023-03-23
chore: refactor how we implement crypto algorithms to make it simpler to add new algorithm implementations in the future

JakeChampion:
Instead of having everything defined in subtle-crypto.cpp/.h - we are breaking the algorithms out into their own file and creating base classes which all the algorthims will derive from.
    
Eventually there will be base classes for all the different algorithm operations such as importing/exporting, encrypting/decrypting, signing/veryifing, wraping/unrwapping etc.
    
This refactor also helps keep subtle-crypto.cpp to contain only the logic required for following the web-crypto specification


elliottt (approved):
This looks great!

PR #467, opened 2023-03-24
feat: Implement JS CryptoKey Interface

JakeChampion:
Spec: https://w3c.github.io/webcrypto/#cryptokey-interface
    
This implements the basic interface of a CryptoKey:
- Exposing a public constructor which throws an JS Error when called
- Defining the interface/instance members (https://w3c.github.io/webcrypto/#cryptokey-interface-members)
- Defining all the internal slots that are required for the interface

This class cannot currently be used by a JS application due to the fact CryptoKey can only be constructed by the SubtleCrypto methods `generateKey` and `importKey`, which are not yet implemented. Future pull-requests will implement those methods, and also include tests to confirm the behavior of the CryptoKey interface

### DIFF
--- a/c-dependencies/js-compute-runtime/builtins/crypto-key.h
+++ b/c-dependencies/js-compute-runtime/builtins/crypto-key.h
@@ -9,18 +9,31 @@ namespace builtins {
 
 enum class CryptoKeyType : uint8_t { Public, Private, Secret };
 
-enum {
-  CryptoKeyUsageEncrypt = 1 << 0,
-  CryptoKeyUsageDecrypt = 1 << 1,
-  CryptoKeyUsageSign = 1 << 2,
-  CryptoKeyUsageVerify = 1 << 3,
-  CryptoKeyUsageDeriveKey = 1 << 4,
-  CryptoKeyUsageDeriveBits = 1 << 5,
-  CryptoKeyUsageWrapKey = 1 << 6,
-  CryptoKeyUsageUnwrapKey = 1 << 7
-};
+class CryptoKeyUsages {
+private:
+  uint8_t mask;
+  static constexpr const uint8_t encrypt_flag = 1 << 0;
+  static constexpr const uint8_t decrypt_flag = 1 << 1;
+  static constexpr const uint8_t sign_flag = 1 << 2;
+  static constexpr const uint8_t verify_flag = 1 << 3;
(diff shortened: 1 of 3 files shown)


c-dependencies/js-compute-runtime/builtins/crypto-key.h
```diff
@@ -0,0 +1,79 @@
...
+  CryptoKeyUsageDecrypt = 1 << 1,
+  CryptoKeyUsageSign = 1 << 2,
+  CryptoKeyUsageVerify = 1 << 3,
+  CryptoKeyUsageDeriveKey = 1 << 4,
+  CryptoKeyUsageDeriveBits = 1 << 5,
+  CryptoKeyUsageWrapKey = 1 << 6,
+  CryptoKeyUsageUnwrapKey = 1 << 7
+};
+
+typedef int CryptoKeyUsageBitmap;
```
elliottt, line 23:
```suggestion
using CryptoKeyUsageBitmap = int;
```

elliottt (commented):
This is looking good! What do you think about making the bitmap type a standalone class? That might be overkill looking at how it's used though. Perhaps moving the enum from the header into the anonymous namespace in `crypto-key.cpp` and turning the values into `constexpr int` instead?

c-dependencies/js-compute-runtime/builtins/crypto-key.h
```diff
@@ -9,18 +9,57 @@ namespace builtins {
...
+  static constexpr const uint8_t derive_key_flag = 1 << 4;
+  static constexpr const uint8_t derive_bits_flag = 1 << 5;
+  static constexpr const uint8_t wrap_key_flag = 1 << 6;
+  static constexpr const uint8_t unwrap_key_flag = 1 << 7;
 
-typedef int CryptoKeyUsageBitmap;
+public:
+  CryptoKeyUsages(uint8_t mask) { this->mask = mask; };
+  CryptoKeyUsages(bool encrypt, bool decrypt, bool sign, bool verify, bool derive_key,
+                  bool derive_bits, bool wrap_key, bool unwrap_key) {
```
elliottt, line 27: Could you move the constructors to crypto-key.cpp?

elliottt (approved):
This looks great! Just one comment about the location for the constructors of `CryptoKeyUsages`.

PR #472, opened 2023-03-29
 feat: Implement subset of crypto.subtle.importKey which can import a JSONWebKey using RSASSA-PKCS1-v1_5

JakeChampion:


c-dependencies/js-compute-runtime/builtins/json-web-key.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+namespace {
+JS::Result<std::optional<std::string>>
+extractStringPropertyFromObject(JSContext *cx, JS::HandleObject object, std::string_view property) {
+  bool has_property;
+  if (!JS_HasProperty(cx, object, property.data(), &has_property)) {
+    return JS::Result<std::optional<std::string>>(JS::Error());
+  }
+  if (!has_property) {
+    return std::optional<std::string>(std::nullopt);
```
elliottt, line 28: Does this change work, or did you run into template argument deduction problems?
```suggestion
    return std::nullopt;
```

c-dependencies/js-compute-runtime/builtins/json-web-key.cpp
```diff
@@ -0,0 +1,353 @@
...
+          }
+
+          // No duplicates allowed
+          if (std::find(key_ops.begin(), key_ops.end(), op) != key_ops.end()) {
+            JS_ReportErrorASCII(
+                cx, "The 'key_ops' member of the JWK dictionary contains duplicate usages");
+            return nullptr;
+          }
+
+          key_ops.push_back(op);
```
elliottt, line 250: Using `push_back` here will copy the string, but `emplace_back` will give you the opportunity to invoke the move constructor for the target cell in the vector.
```suggestion
          key_ops.emplace_back(std::move(op));
```

c-dependencies/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -183,6 +440,300 @@ std::unique_ptr<CryptoAlgorithmDigest> CryptoAlgorithmDigest::normalize(JSContex
...
+    // contains an entry which is not "verify" then throw a SyntaxError.
+    bool isUsagesAllowed = false;
+    // public key
+    if (jwk->d.has_value()) {
+      isUsagesAllowed = usages.canSign();
+    } else {
+      // private key
+      isUsagesAllowed = usages.canVerify();
+    }
+    isUsagesAllowed = isUsagesAllowed || !usages.isEmpty();
```
elliottt, line 541: Why would usages be allowed if `usages` was not empty? The way I'm reading the comment it seems like the two cases are just those of the conditional above.

elliottt (commented):
This is looking great!

c-dependencies/js-compute-runtime/builtins/json-web-key.cpp
JakeChampion: This runs into the deduction problem you've mentioned -- 

>  no viable conversion from returned value of type 'const std::nullopt_t' to function return type 'JS::Result<std::optional<std::string>>

c-dependencies/js-compute-runtime/builtins/json-web-key.cpp
elliottt: I partially take this back: there's no need to use `emplace_back` here, `push_back(std::move(op))` will get the same effect :+1:

elliottt (approved):
This looks great! Just a few suggestions before merging :+1:

c-dependencies/js-compute-runtime/builtins/crypto-key.cpp
```diff
@@ -234,4 +328,194 @@ bool CryptoKey::init_class(JSContext *cx, JS::HandleObject global) {
...
+    if (keyData->privateExponent.empty() || keyData->firstPrimeInfo->primeFactor.empty() ||
+        keyData->secondPrimeInfo->primeFactor.empty()) {
+      return nullptr;
+    }
+  }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+  auto rsa = RSA_new();
+#pragma clang diagnostic pop
```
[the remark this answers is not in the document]
JakeChampion: Issue at https://github.com/fastly/js-compute-runtime/issues/477

c-dependencies/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -3,11 +3,258 @@
...
+    JS_ReportErrorLatin1(cx,
+                         "The JWK member 'n' could not be base64url decoded or contained padding");
+    return nullptr;
+  }
+  auto modulus = modulusResult.unwrap();
+  // Per RFC 7518 Section 6.3.1.1: https://tools.ietf.org/html/rfc7518#section-6.3.1.1
+  if (modulus.starts_with('0')) {
+    modulus = modulus.erase(0, 1);
+  }
+  auto dataResult = GlobalProperties::ConvertJSValueToByteString(cx, jwk->e.value());
```
JakeChampion, line 77:
```suggestion
  auto dataResult = GlobalProperties::convertJSValueToByteString(cx, jwk->e.value());
```

stefan-guggisberg:
@JakeChampion Thanks for landing this PR in 1.7.0.

I was hoping this would solve our problem (3rd party dependencies relying on `subtle.crypto` e.g. for signing AWS requests. However, we're now seeing different errors, such as `Supplied algorithm is not yet supported`.

Here's a code fragment from https://github.com/mhart/aws4fetch which seems to trigger this error:

```javascript
async function hmac(key2, string3) {
  const cryptoKey = await crypto.subtle.importKey(
    "raw",
    typeof key2 === "string" ? encoder2.encode(key2) : key2,
    { name: "HMAC", hash: { name: "SHA-256" } },
    false,
    ["sign"]
  );
  return crypto.subtle.sign("HMAC", cryptoKey, encoder2.encode(string3));
}
```

It seems like the lack of `HMAC` support is causing the error. 

Would this PR solve our problem? https://github.com/fastly/js-compute-runtime/pull/440

Can you give us an ETA when this PR lands in production?

JakeChampion:
@stefan-guggisberg - #440 does solve that issue 👍  - I am breaking that PR up into smaller more manageable PR for the team to review. If you subscribe to #440, when it is closed you should be able to run the code you've posted without any issues 👍

PR #478, opened 2023-04-06
feat: Make the KVStore.prototype.get implementation be async

JakeChampion:
Note: This will not pass on CI until a version of Viceroy _and_ Compute have implementations of the async host-calls.

We model the JS API for ObjectStore methods as async promise-returning methods however, the internal implementation is currently synchronous due to the host-call itself being synchronous.

This PR changes the host-call we use to be the (soon to be real) asynchronous host-call for ObjectStore lookups.

The great thing about this change, is it is not a user-facing API change as we were already returning a Promise. This PR makes that Promise resolution happen asynchronously ☺️

### DIFF
--- a/c-dependencies/js-compute-runtime/builtins/object-store.cpp
+++ b/c-dependencies/js-compute-runtime/builtins/object-store.cpp
@@ -183,31 +183,34 @@ bool parse_and_validate_key(JSContext *cx, JS::UniqueChars *key, size_t len) {
 
 } // namespace
 
-bool ObjectStore::get(JSContext *cx, unsigned argc, JS::Value *vp) {
-  METHOD_HEADER(1)
+fastly_pending_object_store_lookup_handle_t ObjectStore::pending_lookup_handle(JSObject *self) {
+  MOZ_ASSERT(ObjectStore::is_instance(self));
 
-  JS::RootedObject result_promise(cx, JS::NewPromiseObject(cx, nullptr));
-  if (!result_promise) {
-    return ReturnPromiseRejectedWithPendingError(cx, args);
+  JS::Value handle_value =
+      JS::GetReservedSlot(self, static_cast<uint32_t>(Slots::PendingLookupHandle));
+
+  if (handle_value.isInt32()) {
+    return handle_value.toInt32();
   }
 
-  JS::RootedValue key(cx, args.get(0));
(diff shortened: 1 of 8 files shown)


c-dependencies/js-compute-runtime/fastly.wit
```diff
@@ -423,6 +424,10 @@ default world fastly-world {
     // Should object store insert return "inserted" bool?
     object-store-insert: func(store: object-store-handle, key: string, body-handle: body-handle) -> result<_, error>
 
+    object-store-lookup-async: func(store: object-store-handle, key: string) -> result<pending-object-store-lookup-handle, error>
+
+    object-store-lookup-wait: func(h: pending-object-store-lookup-handle) -> result<option<body-handle>, error>
```
guybedford, line 429: Just wondering if there is a poll variation for this (I guess for `option<option<body-handle>>`)?

c-dependencies/js-compute-runtime/builtins/object-store.cpp
```diff
@@ -175,6 +175,54 @@ bool parse_and_validate_key(JSContext *cx, JS::UniqueChars *key, size_t len) {
...
+
+bool ObjectStore::process_pending_object_store_lookup(JSContext *cx, JS::HandleObject self) {
+  MOZ_ASSERT(ObjectStore::is_instance(self));
+
+  fastly_option_body_handle_t ret;
+  fastly_error_t err;
+  bool ok = fastly_object_store_lookup_wait(builtins::ObjectStore::pending_lookup_handle(self),
+                                            &ret, &err);
+  if (!ok) {
+    HANDLE_ERROR(cx, err);
```
guybedford, line 199: Should this reject the promise?

guybedford (approved):

c-dependencies/js-compute-runtime/fastly.wit
JakeChampion: We can use the fastly_async.is_ready hostcall, is that an equivalent of what you are describing?

c-dependencies/js-compute-runtime/builtins/object-store.cpp
JakeChampion: Yes this should reject rather than create an exception, good work spotting this bug!

PR #480, opened 2023-04-09
feat: Add high-resolution timing function "fastly.now()" behind feature flag "--enable-experimental-high-resolution-time-methods"

JakeChampion:
This function accepts no arguments and returns the number of Microseconds since the epoch, midnight, January 1, 1970 UTC.

This function is very useful to have for running performance tests within JavaScript applications.

Still TODO:
- [ ] Decide whether these experimental features should be documented on the reference docs website
- [ ] Decide whether they should also be documented in the TypeScript Definitions

### DIFF
--- a/c-dependencies/js-compute-runtime/js-compute-builtins.h
+++ b/c-dependencies/js-compute-runtime/js-compute-builtins.h
@@ -81,7 +81,28 @@ bool hasWizeningFinished();
 bool isWizening();
 void markWizeningAsFinished();
 
-bool define_fastly_sys(JSContext *cx, JS::HandleObject global);
+class FastlyOptions {
+private:
+  uint8_t mask = 0;
+
+public:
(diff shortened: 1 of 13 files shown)


c-dependencies/js-compute-runtime/js-compute-builtins.h
```diff
@@ -81,7 +81,28 @@ bool hasWizeningFinished();
 bool isWizening();
 void markWizeningAsFinished();
 
-bool define_fastly_sys(JSContext *cx, JS::HandleObject global);
+class FastlyOptions {
+private:
+  uint8_t mask = 0;
```
elliottt, line 86: I would probably make booleans for each option instead of masking. We won't make many of these values, so we probably don't need to resort to bit-packing here :)

elliottt (approved):
This looks great to me! I just had one comment about simplifying the options class a bit, but it's not a blocker.

PR #485, opened 2023-04-11
fix: Lower the supported NodeJS version from 18 or greater to only 18

JakeChampion:
Node 19 is [not supported by tree-sitter yet](https://github.com/tree-sitter/node-tree-sitter/pull/127), and that is a dependency of ours. Until tree-sitter works on Node 19, we should not list Node 19 as supported.


guybedford:
Hmm, this is a serious issue - can we maybe move to another parser that supports it? Node.js 20 goes out this month and that will be very important to support.

JakeChampion:
> Hmm, this is a serious issue - can we maybe move to another parser that supports it? Node.js 20 goes out this month and that will be very important to support.

We can probably fork and patch with the linked PR to bring support on for newer Node versions

Issue #484, opened 2023-04-10
`?` appended to URLs if `searchParams` methods are accessed


twhid:
Greetings,

I've noticed that if `searchParams`  methods are accessed on an instance of a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) then a question mark is always appended to the url.

To reproduce: 

Use this handler in an empty JS starter kit:
```js
/// <reference types="@fastly/js-compute" />

addEventListener("fetch", (event) => event.respondWith(handleRequest(event)));

async function handleRequest(event) {
    const urlObj = new URL('http://www.example.com');
    console.log('href 1 = ', urlObj.href);
    urlObj.searchParams.sort();
    console.log('href 2 = ', urlObj.href);
    return new Response("OK", { status: 200 });
}

```

Then, run `fastly compute serve` and request `http://127.0.0.1:7676` 

Console output:
```shell
INFO: Listening on http://127.0.0.1:7676

INFO: Command output:
--------------------------------------------------------------------------------
Log: href 1 =  http://www.example.com/
Log: href 2 =  http://www.example.com/?
```

I'll admit to having not read the spec 😉 but, this behavior doesn't appear to be to spec. Other JS runtime examples:

A file like this in node v18.15.0:
```js
const http = require('http');
const port = 8080;
const server = http.createServer((req, res) => {
    const reqUrl = new URL('http://www.example.com:8080');
    console.log('href 1 = ', reqUrl.href);
    reqUrl.searchParams.sort();
    console.log('href 2 = ', reqUrl.href);
    res.end('OK');
});
server.listen(port);
console.log(`Node server listening on ${port}`);
```

Outputs:
```
Node server listening on 8080
href 1 =  http://www.example.com:8080/
href 2 =  http://www.example.com:8080/
```

And the Firefox and Chrome consoles output: 
<img width="385" alt="image" src="https://user-images.githubusercontent.com/299945/230948905-cfb6a801-2adf-4491-924d-5c14233cbf6c.png">

+++

Working around this apparent bug creates more complexity than is required for certain scenarios.

Thank you for any feedback.

JakeChampion:
Thank yo for providing a small amount of code which shows the issue 🙏 This Fiddle uses the code you have provided and confirms this is happening https://fiddle.fastly.dev/fiddle/5cffcaa3 -- I can also confirm that this is a bug in our URL implementation 👍 

The bug is in this function:
[...]

PR #487, opened 2023-04-11
fix: When converting a URL to a string, do not add a `?` if there are no query string parameters

JakeChampion:
fixes https://github.com/fastly/js-compute-runtime/issues/484

### DIFF
--- a/c-dependencies/js-compute-runtime/rust-url/src/lib.rs
+++ b/c-dependencies/js-compute-runtime/rust-url/src/lib.rs
@@ -36,13 +36,23 @@ impl JSUrlSearchParams {
         match self.url_or_str {
             UrlOrString::Url(url) => {
                 let url = unsafe { url.as_mut().unwrap() };
-                let mut pairs = url.url.query_pairs_mut();
-                pairs.clear().extend_pairs(self.list.iter());
+                if self.list.is_empty() {
+                    url.url.set_query(None);
+                } else {
+                    let mut pairs = url.url.query_pairs_mut();
+                    pairs.clear();
+                    pairs.extend_pairs(self.list.iter());
+                }
             }
             UrlOrString::Str(_) => {
-                let str = form_urlencoded::Serializer::new(String::new())
-                    .extend_pairs(&*self.list)
-                    .finish();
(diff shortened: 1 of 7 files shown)


c-dependencies/js-compute-runtime/rust-url/src/lib.rs
```diff
@@ -36,13 +36,26 @@ impl JSUrlSearchParams {
         match self.url_or_str {
             UrlOrString::Url(url) => {
                 let url = unsafe { url.as_mut().unwrap() };
-                let mut pairs = url.url.query_pairs_mut();
-                pairs.clear().extend_pairs(self.list.iter());
+                match self.list.is_empty() {
+                    true => {
+                        url.url.set_query(None);
```
JakeChampion, line 41: This is the crux of the bug-fix.
Previously we used `url.query_pairs_mut().clear()` however `url.query_pairs_mut().clear();` is equivalent to `url.set_query(Some(""))`, not `url.set_query(None)` and we really wanted `.set_query(None)`

guybedford (approved):

c-dependencies/js-compute-runtime/rust-url/src/lib.rs
```diff
@@ -36,13 +36,26 @@ impl JSUrlSearchParams {
         match self.url_or_str {
             UrlOrString::Url(url) => {
                 let url = unsafe { url.as_mut().unwrap() };
-                let mut pairs = url.url.query_pairs_mut();
-                pairs.clear().extend_pairs(self.list.iter());
+                match self.list.is_empty() {
```
elliottt, line 39: Could you switch this to an `if-then-else`, as it's matching a `bool` value?

elliottt (approved):
Looks good to me, great find!

twhid:
Hi all -- is this live on fastly production services?

JakeChampion:
> Hi all -- is this live on fastly production services?

Howdy, yes version 1.7.1 has this included 👍  https://github.com/fastly/js-compute-runtime/blob/main/CHANGELOG.md#171-2023-04-11

PR #492, opened 2023-04-12
fix: Mark NodeJS 19 and 20 as supported

JakeChampion:
Thanks to the work done by @guybedford  in https://github.com/fastly/js-compute-runtime/pull/488

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@fastly/js-compute",
   "version": "1.8.0",
   "engines": {
-    "node": "16 - 18",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@fastly/js-compute",
   "version": "1.8.0",
   "engines": {
-    "node": "16 - 18",
+    "node": "16 - 19",
```
JakeChampion, line 5:
```suggestion
    "node": "16 - 20",
```

PR #494, opened 2023-04-13
 feat: Implement subset of crypto.subtle.verify which can verify a signature with a JSONWebKey using RSASSA-PKCS1-v1_5

JakeChampion:
With this, we have all the methods in place for JWT usage in JS for Fastly Compute :D


c-dependencies/js-compute-runtime/builtins/crypto-key.cpp
```diff
@@ -644,4 +644,10 @@ bool CryptoKey::canSign(JS::HandleObject self) {
   return usage.canSign();
 }
 
+bool CryptoKey::canVerify(JS::HandleObject self) {
+  MOZ_ASSERT(is_instance(self));
+  auto usage = CryptoKeyUsages(JS::GetReservedSlot(self, Slots::Usages).toInt32());
```
elliottt, line 649: Similarly to #491, should this check that the `int` fits in a `uint8_t`?

elliottt (approved):
Looks good to me!

PR #497, opened 2023-04-15
feat: add ability to automatically decompress gzip responses returned from `fetch`

JakeChampion:
This patch add a new `fastly` options object to the RequestInit parameter on `fetch` and the `Request` constructor. The new `fastly` options object has one field, `decompress`, if this field is falsey, then no automatic decompression is applied on the response. If this field is truthy, then automatic decompression is applied on the response.

This patch brings in support for automatically decompressing GZIP and not other compression formats.

I've added this under a field named `fastly` as that can be our extension point for other Fastly specific request/response features.

Previously we've added Fastly specific features as a root property on the RequestInit object, such as `backend`, `cacheOverride`, and `cacheKey`. In future patches we could move those to exist under `fastly` instead of the root.

### DIFF
--- a/documentation/docs/globals/fetch.mdx
+++ b/documentation/docs/globals/fetch.mdx
@@ -56,6 +56,9 @@ fetch(resource, options)
       - *Fastly-specific*
     - `cacheOverride` _**Fastly-specific**_
     - `cacheKey` _**Fastly-specific**_
+    - `fastly` _**Fastly-specific**_
+      - `decompressGzip`_: boolean_ _**optional**_
+        - Whether to automatically gzip decompress the Response or not.
 
 ### Return value
 
--- a/documentation/docs/globals/Request/Request.mdx
+++ b/documentation/docs/globals/Request/Request.mdx
@@ -41,3 +41,6 @@ new Request(input, options)
     - `backend` _**Fastly-specific**_
     - `cacheOverride` _**Fastly-specific**_
     - `cacheKey` _**Fastly-specific**_
+    - `fastly` _**Fastly-specific**_
+      - `decompressGzip`_: boolean_ _**optional**_
+        - Whether to automatically gzip decompress the Response or not.
--- a/integration-tests/js-compute/fixtures/request-auto-decompress/bin/index.js
+++ b/integration-tests/js-compute/fixtures/request-auto-decompress/bin/index.js
@@ -1,0 +1,105 @@
+/* eslint-env serviceworker */
+import { env } from 'fastly:env';
+import { pass, fail, assert, assertRejects } from "../../../assertions.js";
+
+addEventListener("fetch", event => {
+    event.respondWith(app(event))
+})
+/**
+ * @param {FetchEvent} event
+ * @returns {Response}
+ */
+function app(event) {
+    try {
+        const path = (new URL(event.request.url)).pathname;
+        console.log(`path: ${path}`)
+        console.log(`FASTLY_SERVICE_VERSION: ${env('FASTLY_SERVICE_VERSION')}`)
+        if (routes.has(path)) {
+            const routeHandler = routes.get(path);
+            return routeHandler()
+        }
+        return fail(`${path} endpoint does not exist`)
+    } catch (error) {
+        return fail(`The routeHandler threw an error: ${error.message}` + '\n' + error.stack)
+    }
+}
+
+const routes = new Map();
+routes.set('/', () => {
+    routes.delete('/');
+    let test_routes = Array.from(routes.keys())
+    return new Response(JSON.stringify(test_routes), { 'headers': { 'content-type': 'application/json' } });
+});
+
+// Request.fastly.decompressGzip option -- automatic gzip decompression of responses
+{
+    routes.set("/request/constructor/fastly/decompressGzip/true", async () => {
+        const request = new Request('https://httpbin.org/gzip', {
+            headers: {
+                accept: 'application/json'
+            },
+            backend: "httpbin",
+            fastly: {
+                decompressGzip: true
+            }
+        });
+        const response = await fetch(request);
+        // This should work because the response will be decompressed and valid json.
+        const body = await response.json();
+        let error = assert(body.gzipped, true, `body.gzipped`)
+        if (error) { return error }
+        return pass()
+    });
+    routes.set("/request/constructor/fastly/decompressGzip/false", async () => {
+        const request = new Request('https://httpbin.org/gzip', {
+            headers: {
+                accept: 'application/json'
+            },
+            backend: "httpbin",
+            fastly: {
+                decompressGzip: false
+            }
+        });
+        const response = await fetch(request);
+        let error = await assertRejects(async function() {
+            // This should throw because the response will be gzipped compressed, which we can not parse as json.
+            await response.json();
+        });
+        if (error) { return error }
+        return pass()
+    });
+
+    routes.set("/fetch/requestinit/fastly/decompressGzip/true", async () => {
+        const response = await fetch('https://httpbin.org/gzip', {
+            headers: {
+                accept: 'application/json'
+            },
+            backend: "httpbin",
+            fastly: {
+                decompressGzip: true
+            }
+        });
+        // This should work because the response will be decompressed and valid json.
+        const body = await response.json();
+        let error = assert(body.gzipped, true, `body.gzipped`)
+        if (error) { return error }
+        return pass()
+    });
+    routes.set("/fetch/requestinit/fastly/decompressGzip/false", async () => {
+        const response = await fetch('https://httpbin.org/gzip', {
+            headers: {
+                accept: 'application/json'
+            },
+            backend: "httpbin",
+            fastly: {
+                decompressGzip: false
+            }
+        });
+        let error = await assertRejects(async function() {
+            // This should throw because the response will be gzipped compressed, which we can not parse as json.
+            await response.json();
+        });
+        if (error) { return error }
+        return pass()
+    });
+}
--- a/integration-tests/js-compute/fixtures/request-auto-decompress/fastly.toml.in
+++ b/integration-tests/js-compute/fixtures/request-auto-decompress/fastly.toml.in
@@ -1,0 +1,23 @@
+# This file describes a Fastly Compute@Edge package. To learn more visit:
+# https://developer.fastly.com/reference/fastly-toml/
+
+authors = ["me@jakechampion.name"]
+description = ""
+language = "other"
+manifest_version = 2
+name = "request-auto-decompress"
+service_id = ""
+
+[scripts]
+  build = "node ../../../../js-compute-runtime-cli.js"
+
+[local_server]
+  [local_server.backends]
+    [local_server.backends.httpbin]
+      url = "https://httpbin.org/"
+
+[setup]
+  [setup.backends]
+    [setup.backends.httpbin]
+      address = "httpbin.org"
+      port = 443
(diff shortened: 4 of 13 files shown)


c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -1929,6 +1949,24 @@ JSObject *Request::create(JSContext *cx, JS::HandleObject requestInstance, JS::H
         JS::GetReservedSlot(input_request, static_cast<uint32_t>(Slots::CacheOverride)));
   }
 
+  if (fastly_val.isObject()) {
+    JS::RootedValue decompress_response_val(cx);
+    JS::RootedObject fastly(cx, fastly_val.toObjectOrNull());
```
guybedford, line 1954: I think this should just be `toObject()`. Note that `isObject()` in Spidermonkey API does not include null.

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -1141,7 +1141,25 @@ bool Request::set_cache_override(JSContext *cx, JS::HandleObject self,
...
+  MOZ_ASSERT(is_instance(self));
+  auto decompress =
+      JS::GetReservedSlot(self, static_cast<uint32_t>(Request::Slots::AutoDecompress)).toBoolean();
+  if (!decompress) {
+    return true;
+  }
 
+  fastly_error_t err;
+  fastly_content_encodings_t encodings_to_decompress = 0;
+  encodings_to_decompress |= FASTLY_CONTENT_ENCODINGS_GZIP;
```
guybedford, line 1155: Why not also include Brotli?

guybedford (commented):
As far as I'm aware, browsers and edge runtimes typically automatically handle decompression in fetch, so that setting this flag would allow us to align with the native behaviour.

As such, I wonder if we should possibly flip the default in a future major version?

Also if we add new decompression algorithms in future, in theory that is a breaking change. Perhaps we should try and add a comprehensive list of algorithms to match browsers to start so that we can immediately get the alignment?

The fact that adding a new compression algorithm is a breaking change regardless makes me think we should at the very least include brotli here as well... or is there some blocker on that?

c-dependencies/js-compute-runtime/builtins/request-response.cpp
JakeChampion: We can't toObject here because we are in the embedded API which does not have a concrete definition of JSObject which we can import. We are guaranteed to return a JSObject pointer though, due to the isObject call above.

c-dependencies/js-compute-runtime/builtins/request-response.cpp
JakeChampion: It isn't supported by the host, the host only supports gzip (for now)

JakeChampion:
> As such, I wonder if we should possibly flip the default in a future major version?

The next major is soon, so we can wait on merging this and ship this with that major.

> Also if we add new decompression algorithms in future, in theory that is a breaking change. Perhaps we should try and add a comprehensive list of algorithms to match browsers to start so that we can immediately get the alignment?
> 
> The fact that adding a new compression algorithm is a breaking change regardless makes me think we should at the very least include brotli here as well... or is there some blocker on that?

The host environment currently only supports gzip. We could block this pr on waiting for brotli on the host (I can likely add brotli support into the host)

c-dependencies/js-compute-runtime/builtins/request-response.cpp
guybedford: Ah right, of course there are API differences!

guybedford:
I do think the developer experience and migration paths for this feature would be much simpler if we landed this alongside brotli as well. How much work would it be to add Brotli support?

guybedford:
Thinking about this further - if we only send `accept-encoding: 'gzip'` currently I suppose there's no conflict to amending that later on? Are `accept-encoding` headers automatically populated by the host?

JakeChampion:
> Thinking about this further - if we only send `accept-encoding: 'gzip'` currently I suppose there's no conflict to amending that later on? Are `accept-encoding` headers automatically populated by the host?

The host does not add `accept-encoding` headers automatically, they need to be provided - for requests which come from client web browsers, the header will exist already - for requests which are made within the guest via `new Request`, the header will need to be added manually by the guest.

What this host-call does (when called with it's GZIP flag) is add some metadata to the request, if the response for the request contains a `content-encoding` header which contains `gzip` or `x-gzip` (as per [8.4.1.3](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.8.4.1.3)) then the response body is decompressed and the content-encoding and content-length headers for the response are removed.

>I do think the developer experience and migration paths for this feature would be much simpler if we landed this alongside brotli as well. How much work would it be to add Brotli support?

I wouldn't think too much work, the implementation of this host-call was made with adding other compression encodings.

>As far as I'm aware, browsers and edge runtimes typically automatically handle decompression in fetch, so that setting this flag would allow us to align with the native behaviour.
>
>As such, I wonder if we should possibly flip the default in a future major version?

I've remembered why we likely don't want to have this on by default with the current implementation of the host-call. The challenge here is that we don't want to auto-decompress all responses because some (I would say the vast vast majority of) responses go straight through to the client, and we don't want to decompress responses going to the client as it'd make the client response be larger than necessary. That's the reason we have this as a manual setting.

It'd be pretty great if this could be on by default and the decompression only happens lazily when the `fastly_http_body.read` host-call is called. If it were implemented that way, then responses forwarded straight to the client will not be decompressed by the host and instead forwarded as is; And if the response's body is ready by the guest via whatever means, the host would decompress the contents first. That isn't what currently happens and would be a bigger piece of work to implement in the host code-base.

guybedford:
The problem is, I believe what we are doing is not actually compatible with other edge runtimes at all... that is Cloudflare does decompress the response automatically just like a browser does. So this will naturally become a standard compat issue over time.

My suggestion was therefore to make sure the flag effectively defaults to the current browser / edge behaviour exactly, such that we effectively only have two modes - auto decompress, or not. Where the only difference between our implementation and other implementations is that we default to retaining the compressed stream. But setting this fastly specific flag then exactly matches the compat paths. In theory that flag default could be switched, and we could make the Fastly compressed handling opt-in instead of opt-out.

Just bringing the compat perspecive, I completely understand the user perspective too though, in fact one reason I couldn't personally use cloudflare for an app in the past was due to wanting transparent brotli pass through...!

JakeChampion:
What we currently have is compatible with the `fetch` specification, because the "codings" are defined by the implementer and not the specification and we currently support no "codings".

[4.6. HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch) step 10.1.1.1.4:
>Set bytes to the result of [handling content codings](https://fetch.spec.whatwg.org/#handle-content-codings) given codings and bytes.

And [handle content codings](https://fetch.spec.whatwg.org/#handle-content-codings):
>To handle content codings given _codings_ and _bytes,_ run these steps:
>1. If _codings_ are not supported, then return _bytes_.
>2. Return the result of decoding _bytes_ with _codings_ as explained in HTTP, if decoding does not result in an error, and failure otherwise. [[HTTP]](https://fetch.spec.whatwg.org/#biblio-http)

I will mention the automatic decompression only if `fastly_http_body.read` host-call is called idea to the Compute@Edge platform team and see what they think to that.

And for my clarification - are we wanting to block this PR until Brotli support is added? Are there other things we are wanting to block this PR on? I'm trying to make sure I've understood it all and what the next steps are 👍

guybedford:
> I will mention the automatic decompression only if fastly_http_body.read host-call is called idea to the Compute@Edge platform team and see what they think to that.

Does this mean that reading the body stream in the fetch API would branch the behaviour? That sounds a little too special-cased to me?

> And for my clarification - are we wanting to block this PR until Brotli support is added?

Are there urgent factors to landing this, or can we work out the compat path a little more? My suggestion is to bundle Brotli (and possibly deflate) yes exactly that's all I'm considering here.

guybedford:
Here's one way to frame the concern - is adding a new decompression algorithm supported under `decompress: true` a breaking change to the runtime?

JakeChampion:
> > I will mention the automatic decompression only if fastly_http_body.read host-call is called idea to the Compute@Edge platform team and see what they think to that.
> 
> Does this mean that reading the body stream in the fetch API would branch the behaviour? That sounds a little too special-cased to me?

Yes that's what it would mean, I thought it might be a good solution because reading a Response's body makes the body become unusable as part of the Response as it will be locked, which means that the same response can't be passed through to the client and also read by the guest.

>Are there urgent factors to landing this, or can we work out the compat path a little more? My suggestion is to bundle Brotli (and possibly deflate) yes exactly that's all I'm considering here.

Nothing urgent, applications can currently decompress GZIP bodies themselves using the `DecompressionStream` class we implemented in the runtime. 

>Here's one way to frame the concern - is adding a new decompression algorithm supported under `decompress: true` a breaking change to the runtime?

I would say yes it's a breaking change 👍 There are currently 8 HTTP Content Codings registered:

Name | Description | Reference | Notes
-- | -- | -- | --
aes128gcm | AES-GCM encryption with a 128-bit content encryption key | [RFC8188] |  
br | Brotli Compressed Data Format | [RFC7932] |  
compress | UNIX "compress" data format [Welch, T., "A Technique          for High Performance Data Compression", IEEE Computer 17(6), June 1984.] | [RFC9110] | Section 8.4.1.1
deflate | "deflate" compressed data ([RFC1951])          inside the "zlib" data format ([RFC1950]) | [RFC9110] | Section 8.4.1.2
exi | W3C Efficient XML Interchange | [W3C Recommendation: Efficient XML          Interchange (EXI) Format] |  
gzip | GZIP file format [RFC1952] | [RFC9110] | Section 8.4.1.3
identity | Reserved | [RFC9110] | Section 12.5.3
pack200-gzip | Network Transfer Format for Java Archives | [JSR 200: Network Transfer Format for Java][Kumar_Srinivasan][John_Rose] |  
x-compress | Deprecated (alias for compress) | [RFC9110] | Section 8.4.1.1
x-gzip | Deprecated (alias for gzip) | [RFC9110] | Section 8.4.1.3
zstd | A stream of bytes compressed using the Zstandard         protocol | [RFC8878] |  


I agree with you that we should try and implement `deflate` and `br` support before landing this 👍 I don't think we worry about the other HTTP Content Coding until someone has a a use-case for them.

guybedford (approved):
This has been stalled for long enough that I'm changing my feedback.

Specifically - changing the option to something like `decompress: 'gzip'` or `decompressGzip: true` (where in future the current option could be added) should resolve the initial concern.

documentation/docs/globals/fetch.mdx
```diff
@@ -56,6 +56,9 @@ fetch(resource, options)
       - *Fastly-specific*
     - `cacheOverride` _**Fastly-specific**_
     - `cacheKey` _**Fastly-specific**_
+    - `fastly` _**Fastly-specific**_
+      - `decompressGzip`_: boolean_ _**optional**_
```
guybedford, line 60: Should we also add this in our `types` folder as well? Or did you want to keep it "experimental"? We can still type it even as experimental I think?

guybedford (approved):

documentation/docs/globals/fetch.mdx
JakeChampion: Yes definitely, I have added that in [`df063dd` (#497)](https://github.com/fastly/js-compute-runtime/pull/497/commits/df063ddc249d9116ee49860165dc7801a5bcf6ca)

PR #498, opened 2023-04-18
feat: implement Response.redirect static method and Response.prototype.redirected getter

JakeChampion:
This patch adds the ability for our builtin classes to have static methods and static properties 

This patch also adds our first static method, `Response.redirect` -- https://fetch.spec.whatwg.org/#dom-response-redirect

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -234,6 +234,7 @@ jobs:
           - request-upstream
           - response
           - response-headers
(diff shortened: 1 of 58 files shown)


elliottt (approved):
This looks great! Sorry that the refactoring to add those new static fields touched everything, perhaps we can figure out a better way to work around that :)

guybedford (approved):

PR #499, opened 2023-04-19
feat: implement Response.json static method

JakeChampion:


JakeChampion:
This is blocked on getting the new JS::ToJSON method from https://github.com/fastly/js-compute-runtime/pull/502

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2384,8 +2384,162 @@ bool Response::redirect(JSContext *cx, unsigned argc, JS::Value *vp) {
   return true;
 }
 
+namespace {
+static bool callbackCalled;
+static bool write_json_to_buf(const char16_t *str, uint32_t strlen, void *out) {
+  callbackCalled = true;
+  auto outstr = static_cast<std::u16string *>(out);
+  outstr->reserve(1 + outstr->length() + strlen);
+  outstr->append(str, strlen);
```
elliottt, line 2393: I think you don't need the `1+` in the reserve here. There's no `0` written to the end, and the call to `shrink_to_fit` on line 2419 is going to remove it through a reallocation anyway. At that point, you can drop the call to `reserve` entirely, as that will happen automatically in the `append`.

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2384,8 +2384,162 @@ bool Response::redirect(JSContext *cx, unsigned argc, JS::Value *vp) {
   return true;
 }
 
+namespace {
+static bool callbackCalled;
+static bool write_json_to_buf(const char16_t *str, uint32_t strlen, void *out) {
```
elliottt, line 2389: These don't need to be `static`, as they're already in the anonymous namespace.
```suggestion
bool callbackCalled;
bool write_json_to_buf(const char16_t *str, uint32_t strlen, void *out) {
```

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2384,8 +2384,162 @@ bool Response::redirect(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  auto make_res = HttpBody::make();
+  if (auto *err = make_res.to_err()) {
+    HANDLE_ERROR(cx, *err);
+    return false;
+  }
+
+  auto body = make_res.unwrap();
+  JS::RootedString string(cx, JS_NewUCStringCopyN(cx, out.c_str(), out.length()));
+  size_t encoded_len;
+  auto stringChars = encode(cx, string, &encoded_len);
```
elliottt, line 2478: Not in this PR, but we should figure out a better api for `encode` to avoid the copy to `string`.

elliottt (approved):
Looks great! Just a couple of suggestions before merging :)

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2384,8 +2384,162 @@ bool Response::redirect(JSContext *cx, unsigned argc, JS::Value *vp) {
   return true;
 }
 
+namespace {
+bool callbackCalled;
+bool write_json_to_buf(const char16_t *str, uint32_t strlen, void *out) {
+  callbackCalled = true;
+  auto outstr = static_cast<std::u16string *>(out);
+  outstr->reserve(1 + outstr->length() + strlen);
```
JakeChampion, line 2392:
```suggestion
```

c-dependencies/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2384,8 +2384,161 @@ bool Response::redirect(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  // 1. Let bytes the result of running serialize a JavaScript value to JSON bytes on data.
+  callbackCalled = false;
+  if (!JS::ToJSON(cx, data, replacer, space, &write_json_to_buf, &out)) {
+    return false;
+  }
+  if (!callbackCalled) {
+    JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_RESPONSE_JSON_INVALID_VALUE);
+    return false;
+  }
+  out.shrink_to_fit();
```
JakeChampion, line 2418:
```suggestion
```

PR #501, opened 2023-04-20
fix(TextDecoder): add (nearly) full support for TextDecoder and TextEncoder

JakeChampion:
we currently only support utf-8 but were not erroring if being supplied a label which was not for the utf-8 encoding -- this patch adds support for (nearly) all the encoding formats for TextDecoder and TextEncoder. This also adds support for calling TextDecoder with the `fatal`/`ignoreBOM`/`stream` options.


c-dependencies/js-compute-runtime/builtins/shared/text-decoder.cpp
```diff
@@ -17,10 +18,33 @@ bool TextDecoder::decode(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+      JS::GetReservedSlot(self, static_cast<uint32_t>(TextDecoder::Slots::Fatal)).toBoolean();
+  auto ignoreBOM =
+      JS::GetReservedSlot(self, static_cast<uint32_t>(TextDecoder::Slots::IgnoreBOM)).toBoolean();
+
+  JS::RootedString str(cx);
+  if (ignoreBOM) {
+    auto chars = JS::UTF8Chars((char *)data->data(), data->size());
+    size_t length;
+    auto lossyUTF8Chars = JS::LossyUTF8CharsToNewTwoByteCharsZ(cx, chars, &length, js::MallocArena);
+    JS::UniqueTwoByteChars lossyChars(static_cast<char16_t *>(JS_malloc(cx, length + 1)));
```
elliottt, line 31: Is the `length + 1` necessary? You're passing the length into `JS_NewUCStringDontDeflate`, which suggests it doesn't need to be null-terminated?

guybedford:
Amazing, will be very happy to see `encodeInto` supported! What is the size effect of this PR? I did a quick test on Deno and Node.js, and they don't support this range of decodings themselves. Are they all necessary? I can do a more formal analysis of supported decodings between runtimes if it helps. Just think we should always weigh up the cost of things against how important the compatibility is.

guybedford:
> Version 1.11.0 is 21.7 MB
> This PR is 22.1 MB

This may seem small, but these numbers can add up fast when you consider the breadth of specifications we are implementing. It's important to keep a close eye on.

By the definition of must support there, both Node.js and Deno are in violation of the spec, so there's clearly some mismatch of expectations somewhere.

Not blocking, just thinking about the future. And noting that sizes can get out of hand if we don't actively think about binary size with all feature development, and make careful implementation decisions above and beyond just "meeting the spec".

JakeChampion:
> I can do a more formal analysis of supported decodings between runtimes if it helps.

@guybedford  -- if you could, that'd be great, let's add what others also have and we can look to add more as and when they are requested by users

guybedford (approved):

PR #502, opened 2023-04-20
Update to SpiderMonkey 112

JakeChampion:
I thought I did the 12,000+ commit rebase successfully of SpiderMonkey 110 to 112 (with our patches)...

It looks like perhaps I was not successful as we have some new test failures for the streams implementation

### DIFF
--- a/c-dependencies/js-compute-runtime/builtins/transform-stream.cpp
+++ b/c-dependencies/js-compute-runtime/builtins/transform-stream.cpp
@@ -146,8 +146,9 @@ bool pipeThrough(JSContext *cx, unsigned argc, JS::Value *vp) {
 
 bool initialize_additions(JSContext *cx, JS::HandleObject global) {
   JS::RootedValue val(cx);
-  if (!JS_GetProperty(cx, global, "ReadableStream", &val))
+  if (!JS_GetProperty(cx, global, "ReadableStream", &val)) {
     return false;
+  }
   JS::RootedObject readableStream_builtin(cx, &val.toObject());
 
   if (!JS_GetProperty(cx, readableStream_builtin, "prototype", &val)) {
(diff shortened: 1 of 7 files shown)


guybedford:
It seems like the issue here is for:

```js
Object.defineProperty(Object.prototype, 'start', get() { return 'start' });
```

which is throwing:

```
Unhandled rejection with value: object "TypeError: can't define property "start": Object is not extensible"
```

So somehow, the global intrinsics are being frozen, which sounds like `setFreezeBuiltins` somehow changed.

I tried adding to `js-compute-runtime.cpp`:

```
JS::RealmOptions realm_options;
realm_options.creationOptions().setFreezeBuiltins(false);
```

But that doesn't seem to resolve the issue. Seems like we should figure this out...

c-dependencies/js-compute-runtime/js-compute-runtime.cpp
```diff
@@ -117,13 +117,8 @@ bool init_js() {
 
   // TODO: check if we should set a different creation zone.
   JS::RealmOptions options;
-  options.creationOptions()
-      .setStreamsEnabled(true)
-      .setReadableByteStreamsEnabled(true)
-      .setBYOBStreamReadersEnabled(true)
-      .setReadableStreamPipeToEnabled(true)
-      .setWritableStreamsEnabled(true)
```
JakeChampion, line 125: These have all been merged into the streamsEnabled option 👍

c-dependencies/js-compute-runtime/builtins/transform-stream.cpp
```diff
@@ -222,7 +224,7 @@ bool ExtractStrategy(JSContext *cx, JS::HandleValue strategy, double default_hwm
     if (!JS::ToNumber(cx, val, hwm)) {
       return false;
     }
-    if (mozilla::IsNaN(*hwm) || *hwm < 0) {
+    if (std::isnan(*hwm) || *hwm < 0) {
```
JakeChampion, line 227: Between Mozilla Gecko 110 and 112, they removed mozilla::IsNaN - I've replaced the call with its `std` equivalent, which is also what is now used in Gecko's codebase

JakeChampion:
@guybedford and @elliottt -- The upgrade is now fully working, all the same tests pass as before the upgrade ☺️

elliottt (approved):
Awesome! :tada:

PR #505, opened 2023-04-21
feat: Add MD5 support into crypto.subtle.digest

JakeChampion:
MD5 is not part of the WebCrypto API standard as a recommended algorithm to implement but we are adding support for it, which is allowed by the Web Crypto API.

The reason we are adding it is because some customers use legacy systems that require MD5 and it is better for performance if we implement it in native code rather than the customer implementing it is JS


guybedford (approved):

elliottt (approved):
Looks good!

PR #514, opened 2023-04-28
Rename c-dependencies to runtime

elliottt:
Rename the top-level `c-dependencies` directory to `runtime`, and update all references. This is mostly to improve discoverability of the sources, if someone is casually browing the repository contents.


guybedford (approved):
Definitely much nicer. We could even consider fully flattening this folder into the base of the repo.

elliottt:
It's really not obvious to me why the dependency review is failing. There are not changes to Cargo.lock for either `rust-url` or `rust-encoding`, so this should behave the same as it does on main.

guybedford:
Perhaps this is a conflict with the `runtime` name and the "fail on scopes" in https://github.com/fastly/js-compute-runtime/blob/main/.github/workflows/dependencies.yml#LL15C9-L15C32 ?

ci/clang-format.ignore
```diff
@@ -1,2 +1,3 @@
-c-dependencies/js-compute-runtime/third_party/wizer.h
-c-dependencies/js-compute-runtime/rust-url/rust-url.h
+runtime/js-compute-runtime/third_party/wizer.h
+runtime/js-compute-runtime/rust-url/rust-url.h
+runtime/js-compute-runtime/rust-encoding/rust-encoding.h
```
JakeChampion, line 3: thank you for adding this 🙏

PR #519, opened 2023-05-05
feat: implement Fanout support for JS SDK

JakeChampion:
Below is a product description of Fanout:

------

Fanout is a publish/subscribe message broker operating at the Fastly edge, powered by [Pushpin](https://pushpin.org/). This makes it easy to build and scale realtime/streaming applications which push data instantly to browsers, mobile apps, servers, and other devices.

Without Fanout, clients make short-lived HTTP requests to Fastly, and Fastly responds with a resource from cache or an origin server. When a request is passed to Fanout, it keeps that connection open indefinitely, assigns the connection to one or more channels, and waits for data to be published to those channels. Origin servers or Fastly services may use an API to publish data to a channel, and Fanout will distribute that data to all the clients which have a subscription to that channel.

Any regular HTTP request may be upgraded into an event-driven response using a transport such as Server-Sent-Events, or Long-Polling. If the request from the client is a WebSocket request, the resulting stream is bidirectional.

This architecture has some nice advantages over more proprietary streaming services:

- Integrated into your Fastly service, so any pre-processing you do to requests to your Fastly service (e.g. for authentication), you can do to streaming requests too.
- Any HTTP client or server can be used - including serverless / functions-as-a-service.
- Any HTTP response can be turned into an evented stream (e.g. a progressive JPEG, a log file, or an API endpoint that performs asynchronous operations).
- No need to use a separate domain for streaming data.

-----

Most JavaScript Compute Services have an entry function which looks like this:
```js
addEventListener("fetch", event => event.respondWith(handleRequest(event)));
```
Where `handleRequest` contains the logic for the application.

With this in mind, I wanted to design a Fanout API which works with this type of entry function.
What I landed on was a function which returns a JS Response, this JS Response has some internal slots set to indicate it's a Fanout response, if the JS Response is then given to `event.respondWith` then instead of calling `fastly_http_resp_send_downstream` we call `fastly_http_req_redirect_to_grip_proxy` which is the Fanout hostcall.

This is what it looks like in an application:
```js
import { createFanoutHandoff } from "fastly:fanout";

  try {
    const url = new URL(event.request.url);
    if (url.pathname === '/stream') {
      return createFanoutHandoff(event.request, 'fanout');
    } else {
      return new Response('oopsie, make a request to /stream for some fanout goodies', { status: 404 });
    }
  } catch (error) {
    console.error({error});
    return new Response(error.message, {status:500})
  }
}
```


runtime/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1129,8 +1129,8 @@ bool process_pending_request(JSContext *cx, HandleObject request) {
     return false;
   }
 
-  RootedObject response(
-      cx, builtins::Response::create(cx, response_instance, response_handle, body, true));
+  RootedObject response(cx, builtins::Response::create(cx, response_instance, response_handle, body,
+                                                       true, false, nullptr));
```
elliottt, line 1133: One pattern that I like for boolean arguments like this is to create a local with a more informative name, and then use that as the argument.
```suggestion
  bool is_upstream = true;
  bool is_grip_upgrade = false;
  RootedObject response(cx, builtins::Response::create(cx, response_instance, response_handle, body,
                                                       is_upstream, is_grip_upgrade, nullptr));
```

runtime/js-compute-runtime/builtins/fastly.cpp
```diff
@@ -130,6 +131,70 @@ bool Fastly::includeBytes(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+    JS_ReportErrorUTF8(cx, "createFanoutHandoff: Backend parameter can not be an empty string");
+    return false;
+  }
+
+  if (length > 254) {
+    JS_ReportErrorUTF8(cx, "createFanoutHandoff: name can not be more than 254 characters");
+    return false;
+  }
+
+  char *buf = new char[length + 1];
```
elliottt, line 181: What do you think about making this a `JS::UniqueChars` instead? That way we'd ensure that ownership is transferred to the request, or that we'd clean it up if we failed to create the response. Additionally, we'd then have an explicit point where we're leaking the memory in `Response::create`, when we would be releasing the memory to write the pointer into the js object that holds it.

elliottt (commented):
I had a couple of comments, but otherwise this looks ready to go!

runtime/js-compute-runtime/builtins/fastly.cpp
elliottt: Additionally, make sure to write the `0` byte at the end if this needs to be null-terminated.

JakeChampion: `JS::UniqueChars` with `std::move` and `.release()` is a great idea 👍
We can use the `backend_chars` variable above which is a `JS::UniqueChars` of the backend name

runtime/js-compute-runtime/js-compute-builtins.cpp
JakeChampion: This is the correct pattern for sure 👍 
I then started to wonder if we could/should use the named parameters pattern -- https://pdimov.github.io/blog/2020/09/07/named-parameters-in-c20/

runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2745,6 +2762,10 @@ JSObject *Response::create(JSContext *cx, JS::HandleObject response,
   JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::Redirected), JS::FalseValue());
   JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::IsUpstream),
                       JS::BooleanValue(is_upstream));
+  JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::IsGripUpgrade),
+                      JS::BooleanValue(is_grip));
+  JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::GripBackend),
+                      JS::PrivateValue(std::move(backend.release())));
```
elliottt, line 2768: Could you move this assignment to right before the `return`? That way if the error return on line 2775 happens, the pointer will still get freed.

elliottt (approved):
Looks good to me, with the one comment about moving the slot initialization to right before the return addressed :+1:

Issue #520, opened 2023-05-09
Build failed with SpiderMonkey release library


fuxiaohei:
Follow this doc https://github.com/fastly/js-compute-runtime/blob/main/DEVELOPMENT.md. It show errors below:

![image](https://user-images.githubusercontent.com/2142787/237030994-5338cefb-ea5f-41a4-b0fc-556de9925c57.png)

I find it compiles SpiderMonkey in CI workflow, not use release library. 

I use `(cd runtime/spidermonkey && sh build-engine.sh release)` to get SpiderMoneky in local. Everything is ok.

Is the `SpiderMonkey release library` broken ?

JakeChampion:
Thanks for the report @fuxiaohei -- I have opened a pull-request to fix this now 🙇 https://github.com/fastly/js-compute-runtime/pull/521

PR #521, opened 2023-05-09
chore: use same rust version as https://github.com/fastly/spidermonkey-wasi-embedding/

JakeChampion:
They drifted apart because spidermonkey-wasi-embedding does not have a rust-toolchain file, which meant the pre-built spidermonkey-wasi would no longer link due to having conflicting/duplicate symbols.

This patch updates our rust-toolchain to match the version used in spidermonkey-wasi-embedding - I have also added the same file to the spidermonkey-wasi-embedding project ot ensure they are on the same version.

Fixes https://github.com/fastly/js-compute-runtime/issues/520

### DIFF
--- a/runtime/rust-toolchain.toml
+++ b/runtime/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.65.0"
+channel = "1.68.2"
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Thanks for fixing this!

PR #522, opened 2023-05-10
feat: Implement all the web console methods

JakeChampion:
Corrects the function lengths for all the methods to be `0` instead of `1` as all the functions parameters are optional.

Adds a Symbol.toStringTag property with a value of `"console"`.

Fixes the prototype chain for `globalThis.console` to match the specification.

Bringing in implementations for:
- console.assert -- This returns true is the first parameter is truthy. If first param is falsey then this writes to Error
- console.clear - This is a no-op :D 
- console.count - Increments a counter for the provided label, if no label provided, then the label "default" is used.
- console.countReset - Resets a counter for the provided label, if no label provided, then the label "default" is used.
- console.dir - Exactly the same as console.log
- console.dirxml - Exactly the same as console.log
- console.group - Exactly the same as console.log
- console.groupCollapsed - Exactly the same as console.log
- console.groupEnd - This is a no-op :D 
- console.table - Exactly the same as console.log
- console.time - Adds the provided label to the timer map for use with timeLog or timeEnd, if no label provided, then the label "default" is used.
- console.timeLog - Logs the current value of a timer that was previously started by calling console.time().
- console.timeEnd - The console.timeEnd() stops a timer that was previously started by calling console.time().
- console.trace - Logs a stack trace.


runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  auto startTime = timer_map.extract(label).mapped();
+
+  // 4. Let duration be a string representing the difference between the current time and startTime,
+  // in an implementation-defined format.
+  auto duration = JS_Now() - startTime;
+  // 5. Let concat be the concatenation of label, U+003A (:), U+0020 SPACE, and duration.
+  std::string concat = label;
+  concat += ": ";
+  concat += std::to_string(duration);
+  concat += "ns";
```
guybedford, line 697: Typically this seems to work in seconds (node.js) or milliseconds (chrome / deno). With nanoseconds, this is 9 orders of magnitude smaller than milliseconds. Perhaps we should do some rounding or at least add a decimal place?

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
-    JS_FN("error", (console_out<Console::LogType::Error>), 1, JSPROP_ENUMERATE),
+    JS_FN("assert", assert, 0, JSPROP_ENUMERATE),
+    JS_FN("clear", no_op, 0, JSPROP_ENUMERATE),
+    JS_FN("count", count, 0, JSPROP_ENUMERATE),
+    JS_FN("countReset", countReset, 0, JSPROP_ENUMERATE),
+    JS_FN("debug", (console_out<Console::LogType::Debug>), 0, JSPROP_ENUMERATE),
+    JS_FN("dir", dir, 0, JSPROP_ENUMERATE),
+    JS_FN("dirxml", (console_out<Console::LogType::Log>), 0, JSPROP_ENUMERATE),
+    JS_FN("error", (console_out<Console::LogType::Error>), 0, JSPROP_ENUMERATE),
+    JS_FN("group", (console_out<Console::LogType::Log>), 0, JSPROP_ENUMERATE),
```
guybedford, line 793: Node.js and Deno have an indentation counter for grouping, although doesn't seem to be that valuable a feature to skip for now to me.

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  if (!CaptureCurrentStack(cx, &stack, JS::StackCapture(JS::MaxFrames(1u << 7)))) {
+    return false;
+  }
+  auto principals = JS::GetRealmPrincipals(JS::GetCurrentRealmOrNull(cx));
+  JS::RootedString str(cx);
+  if (!BuildStackString(cx, principals, stack, &str)) {
+    return false;
+  }
+  size_t length;
+  auto stack_string = encode(cx, str, &length);
```
guybedford, line 748: Great!

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+      // strip quotes for direct string logs
+      if (source[0] == '"' && source[source.length() - 1] == '"') {
+        source = source.substr(1, source.length() - 2);
+      }
+      if (message.length()) {
+        message += " ";
+        message += source;
+      } else {
+        message += source;
+      }
```
guybedford, line 478: This pattern is repeated quite a lot, perhaps let's just refactor it into a separate function?

guybedford (approved):
Great stuff

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  size_t count;
+  // 2. If map[label] exists, set map[label] to map[label] + 1.
+  if (count_map.contains(label)) {
+    count = count_map.at(label);
+    count += 1;
+  } else {
+    // 3. Otherwise, set map[label] to 1.
+    count = 1;
+  }
+  count_map[label] = count;
```
elliottt, line 514:
```suggestion
  size_t count = 0;
  // 2. If map[label] exists, set map[label] to map[label] + 1.
  auto it = count_map.find(label);
  if (it != count_map.end()) {
    count = it->second;
  }
  count += 1;
  count_map[label] = count;
```

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -431,22 +434,390 @@ static bool console_out(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+    if (!label_string) {
+      return false;
+    }
+    label = label_string.get();
+  } else {
+    label = "default";
+  }
+  // 2. If map[label] exists, set map[label] to 0.
+  if (count_map.contains(label)) {
+    count_map[label] = 0;
```
elliottt, line 543:
```suggestion
  auto it = count_map.find(label);
  if (it != count_map.end()) {
    it->second = 0;
```

elliottt (approved):
Looks good to me! Just a few suggestions for map query/update patterns.

runtime/js-compute-runtime/builtins/shared/console.cpp
JakeChampion: node.js does dynamic formatting, it will use ms first and then if/when the period increases it will move to seconds/minutes/etc

Firefox always uses ms
Safari always uses fractional ms

I've updated our implementation to always use fractional ms 👍

PR #523, opened 2023-05-10
feat: Add Backend.exists, Backend.fromName, Backend.isHealthy, and Backend.prototype.toName

JakeChampion:
This adds new methods to the `Backend` class which enable the application to retrieve information about any backend, this is particularly useful for checking if the backend is "healthy".

Added:
- [x] Backend.exists(name) - Check whether a backend with the given name exists for the Fastly Service.
- [x] Backend.fromName(name) - Check whether a backend with the given name exists for the Fastly Service and if it does, then returns an instance of Backend for the given name.
- [x] Backend.isHealthy(name) - Returns whether the backend with the given name is healthy.

- [x] Backend.prototype.toName() - Return the name for the Backend instance.

### DIFF
--- a/runtime/js-compute-runtime/builtins/backend.cpp
+++ b/runtime/js-compute-runtime/builtins/backend.cpp
@@ -813,7 +813,7 @@ bool Backend::to_string(JSContext *cx, unsigned argc, JS::Value *vp) {
 }
 
 namespace {
-JSString *parse_and_validate_name(JSContext *cx, JS::HandleValue name_val) {
+host_api::HostString parse_and_validate_name(JSContext *cx, JS::HandleValue name_val) {
   if (name_val.isNullOrUndefined()) {
     JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_BACKEND_NAME_NOT_SET);
     return nullptr;
@@ -831,7 +831,7 @@ JSString *parse_and_validate_name(JSContext *cx, JS::HandleValue name_val) {
     JS_ReportErrorNumberASCII(cx, GetErrorMessage, nullptr, JSMSG_BACKEND_NAME_EMPTY);
     return nullptr;
   }
-  return name;
+  return core::encode(cx, name);
 }
 } // namespace
 
@@ -842,13 +842,11 @@ bool Backend::exists(JSContext *cx, unsigned argc, JS::Value *vp) {
     return false;
   }
 
-  JS::RootedString name(cx, parse_and_validate_name(cx, args.get(0)));
+  auto name = parse_and_validate_name(cx, args.get(0));
   if (!name) {
     return false;
   }
-  auto name_chars = core::encode(cx, name);
-  std::string_view name_view = name_chars;
-  auto res = host_api::Backend::exists(name_view);
+  auto res = host_api::Backend::exists(name);
   if (auto *err = res.to_err()) {
     HANDLE_ERROR(cx, *err);
     return false;
@@ -865,12 +863,11 @@ bool Backend::from_name(JSContext *cx, unsigned argc, JS::Value *vp) {
     return false;
   }
 
-  JS::RootedString name(cx, parse_and_validate_name(cx, args.get(0)));
+  auto name = parse_and_validate_name(cx, args.get(0));
   if (!name) {
     return false;
   }
-  auto name_chars = core::encode(cx, name);
-  auto res = host_api::Backend::exists(name_chars);
+  auto res = host_api::Backend::exists(name);
   if (auto *err = res.to_err()) {
     HANDLE_ERROR(cx, *err);
     return false;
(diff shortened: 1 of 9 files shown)


runtime/js-compute-runtime/builtins/backend.h
```diff
@@ -47,7 +47,16 @@ class Backend : public BuiltinImpl<Backend> {
...
   static bool set_sni_hostname(JSContext *cx, JSObject *backend, JS::HandleValue sniHostname_val);
   static bool set_name(JSContext *cx, JSObject *backend, JS::HandleValue name_val);
+
+  // static methods
+  static bool exists(JSContext *cx, unsigned argc, JS::Value *vp);
+  static bool fromName(JSContext *cx, unsigned argc, JS::Value *vp);
+  static bool isHealthy(JSContext *cx, unsigned argc, JS::Value *vp);
+
+  // prototype methods
+  static bool toName(JSContext *cx, unsigned argc, JS::Value *vp);
```
elliottt, line 57: Could you change these to snake case?

runtime/js-compute-runtime/host_interface/component/fastly_world_adapter.cpp
```diff
@@ -1019,3 +1019,41 @@ bool fastly_compute_at_edge_cache_get_state(fastly_compute_at_edge_cache_handle_
...
+                                           fastly_compute_at_edge_types_error_t *err) {
+  uint32_t ret_int;
+  if (!convert_result(fastly::backend_exists(backend->ptr, backend->len, &ret_int), err)) {
+    return false;
+  }
+  *ret = (bool)ret_int;
+  return true;
+}
+
+fastly_compute_at_edge_backend_backend_health_t convert_fastly_backend_health(uint32_t version) {
```
elliottt, line 1036: Should this take a `BACKEND_HEALTH` value instead?

runtime/js-compute-runtime/builtins/backend.cpp
```diff
@@ -806,41 +806,153 @@ JSString *Backend::name(JSContext *cx, JSObject *self) {
...
-};
-
-const JSFunctionSpec Backend::methods[] = {JS_FN("toString", toString, 0, JSPROP_ENUMERATE),
-                                           JS_FS_END};
-
-const JSPropertySpec Backend::properties[] = {JS_PS_END};
-
-bool Backend::set_name(JSContext *cx, JSObject *backend, JS::HandleValue name_val) {
+namespace {
+JSString *parse_and_validate_name(JSContext *cx, JS::HandleValue name_val) {
```
elliottt, line 816: Do you know if returning a `JSString *` is safe here? Its value is guaranteed to not move while the `JS::RootedString` is present, but once it's returned I'm not sure there's anything that would prevent the GC from moving the value.

Perhaps and easier change would be to return a `host_api::HostString`, given that each use of this function immediately calls `core::encode` on the return value?

runtime/js-compute-runtime/builtins/backend.cpp
```diff
@@ -806,41 +806,153 @@ JSString *Backend::name(JSContext *cx, JSObject *self) {
...
+    return false;
+  }
+
+  JS::RootedString name(cx, parse_and_validate_name(cx, args.get(0)));
+  if (!name) {
+    return false;
+  }
+  auto nameChars = core::encode(cx, name);
+  std::string_view name_view = nameChars;
+  auto res = host_api::Backend::exists(name_view);
```
elliottt, line 874: `host_api::HostString` will automatically convert to a `std::string_view`, so you can skip the intermediate definition :)
```suggestion
  auto res = host_api::Backend::exists(nameChars);
```

runtime/js-compute-runtime/builtins/backend.cpp
```diff
@@ -806,41 +806,153 @@ JSString *Backend::name(JSContext *cx, JSObject *self) {
...
+  }
+
+  res = host_api::Backend::isHealthy(name_view);
+  if (auto *err = res.to_err()) {
+    HANDLE_ERROR(cx, *err);
+    return false;
+  }
+  auto healthy = res.unwrap();
+
+  args.rval().setBoolean(healthy);
```
elliottt, line 939:
```suggestion

  args.rval().setBoolean(res.unwrap());
```

runtime/js-compute-runtime/host_interface/host_api.cpp
```diff
@@ -1217,4 +1217,34 @@ Result<std::optional<HostString>> Fastly::purge_surrogate_key(std::string_view k
...
+Result<bool> Backend::isHealthy(std::string_view name) {
+  Result<bool> res;
+
+  auto name_str = string_view_to_world_string(name);
+  fastly_compute_at_edge_backend_backend_health_t ret;
+  fastly_compute_at_edge_types_error_t err;
+  if (!fastly_compute_at_edge_backend_is_healthy(&name_str, &ret, &err)) {
+    res.emplace_err(err);
+  } else {
+    res.emplace(ret == FASTLY_COMPUTE_AT_EDGE_BACKEND_BACKEND_HEALTH_HEALTHY);
```
elliottt, line 1244: This will consider the `unknown` response as unhealthy, is that intended? Viceroy unconditionally reports backends as `unknown`, so it might be nice to be able to distinguish those cases.

elliottt (commented):
Looking great!

I had one question about the interpretation of the `unknown` health status, and a few c++ wibbles :)

elliottt (approved):
This looks good to me! Do you know why the e2e test failed?

JakeChampion:
> This looks good to me! Do you know why the e2e test failed?

It fails because we don't have a healthcheck assigned to the backend, so it will always report as unknown -- i've updated the test to reflect this behaviour 👍

PR #528, opened 2023-05-11
Re-generate fastly-world with wit-bindgen 0.6

elliottt:
Re-generate the fastly-world with wit-bindgen version 0.6. The main changes introduced by this were related to the names of generated types changing to be prefixed with `fastly_world_` instead of just `fastly_`. I've also committed the object file that wit-bindgen generates, and started linking it into the component wasm.

I have also renamed the exported interface in fastly.wit from `js-compute-runtime` to `compute-at-edge`, so that it's easier to test a corresponding generic compute-at-edge host in viceroy.


elliottt:
@guybedford: I'm not sure how to handle the npm package failure, as I think that it stems from needing a newer version of the `wasi_snapshot_preview1.reactor.wasm`.

runtime/js-compute-runtime/fastly.wit
```diff
@@ -486,7 +486,7 @@ default world fastly-world {
     purge-surrogate-key: func(surrogate-key: string, soft-purge: bool) -> result<purge-result, error>
   }
 
-  export js-compute-runtime: interface {
+  export compute-at-edge: interface {
```
elliottt, line 489: This change was necessary as we need to name the exported interface in the host side as well. Happy to change to a different name, but this seemed generic enough for multiple sdks to target.

guybedford (approved):

PR #535, opened 2023-05-18
Fix typos and minor doc changes

andrebenedetti:
Hello!

Sorry to bother again with small things :zany_face: 

Spotted a few typos and also I've ran into a problem running `sh download-engine.sh` instead of `bash download-engine.sh` getting the error `Illegal option -o pipefail`, I believe due to `#!/usr/bin/env bash` resolving to `sh` on Ubuntu 22.04 and not recognizing -o pipefail as a legal option

### DIFF
--- a/js-compute-runtime-cli.js
+++ b/js-compute-runtime-cli.js
@@ -21,11 +21,11 @@ if (version) {
   await printHelp();
 } else {
   // This is a dynamic import because this import will throw an error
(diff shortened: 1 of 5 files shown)


js-compute-runtime-cli.js
andrebenedetti, line 1: I suppose a "not" was missing here

JakeChampion (approved):

JakeChampion:
Thank you for finding and fixing these!

PR #536, opened 2023-05-18
Abstract over async handles

elliottt:
Introduce an `AsyncHandle` type to the host api, and refactor the event loop to use them instead.

I reworked the interface to the underlying `fastly_async_io_select` host call, so that it's now a bit more clear when the timeout has expired. I've also added comments to `process_pending_async_tasks` to better explain the interactions between the two action queues.

### DIFF
--- a/runtime/js-compute-runtime/js-compute-builtins.cpp
+++ b/runtime/js-compute-runtime/js-compute-builtins.cpp
@@ -1105,24 +1105,17 @@ static bool init(JSContext *cx, HandleObject global) {
 
 bool has_pending_async_tasks() { return pending_async_tasks->length() > 0 || !timers->empty(); }
 
-bool process_body_read(JSContext *cx, HandleObject streamSource);
-
-bool process_pending_request(JSContext *cx, HandleObject request) {
-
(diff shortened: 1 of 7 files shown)


runtime/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1210,62 +1202,60 @@ bool process_pending_async_tasks(JSContext *cx) {
...
-  uint32_t ready_index = ret.val;
-
-  if (ready_index == UINT32_MAX) {
-    // The index will be UINT32_MAX if the timeout expires before any objects are ready for I/O.
-    return true;
-  }
+  // At this point we know that handles is not empty, and that ready_index is valid, both because
+  // the timeout wasn't reached. If handles was empty and timeout was zero, we would have errored
+  // out after the call to `select`. If the timeout was non-zero and handles was empty, the timeout
+  // would expire and we would exit through the path that runs the first timer.
```
JakeChampion, line 1242: Thank you for this great comment 🙌

JakeChampion (approved):

PR #541, opened 2023-05-26
Add option to use `weval`, an experimental JS AOT option using partial evaluation.

cfallin:
This PR pulls in some still-experimental work to partially evaluate the SpiderMonkey JS interpreter together with JS scripts, ahead-of-time, to produce compiled code.

It should not yet be considered production-worthy, pending more testing and proving out its stability and performance; but, it is at a point where we believe it should be under a *non-default* option so that users of the SDK may try it. No guarantees! It may explode your bytecode!

The support works by bundling two builds of the engine Wasm -- with and without the weval intrinsics -- and invoking either Wizer with the "normal" build or weval with the "weval" build, as appropriate. The `--enable-weval` option selects the latter.

I have measured weval showing speedups of ~30% on a Markdown renderer on my machine, and up to ~3.5x on the `crypto` benchmark in the Octane suite (1.83x speedup geomean).

This refers to a hash for the `runtime/spidermonkey` submodule that is in fastly/spidermonkey-wasi-embedding#17; that will need to be merged first (and probably bytecodealliance/gecko-dev#5 before that).

I'm not entirely sure about some of the build/packaging integration; that's the bulk of the changes here and will need special review.

The package deps include `@cfallin/weval`; if it's preferable, I should probably move the repo to BA and get the appropriate permissions to publish it as `@bytecodealliance/weval`.


src/parseInputs.js
```diff
@@ -93,5 +98,17 @@ export async function parseInputs(cliInputs) {
...
+
+  if (!wasmEngine) {
+    wasmEngine = wasmEngineBase;
+    if (component) {
+      wasmEngine += "-component";
+    }
+    if (enableWeval) {
+      wasmEngine += "-weval.wasm";
+    } else {
+      wasmEngine += "-normal.wasm";
```
guybedford, line 110: These asset names are also hard-coded in https://github.com/fastly/js-compute-runtime/blob/main/.github/workflows/main.yml#L119, which I think would need to be updated too.

runtime/js-compute-runtime/Makefile
```diff
@@ -67,6 +67,9 @@ else
   WASM_STRIP = wasm-opt --strip-debug -o $1 $1
```
guybedford, line 67: We likely do have to tackle this one head-on here I think - `wasm-opt` is in PATH in all our current workflows, so would need to be renamed. Binaryen bins are added to PATH in https://github.com/fastly/js-compute-runtime/blob/main/.github/workflows/main.yml#L93.

guybedford (commented):
So exciting, can't wait to try it out! I tried building this locally against wasi-sdk@20, and got the error:

```
 0:56.44 js/src/build/libjs_static.a
 0:56.82 js/src/shell/js
 0:56.86 wasm-ld: warning: unknown -z value: noexecstack
 0:56.86 wasm-ld: warning: unknown -z value: text
 0:56.86 wasm-ld: warning: unknown -z value: relro
 0:56.86 wasm-ld: warning: unknown -z value: nocopyreloc
 0:57.50 clang-16: error: unable to execute command: posix_spawn failed: Exec format error
 0:57.50 clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
 0:57.50 make[4]: *** [../../../dist/bin/js] Error 1
 0:57.50 make[3]: *** [js/src/shell/target] Error 2
 0:57.50 make[2]: *** [compile] Error 2
 0:57.50 make[1]: *** [default] Error 2
 0:57.50 make: *** [build] Error 2
 0:57.53 0 compiler warnings present.
```

`wasm-ld --version` gives `LLD 16.0.0`.

src/compileApplicationToWasm.js
```diff
@@ -4,11 +4,12 @@ import { mkdir, readFile } from "node:fs/promises";
 import { isFile } from "./isFile.js";
 import { isFileOrDoesNotExist } from "./isFileOrDoesNotExist.js";
 import wizer from "@bytecodealliance/wizer";
+import weval from "@cfallin/weval";
```
guybedford, line 7: A standard technique for lazy initialization in JavaScript would be to turn this into a dynamic `import()` nested into the async code path where Weval is used only.

JakeChampion:
@cfallin, as this is from a fork, our CI won't work correctly as it will not allow access to the GitHub Secrets which are required for the full test suite to run

I've branched from your fork and made patches to the GitHub Workflows/Jobs, which is available [here](https://github.com/fastly/js-compute-runtime/compare/jake/weval-edits?expand=1) -- the [latest CI run](https://github.com/fastly/js-compute-runtime/actions/runs/5091597139/jobs/9152405596) is showing that some of the web-platform-test results have changed when using the weval build.

To run the web-platform-tests locally and view the results, you can do the following from the project root:
```
./tests/wpt-harness/run-wpt.sh --update-expectations -vv
```

To make it simpler to see the difference test results, I've also committed and pushed the web-platform-test results which were updated when using the weval build -- [64ffa2f18597117302d2696138df97aac2ccd890](https://github.com/fastly/js-compute-runtime/commit/64ffa2f18597117302d2696138df97aac2ccd890)

cfallin:
@JakeChampion thanks, I'm able to reproduce this locally and I'm taking a look!

cfallin:
After a lot of debugging, I did the obvious thing and tried repo-version combinations: reverting the `gecko-dev` repo to its pre-weval state (`69d3daaa3aa4589ac8e60a94c7c7e0592882d272`) but keeping spidermonkey-wasi-embedding at `5ff4d4babce2fc02ca7fbc14ce74fa4cdc197b0f` continues to show WPT errors; so perhaps it's a build option issue; continuing to bisect.

cfallin:
@JakeChampion actually, after reverting to this repo's `main1`, I see the same failures I had been seeing on this branch locally, so I think something is broken with my testing setup. In particular I see this line

```
Done. Stats:                                                       3742 / 4967 (   +0,    -0,    ?0) passing in 86955ms
```

I might need your help to dig more deeply!

cfallin:
This has been superseded by #883 with a new interpreter (PBL) processed by weval; closing this PR!

PR #552, opened 2023-06-07
feat: implement SimpleCache.getOrSet method

JakeChampion:
Note: Half the diff is tests 👍 

SimpleCache.getOrSet is a simplified interface to Fastly's request collapsing functionality for lookups and inserts into the service's cache, the method will either return the content in the cache under the supplied key, or it will evaluate the supplied function and insert it's result into the cache, and then return the content in the cache under the supplied key. 

The SimpleCache classes definition now includes:
```ts
static getOrSet(key: string, set: () => Promise<{value: BodyInit,  ttl: number}>): Promise<SimpleCacheEntry>;
static getOrSet(key: string, set: () => Promise<{value: ReadableStream, ttl: number, length: number}>): Promise<SimpleCacheEntry>;
```

Example application:
```js
import { SimpleCache } from 'fastly:cache';

addEventListener('fetch', event => event.respondWith(app(event)));

async function app(event) {
  const path = new URL(event.request).pathname;
  let page = SimpleCache.getOrSet(path, async function () {
    return {
      value: await render(path),
      // Store the page in the cache for 1 minute.
      ttl: 60
    };
  });
  return new Response(page, {
    headers: {
      'content-type': 'text/plain;charset=UTF-8'
    }
  });
}

async function render(path) {
  // expensive/slow function which constructs and returns the contents for a given path
  await new Promise(resolve => setTimeout(resolve, 10_000));
  return path;
}
```


runtime/js-compute-runtime/builtins/cache-simple.cpp
```diff
@@ -156,6 +156,446 @@ JS::Result<std::string> createSurrogateKeyFromCacheKey(JSContext *cx,
...
+
+  if (!JS_GetProperty(cx, extraObj, "handle", &handleVal)) {
+    return RejectPromiseWithPendingError(cx, promise);
+  }
+  MOZ_ASSERT(handleVal.isInt32());
+  fastly_cache_handle_t handle = handleVal.toInt32();
+  fastly_error_t err;
+  JS::RootedValue keyVal(cx);
+  if (!JS_GetProperty(cx, extraObj, "key", &keyVal)) {
+    if (!fastly_transaction_cancel(handle, &err)) {
```
elliottt, line 182: We could come up with an RAII transaction class that would automatically cancel the transaction, unless explicitly released. No need to do that in this PR, but it would be a nice cleanup :)

runtime/js-compute-runtime/builtins/cache-simple.cpp
```diff
@@ -156,6 +156,446 @@ JS::Result<std::string> createSurrogateKeyFromCacheKey(JSContext *cx,
...
+    HANDLE_ERROR(cx, err);
+    return false;
+  }
+
+  // Check if a fresh cache item was found, if that's the case, then we will resolve
+  // with a SimpleCacheEntry containing the value. Else, call the content-provided
+  // function in the `set` parameter and insert it's returned value property into the
+  // cache under the provided `key`, and then we will resolve with a SimpleCacheEntry
+  // containing the value.
+  alignas(4) fastly_cache_lookup_state_t state;
```
elliottt, line 480: Is the alignment annotation here necessary?

elliottt (approved):
This looks great! Let's pair soon on adding a trasaction managing type to clean up the error handling a bit :)

PR #556, opened 2023-06-09
fix: only apply our pipeTo/pipeThrough optimisations to TransformStreams who have no transformers (IdentityStreams).

JakeChampion:
We have a performance optimisation for when the receiver is a native source, whereby we append the entire source body to the destination using a single native hostcall, this greatly improves the performance (https://barstool.engineering/a-real-world-comparison-between-cloudflare-workers-and-fastly-compute-edge/)

This optimisation however, should only be applied when the TransformStream is being used in this particular, having not supplied a transformer to the TransformStream constuctor, i.e. an IdentityStream. If we apply the optimisation when a transformer is supplied, the transformer would never get called, which is not correct behavior.

We want to apply the optimisation in this scenario:
```ts
function combineStreams(streams: ReadableStream[]): ReadableStream {
  const stream = new TransformStream()
  _combineStreams(streams, stream.writable)
  return stream.readable
}

async function _combineStreams(sources: ReadableStream[], destination: WritableStream) {
  for (const stream of sources) {
    await stream.pipeTo(destination, {
      preventClose: true
    })
  }
  destination.close()
}
```

And we don't want to apply the optimisation in any other scenario such as:
```js
function upperCaseStream() {
  return new TransformStream({
    transform(chunk, controller) {
      controller.enqueue(chunk.toUpperCase());
    },
  });
}
```

### DIFF
--- a/runtime/js-compute-runtime/builtins/transform-stream.cpp
+++ b/runtime/js-compute-runtime/builtins/transform-stream.cpp
@@ -42,8 +42,7 @@ bool pipeTo(JSContext *cx, unsigned argc, JS::Value *vp) {
   JS::RootedObject target(cx, args[0].isObject() ? &args[0].toObject() : nullptr);
   if (target && builtins::NativeStreamSource::stream_has_native_source(cx, self) &&
       JS::IsWritableStream(target) && builtins::TransformStream::is_ts_writable(cx, target)) {
-    auto ts = builtins::TransformStream::ts_from_writable(cx, target);
-    if (ts) {
+    if (auto ts = builtins::TransformStream::ts_from_writable(cx, target)) {
       auto streamHasTransformer =
           JS::GetReservedSlot(ts, builtins::TransformStream::Slots::HasTransformer).toBoolean();
       // We only want to apply the optimisation on identity-streams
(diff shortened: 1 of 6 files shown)


triblondon (approved):
👍  for the test case

runtime/js-compute-runtime/builtins/transform-stream.cpp
```diff
@@ -42,7 +42,15 @@ bool pipeTo(JSContext *cx, unsigned argc, JS::Value *vp) {
   JS::RootedObject target(cx, args[0].isObject() ? &args[0].toObject() : nullptr);
   if (target && builtins::NativeStreamSource::stream_has_native_source(cx, self) &&
       JS::IsWritableStream(target) && builtins::TransformStream::is_ts_writable(cx, target)) {
-    builtins::NativeStreamSource::set_stream_piped_to_ts_writable(cx, self, target);
+    auto ts = builtins::TransformStream::ts_from_writable(cx, target);
+    if (ts) {
```
JakeChampion, line 46:
```suggestion
    if (auto ts = builtins::TransformStream::ts_from_writable(cx, target)) {
```

elliottt (approved):
Looks great!

PR #558, opened 2023-06-09
feat: implement web performance api

JakeChampion:
This is an implementation of both of the web performance APIs which are in the WinterCG minimum common runtime API list

https://w3c.github.io/hr-time/#dom-performance-now

And 

https://w3c.github.io/hr-time/#timeorigin-attribute


runtime/js-compute-runtime/builtins/shared/performance.cpp
```diff
@@ -0,0 +1,58 @@
...
+  auto finish = std::chrono::high_resolution_clock::now();
+  auto duration = FpMilliseconds(finish - builtins::Performance::timeOrigin.value()).count();
+
+  JS::RootedValue elapsed(cx, JS::Float32Value(duration));
+  args.rval().set(elapsed);
+  return true;
+}
+
+bool Performance::timeOrigin_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  MOZ_ASSERT(builtins::Performance::timeOrigin.has_value());
```
guybedford, line 28: If I'm reading this right, any `performance.timeOrigin` get during Wizering time will fail? That seems like a valid behaviour to me as this represents a time before the origin time, but we should at least give it a nice proper error.

Another alternative might be to maintain `timeOrigin` as the Wizer initialization time and then ensure that `performance.now()` gives a time that represents the time since the Wizering time origin. That could be done by "rebasing" the `timeOrigin`, basically to measure the difference on Wizer reinitialization and to add that difference to all `performance.now()` measurements to create a continuous cross-execution time. This kind of approach might even be useful in the context of possible multi-pause-resume states in future. But it does mean making the `timeOrigin` somewhat useless by design, so perhaps worth reconsidering at a later date.

guybedford (commented):
👍

PR #566, opened 2023-06-22
feat: implement and expose JavaScript classes for Fastly's Compute Core Cache feature set

JakeChampion:
5 JavaScript Classes have been added, which together allow Fastly Compute Services written in JavaScript to have the ability to use Fastly's Compute Core Cache.

Below is an example of what a JavaScript application using these features might look like:

```js
import { CoreCache } from "fastly:cache";

/**
 * @param {FetchEvent} event
 * @returns {Promise<Response>}
*/
async function app(event) {
    const path = (new URL(event.request.url)).pathname
    let entry = CoreCache.transactionLookup(path)
    if (entry.state().mustInsertOrUpdate()) {
        let [writer, reader] = entry.insertAndStreamBack({
            maxAge: 60 * 1000
        })
        writer.append(`hello from ${path}`);
        writer.close();
        return new Response(reader.body(), {
            headers: {
                "x-cache": "MISS",
                "x-cache-hits": 0
            }
        });
    } else {
        return new Response(entry.body(), {
            headers: {
                "x-cache": "HIT",
                "x-cache-hits": entry.hits()
            }
        });
    }
}

    event.respondWith(app(event))
})
```

With the above application, this is what some requests and responses look like:

```shell
❯ xh 'https://frequently-strong-boxer.edgecompute.app/dog'
HTTP/2.0 200 OK
date: Thu, 22 Feb 2024 16:15:22 GMT
x-cache: MISS
x-cache-hits: 0
x-served-by: cache-lhr7356-LHR

hello from /dog
❯ xh 'https://frequently-strong-boxer.edgecompute.app/dog'
HTTP/2.0 200 OK
date: Thu, 22 Feb 2024 16:15:23 GMT
x-cache: HIT
x-cache-hits: 1
x-served-by: cache-lhr7385-LHR

hello from /dog
❯ xh 'https://frequently-strong-boxer.edgecompute.app/bird'
HTTP/2.0 200 OK
date: Thu, 22 Feb 2024 16:15:26 GMT
x-cache: MISS
x-cache-hits: 0
x-served-by: cache-lhr7379-LHR

hello from /bird
❯ xh 'https://frequently-strong-boxer.edgecompute.app/bird'
HTTP/2.0 200 OK
date: Thu, 22 Feb 2024 16:15:27 GMT
x-cache: HIT
x-cache-hits: 1
x-served-by: cache-lhr7325-LHR

hello from /bird
```


runtime/js-compute-runtime/Makefile
```diff
@@ -404,7 +404,7 @@ compile_commands.json:
 			sep=","; \
 			echo "{ \"directory\": \"$(FSM_SRC)\","; \
 			echo "  \"command\": \"$(WASI_CXX) $(CXX_FLAGS) $(INCLUDES) $(DEFINES)\","; \
-			echo -n "  \"file\": \"$${file}\"}"; \
+			echo "  \"file\": \"$${file}\"}"; \
```
JakeChampion, line 407:
```suggestion
			echo -n "  \"file\": \"$${file}\"}"; \
```

runtime/js-compute-runtime/builtins/cache-simple.cpp
```diff
@@ -617,27 +563,20 @@ bool SimpleCache::set(JSContext *cx, unsigned argc, JS::Value *vp) {
...
-    double number;
-    if (!JS::ToNumber(cx, length_val, &number)) {
-      return false;
-    }
-    if (number < 0 || std::isnan(number) || std::isinf(number)) {
-      JS_ReportErrorASCII(
-          cx, "SimpleCache.set: length parameter is an invalid value, only positive numbers can "
-              "be used for length values.");
-      return false;
+    if (args.hasDefined(3)) {
```
JakeChampion, line 566: The length parameter is not actually required, what happens if the length is supplied is that any clients that begin streaming the item’s contents before it is completely provided will see the promised length which allows them to, for example, use content-length instead of transfer-encoding: chunked if the item is used as the body of a Request or Response

runtime/js-compute-runtime/builtins/cache-simple.cpp
```diff
@@ -346,29 +296,25 @@ bool SimpleCache::getOrSetThenHandler(JSContext *cx, JS::HandleObject owner, JS:
     if (!JS_HasProperty(cx, insertionObject, "length", &found)) {
       return false;
     }
-    if (!found) {
-      JS_ReportErrorASCII(cx, "SimpleCache.getOrSet: length property is required when the value "
```
JakeChampion, line 350: The length parameter is not actually required, what happens if the length is supplied is that any clients that begin streaming the item’s contents before it is completely provided will see the promised length which allows them to, for example, use content-length instead of transfer-encoding: chunked if the item is used as the body of a Request or Response

runtime/js-compute-runtime/host_interface/wit/deps/fastly/compute-at-edge.wit
```diff
@@ -824,18 +830,25 @@ interface cache {
...
+  /// choosing a new waiter to perform the insertion/update.
+  close: func(handle: handle) -> result<_, error>
+
   get-state: func(handle: handle) -> result<lookup-state, error>
 
-  /// Gets the user metadata of the found object, returning the `optional-none` error if there
+  /// Gets the user metadata of the found object, returning `none` if there
   /// was no found object.
-  get-user-metadata: func(handle: handle) -> result<string, error>
+  get-user-metadata: func(handle: handle) -> result<list<u8>, error>
```
JakeChampion, line 851: I changed this from string as the supplied metadata can be any arbitrary bytes

types/index.d.ts
```diff
@@ -8,6 +10,4 @@
 /// <reference path="geolocation.d.ts" />
 /// <reference path="logger.d.ts" />
 /// <reference path="kv-store.d.ts" />
-/// <reference path="cache.d.ts" />
 /// <reference path="secret-store.d.ts" />
```
JakeChampion, line 13:
```suggestion
/// <reference path="secret-store.d.ts" />
/// <reference path="globals.d.ts" />
```

runtime/js-compute-runtime/builtins/cache-core.cpp
```diff
@@ -0,0 +1,1162 @@
...
+      JS::RootedValue headers_val(cx, JS::ObjectValue(*headers));
+      JS::RootedObject requestInstance(cx, Request::create_instance(cx));
+      if (!requestInstance) {
+        return JS::Result<host_api::CacheLookupOptions>(JS::Error());
+      }
+
+      JS::RootedValue input(cx, JS::StringValue(JS_NewStringCopyZ(cx, "http://example.com")));
+      JS::RootedObject request(cx,
+                               builtins::Request::create(cx, requestInstance, input, headers_val));
+      options.request_headers = host_api::HttpReq(builtins::Request::request_handle(request));
```
elliottt, line 54: Could you add a comment about why `http://example.com` is used here?

runtime/js-compute-runtime/builtins/cache-core.cpp
```diff
@@ -0,0 +1,1162 @@
...
+  }
+  JS::RootedObject options_obj(cx, &options_val.toObject());
+
+  JS::RootedValue maxAge_val(cx);
+  if (!JS_GetProperty(cx, options_obj, "maxAge", &maxAge_val)) {
+    return JS::Result<host_api::CacheWriteOptions>(JS::Error());
+  }
+  // maxAge property is required
+  if (maxAge_val.isUndefined()) {
+    JS_ReportErrorASCII(cx, "maxAge is required");
```
elliottt, line 85: What do you think about merging these two? We'd overwrite the error about field lookup with the `maxAge is required` message, but I think that might be more useful?

runtime/js-compute-runtime/builtins/cache-core.cpp
```diff
@@ -0,0 +1,1162 @@
...
+    if (!headers) {
+      return JS::Result<host_api::CacheWriteOptions>(JS::Error());
+    }
+    JS::RootedValue headers_val(cx, JS::ObjectValue(*headers));
+    JS::RootedObject requestInstance(cx, Request::create_instance(cx));
+    if (!requestInstance) {
+      return JS::Result<host_api::CacheWriteOptions>(JS::Error());
+    }
+
+    JS::RootedValue input(cx, JS::StringValue(JS_NewStringCopyZ(cx, "http://example.com")));
```
elliottt, line 336: Could you add a comment about why this constant is used?

elliottt (commented):
This seems pretty reasonable to me! In the future, do you think there's a way that we could break changes like this into multiple parts? There's a lot of change in this one :)

Do you know what's causing the WPT suite to fail?

runtime/js-compute-runtime/host_interface/host_api.h
```diff
@@ -641,6 +650,27 @@ class SecretStore final {
   Result<std::optional<Secret>> get(std::string_view name);
 };
 
+class Body final {
+public:
+  using Handle = uint32_t;
```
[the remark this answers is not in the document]
JakeChampion: It's completely unused 🙈

runtime/js-compute-runtime/builtins/cache-core.cpp
JakeChampion: That's a really nice improvement for the end-user 🙌

runtime/js-compute-runtime/builtins/cache-core.cpp
JakeChampion: Good point, I've now added one in [`38a8ce9` (#566)](https://github.com/fastly/js-compute-runtime/pull/566/commits/38a8ce9612e3af8081304979ee6983d5cfa679c8)

runtime/js-compute-runtime/builtins/cache-core.cpp
JakeChampion: Good point, I've now added one in [`38a8ce9` (#566)](https://github.com/fastly/js-compute-runtime/pull/566/commits/38a8ce9612e3af8081304979ee6983d5cfa679c8)

JakeChampion:
> This seems pretty reasonable to me! In the future, do you think there's a way that we could break changes like this into multiple parts? There's a lot of change in this one :)

Apologies, yes, I reckon there is a way to break it up to make it simpler to review and land in incremental pieces (Or perhaps some sort of stacked diff/pr system so we still land it as one chunk of work into main?)

> Do you know what's causing the WPT suite to fail?

Yes, that was the refactoring of the parseBodyInit function, it turns that all the copies of that function were not in fact completely identical, the implementation for Request/Response actually sets a content-type header under certain conditions. I've reverted that change and now the WPT are happy once again

elliottt (approved):
Awesome! :tada:

PR #574, opened 2023-07-06
feat: Add ability to disable connection-pooling behavior for Dynamic Backends

JakeChampion:
This commit adds a new `dontPool` optional property to the `Backend` constructor, which indicates whether or not to disable connection-pooling.

Connection pooling is the ability to re-use connections across different executions/sessions of a Fastly Compute@Edge Service Fastly considers two backends "the same" if they’re registered with the same name and the exact same settings. In those cases, when connection pooling is enabled, if an execution/session opens a connection to a backend, the connection will be left open, and can be re-used by another execution/session. This pooling behaviour can help improve backend latency, by removing the need for the initial network / TLS handshake(s). By default, pooling is enabled for Dynamic Backends, but can be disabled when using the Explicit Dynamic Backend API.


elliottt (commented):
This is looking great! Do you need to update `register_dynamic_backend` in host_api.cpp as well?

JakeChampion:
> This is looking great! Do you need to update `register_dynamic_backend` in host_api.cpp as well?

You are right, yes -- that has now been updated 👍

elliottt (approved):
Looks great!

PR #576, opened 2023-07-07
feat!: Rename SimpleCache.delete to SimpleCache.purge and require purge options to be supplied as the second parameter

JakeChampion:
We are renaming because "purge" is already a well-known and documented concept for removing content from Fastly's cache.

The new addition of a second argument allows the caller to decide what scope to purge the content from, currently they can choose to purge from all of Fastly ("global") or from the POP that contains the currently executing instance ("pop"). We do not provide a default option right now, in the future we may provide a default option, if we discover a common pattern is being used.

Here is an example of migrating an application using SimpleCache.delete to SimpleCache.purge with the same behaviour:
```diff
/// <reference types="@fastly/js-compute" />

import { SimpleCache } from 'fastly:cache';

addEventListener('fetch', event => event.respondWith(app(event)));

async function app(event) {
  const url = new URL(event.request);
  const path = url.pathname;
  if (url.searchParams.has('delete')) {
-    SimpleCache.delete(path);
+    SimpleCache.purge(path, { scope: "global" });
    return new Response(page, { status: 204 });
  }

  let page = SimpleCache.getOrSet(path, async () => {
    return {
      value: await render(path),
      // Store the page in the cache for 1 minute.
      ttl: 60
    }
  });
  return new Response(page, {
    headers: {
      'content-type': 'text/plain;charset=UTF-8'
    }
  });
}

async function render(path) {
  // expensive/slow function which constructs and returns the contents for a given path
  await new Promise(resolve => setTimeout(resolve, 10_000));
  return path;
}


```


documentation/docs/fastly:cache/SimpleCache/purge.mdx
```diff
@@ -0,0 +1,89 @@
...
+```js
+purge(key, options)
+```
+
+### Parameters
+
+- `key` _: string_
+  - The key to purge from within the cache.
+
+- `options` _: object_
```
guybedford, line 23: If scope is the main required configuration, would it be worth considering a convenience form like `purge(key, 'global')` or `purge(key, 'local')`?

types/fastly:cache.d.ts
```diff
@@ -1,11 +1,16 @@
 declare module "fastly:cache" {
+
+  interface PurgeOptions {
+    scope: "pop" | "global"
```
guybedford, line 4: It would be great to start documenting types. Ideally we'd generate the mdx from the types even.

documentation/docs/fastly:cache/SimpleCache/purge.mdx
```diff
@@ -0,0 +1,89 @@
...
+      - "global" - This will remove the content from all of Fastly.
+      - "pop" - This will remove the content from the POP that contains the currently executing instance.
+  - `headers`
+    - : Any headers you want to add to your request, contained
+      within a `Headers` object or an object literal with `String` values.
+  - `body`
+    - : Any body that you want to add to your request: this can be an `ArrayBuffer`, a `TypedArray`, a `DataView`, a `URLSearchParams`, string object or literal, or a `ReadableStream` object.
+  - `backend` _**Fastly-specific**_
+  - `cacheOverride` _**Fastly-specific**_
+  - `cacheKey` _**Fastly-specific**_
```
guybedford, line 36: I don't see these implemented or tested, is it worth continuing to document them?

guybedford (approved):

types/fastly:cache.d.ts
JakeChampion: We originally did generate docs from types but it didn't give us as much expressive documentation capabilities as were we wanting to have

guybedford: Nothing can beat the user experience of having docs in the types though as well.

PR #585, opened 2023-07-12
Name the fastly interface, and import it by name

elliottt:
Rework fastly.wit to no longer use an inline interface for the fastly host
interface, and import it by name. This change necessitated re-generating the
fastly-world, and updating names throughout the runtime.


guybedford (approved):
Seems good to me. Note that if this was due to binding name conflicts, there are some wit-bindgen PRs currently with some fixes. But the explicit interface seems a better pattern definitely.

PR #591, opened 2023-07-14
chore: Move encoding operations to core/encode.h, and make them return HostString

elliottt:
Instead of returning length through an out parameter from the `encode` functions, instead make a `HostString` which has both a `UniqueChars` and the length of the string.

### DIFF
--- a/runtime/js-compute-runtime/builtins/shared/console.cpp
+++ b/runtime/js-compute-runtime/builtins/shared/console.cpp
@@ -1,13 +1,15 @@
 #include "console.h"
+#include "core/encode.h"
+#include <chrono>
+#include <map>
+
 #include <js/Array.h>
-#include <js/PropertyAndElement.h>
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Winvalid-offsetof"
 #pragma clang diagnostic ignored "-Wdeprecated-enum-enum-conversion"
+#include <js/PropertyAndElement.h>
 #include <js/experimental/TypedData.h>
 #pragma clang diagnostic pop
-#include <chrono>
-#include <map>
 
 namespace {
 using FpMilliseconds = std::chrono::duration<float, std::chrono::milliseconds::period>;
@@ -217,13 +219,12 @@ JS::Result<mozilla::Ok> ObjectToSource(JSContext *cx, std::string &sourceOut, JS
       JS::RootedValue v(cx, SymbolValue(id.toSymbol()));
       MOZ_TRY(ToSource(cx, sourceOut, v, visitedObjects));
     } else {
-      size_t message_len;
       JS::RootedValue v(cx, js::IdToValue(id));
-      auto msg = encode(cx, v, &message_len);
+      auto msg = fastly::core::encode(cx, v);
       if (!msg) {
         return JS::Result<mozilla::Ok>(JS::Error());
       }
-      sourceOut += std::string(msg.get(), message_len);
+      sourceOut += std::string(msg.begin(), msg.len);
     }
 
     sourceOut += ": ";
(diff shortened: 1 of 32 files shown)


runtime/js-compute-runtime/builtins/shared/text-decoder.cpp
```diff
@@ -198,18 +199,17 @@ bool TextDecoder::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
...
   // 2. If label is an ASCII case-insensitive match for any of the labels listed in the table
   // below, then return the corresponding encoding; otherwise return failure. JS-Compute-Runtime:
-  size_t length;
   jsencoding::Encoding *encoding;
   if (label_value.isUndefined()) {
     encoding = const_cast<jsencoding::Encoding *>(jsencoding::encoding_for_label_no_replacement(
         reinterpret_cast<uint8_t *>(const_cast<char *>("UTF-8")), 5));
   } else {
-    auto label_chars = encode(cx, label_value, &length);
+    auto label_chars = fastly::core::encode(cx, label_value);
```
guybedford, line 207: Will this `fastly::core` automatically be pulled into the shared builtin libraries? If so can we work towards making this a vendor neutral approach for ComponentizeJS?

guybedford (commented):
Is there a reason not to use C++ standards here? As much as possible, it would be nice to think in terms of the bindgen moving towards a model as if it were on top of an idiomatic C++ wit bindgen interface (and ideally including the work to lower into that over time...)

runtime/js-compute-runtime/builtins/shared/text-decoder.cpp
elliottt: We could definitely do that. I'm not attached to using the top-level `fastly` namespace, but I would like to start using namespaces a bit more aggressively in the codebase to avoid name clashes that come up periodically.

guybedford: If we can continue to treat the builtins folder as an abstraction that can move towards being runtime netural I think that would be a huge help to the ComponentizeJS work. If we need to figure out a name for such namespacing, perhaps we need to do some bikeshedding on that? `smwcg_builtins::` or something?

guybedford:
That makes sense, thanks for clarifying. Separately there's the UTF16 story to put some thought to here as well.

runtime/js-compute-runtime/builtins/shared/text-decoder.cpp
elliottt: That works for me. We could also move `HostString` somewhere else, as it's just a thin wrapper over spidermonkey types.

elliottt: Do you mind if we leave this to a follow-up PR? My goal with this PR was  to continue breaking up some of the larger headers and reduce dependence on apis that can change when we re-generate the fastly-world c bindings.

guybedford: Certainly, that would be fine.

runtime/js-compute-runtime/builtins/shared/console.cpp
```diff
@@ -495,12 +491,11 @@ static bool count(JSContext *cx, unsigned argc, JS::Value *vp) {
...
   if (args.hasDefined(0)) {
     auto label_val = args.get(0);
-    size_t length;
-    auto label_string = encode(cx, label_val, &length);
+    auto label_string = fastly::core::encode(cx, label_val);
     if (!label_string) {
       return false;
     }
-    label = label_string.get();
+    label = std::string{label_string.begin(), label_string.len};
```
guybedford, line 498: Would this be a copy?

guybedford (approved):
As long as you are ok with me making a vendor-neutral renaming PR when I do the ComponentizeJS builtins upgrade, this seems fine to me.

runtime/js-compute-runtime/builtins/shared/console.cpp
elliottt: Yes, every place we use a `std::string` it'll be a copy. I think we can refactor this to avoid the copies, but that seemed like a good change for a separate PR.

elliottt: Also this isn't introducing a new copy, just being explicit about the construction of the intermediate string -- the assignment of a char pointer to a string will copy the contents of the char pointer into the string destination.

PR #593, opened 2023-07-17
chore: Introduce the core namespace

elliottt:
Rename `fastly::core` to just `core`, and move other symbols from the core directory into the `core` namespace.


guybedford (approved):
Thanks for doing this one.

PR #594, opened 2023-07-17
chore: Split the event loop out of js-compute-builtins.cpp

elliottt:
Move the event loop to its own file, and fix a memory leak with timer management. (We weren't deleting `Timer`s that had been allocated with `new`, so I switched to storing `std::unique_ptr<Timer>` values to ensure that they get free'd.)


guybedford (approved):
Would be nice if we could separate the exact HTTP APis from the event loop code a bit better, but this seems a good separation to make overall.

PR #595, opened 2023-07-17
chore: Add the host_api namespace to all host_api declarations

elliottt:
Avoid any further namespace collisions with `host_api` declarations.


elliottt:
> How are you defining `host_api` here? It seems to include both Fastly and standards and low-level APIs? Could we try to come up with a clear definition?

I'm just naming it after the header that defines them for now. We can always split it up further as we go.

PR #598, opened 2023-07-20
feat: use spidermonkey version 115

JakeChampion:
Dependent upon https://github.com/fastly/spidermonkey-wasi-embedding/pull/19 landing


elliottt (approved):
:tada:

PR #601, opened 2023-07-21
fix: reduce memory usage by caching client getters when they are first called

JakeChampion:
Currently we will create new buffers/strings whenever any of the client getters are invoked, considering these never change for the duration of the application's lifetime, let's cache them so that we don't use up more memory than necessary. Currently you can use up all available memory just by reading `event.client.tlsClientCertificate` a couple times.

### DIFF
--- a/integration-tests/js-compute/fixtures/client/bin/index.js
+++ b/integration-tests/js-compute/fixtures/client/bin/index.js
@@ -13,9 +13,6 @@ routes.set("/client/tlsJA3MD5", event => {
     return pass('ok')
 });
 routes.set("/client/tlsClientHello", event => {
-    console.log('hello');
-    console.log('event.client.tlsJA3MD5', event.client.tlsJA3MD5);
-    console.log('event.client.tlsClientHello', event.client.tlsClientHello);
     error = assert(event.client.tlsClientHello instanceof ArrayBuffer, true, 'event.client.tlsClientHello instanceof ArrayBuffer')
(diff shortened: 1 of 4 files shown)


integration-tests/js-compute/fixtures/client/bin/index.js
```diff
@@ -13,9 +13,6 @@ routes.set("/client/tlsJA3MD5", event => {
     return pass('ok')
 });
 routes.set("/client/tlsClientHello", event => {
-    console.log('hello');
-    console.log('event.client.tlsJA3MD5', event.client.tlsJA3MD5);
-    console.log('event.client.tlsClientHello', event.client.tlsClientHello);
```
JakeChampion, line 18: debugging logic that got left behind in a previous commit

guybedford (approved):
> Currently you can use up all available memory just by reading event.client.tlsClientCertificate a couple times.

Does this PR definitely resolve this? Is that not a memory leak as opposed to caching issue? Or is the cause the GC running too infrequently to be able to resolve it?

JakeChampion:
> > Currently you can use up all available memory just by reading event.client.tlsClientCertificate a couple times.
> 
> Does this PR definitely resolve this? Is that not a memory leak as opposed to caching issue? Or is the cause the GC running too infrequently to be able to resolve it?

Yes, this resolves the issue 👍
I do think we can also optimise our GC some more as well

PR #603, opened 2023-07-21
update wpt submodule and url crate

JakeChampion:
The wpt update brings in more tests
The url crate update passes some more of the url tests 😄

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -11,5 +11,5 @@ jobs:
     steps:
     - uses: actions/dependency-review-action@v2.2.0
       with:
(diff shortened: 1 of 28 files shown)


elliottt (approved):
There's a lot of new failures in addition to the new passing tests, should we be spend some time investigating to see if there are any quick fixes?

JakeChampion:
> There's a lot of new failures in addition to the new passing tests, should we be spend some time investigating to see if there are any quick fixes?

I think once the rust-url project is in a [better place for contribution](https://github.com/servo/rust-url/issues/850) we should definitely help out - perhaps we can help with the current effort to make contributions easier

PR #609, opened 2023-07-27
Split the main fastly interface of fastly.wit

elliottt:
Split `fastly.wit` into different interfaces that follow the existing module separation of the core wasm host imports.


runtime/js-compute-runtime/fastly.wit
```diff
@@ -553,17 +334,293 @@ interface fastly {
...
+/*
+ * Fastly HTTP Req
+ */
+interface http-req {
+
+  use types.{
+    request-handle, http-cache-override-tag, http-version, body-handle, error,
+    response, pending-request-handle, content-encodings, framing-headers-mode,
+    dynamic-backend-config
+  }
```
guybedford, line 422: When it comes to type sharing, the WASI approach is to have a subsystem specific interface - eg:

```
interface http {
  type request-handle = u32 // etc
}

interface http-req {
  use http.{request-handle}
}
```

Or even call it `http-types` directly.

guybedford (commented):
Very nice work, just a couple of notes.

runtime/js-compute-runtime/fastly.wit
```diff
@@ -553,17 +334,293 @@ interface fastly {
...
     must-insert-or-update,
   }
+}
+
+/*
+ * Fastly Log
+ */
+interface log {
+
+  use types.{error, log-endpoint-handle}
```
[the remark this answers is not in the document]
elliottt: Happy to move types only used in a single interface into those interfaces. That's how I had it originally, but decided somewhat arbitrarily to move everything into the types interface.

runtime/js-compute-runtime/fastly.wit
elliottt: That works for me!

guybedford (approved):
👍

PR #619, opened 2023-08-06
enable pbl

JakeChampion:
depends upon https://github.com/fastly/spidermonkey-wasi-embedding/pull/20


runtime/js-compute-runtime/js-compute-runtime.cpp
```diff
@@ -122,6 +123,8 @@ bool init_js() {
     return false;
   }
 
+  JS_SetGlobalJitCompilerOption(cx, JSJitCompilerOption::JSJITCOMPILER_PORTABLE_BASELINE_ENABLE, 1);
```
cfallin, line 126: We should probably set an "eager" option too, or at least experiment with it -- all my perf results were with eager tiering.  It looks like I haven't added that as a `JSJitCompilerOption` -- I'll add it tomorrow (PDT)!

JakeChampion:
this landed in #628

PR #622, opened 2023-08-09
chore: Add client-cert and client-key to dynamic-backend-config

elliottt:
Add the `client-cert` and `client-key` fields to the dynamic backend config record. We don't currently expose these fields through the public API, so this change should be a no-op.


runtime/js-compute-runtime/host_interface/wit/deps/fastly/compute-at-edge.wit
```diff
@@ -101,6 +105,8 @@ interface http-types {
     ca-cert: option<string>,
     ciphers: option<string>,
     sni-hostname: option<string>,
+    client-cert: option<string>,
+    client-key: secret-handle,
```
elliottt, line 109: I wonder if it would be nicer to expose this as an optional record that includes both the cert and key? The viceroy implementation will only look at the `client-key` if the `client-cert` option mask is set.

guybedford: That sounds like it might better capture the interface to me as well. Other fields here could possibly follow a similar pattern even.

guybedford (approved):

runtime/js-compute-runtime/host_interface/wit/deps/fastly/compute-at-edge.wit
elliottt: I'll make that change before merging then, thanks!

guybedford (approved):

Issue #624, opened 2023-08-16
ClientInfo type doesn't contain the new fields from v3


noahcalvo:
From the v3.0.0 changelog, 5 new fields were added to `event.client`. https://github.com/fastly/js-compute-runtime/blob/3cd6844faa71194b3cac7d7dfcc366c3b2524242/CHANGELOG.md?plain=1#L92-L96

`event.client` is of type `ClientInfo` and the type definition wasn't updated.
https://github.com/fastly/js-compute-runtime/blob/3cd6844faa71194b3cac7d7dfcc366c3b2524242/types/globals.d.ts#L260-L266

This is causing a typescript error when trying to reference `event.client.tlsCipherOpensslName` (or any of the other new fields)

JakeChampion:
Thank you for reporting this @noahcalvo - I have opened a pull-request to document these in both the typescript definitions and the reference documentation website -- https://github.com/fastly/js-compute-runtime/pull/625

PR #625, opened 2023-08-16
fix: Add documentation and type definitions for the new event.client.* fields

JakeChampion:
When I added these fields in v3.0.0 I forgot to also include typescript definitions and documentation for them
Resolves https://github.com/fastly/js-compute-runtime/issues/624


elliottt (approved):
:shipit:

PR #632, opened 2023-08-23
use pbl in all the test apps

JakeChampion:
This turns on pbl for all our test apps, which means we run pbl in debug mode for tests via viceroy and pbl in release mode for test via c@e

This also pulls in a newer version of gecko-dev which includes fixes for some debug-asserts in PBL (thank you @cfallin for PBL and all the fixes for debug mode 🙌 )


elliottt (approved):
Looks good to me! As we discussed, it might be nice to start merging the tests together so that we only need to run a single app to test all functionality. That way we could build it in multiple configurations without incurring the cross-product of c@e deployments :D

PR #633, opened 2023-08-24
Replace spidermonkey-wasi-embedding submodule with gecko-dev

JakeChampion:
This change makes it so that contributions which require modifying our branch of gecko-dev, don't require having to edit 3 repositories, and instead just 2 (gecko-dev and this repo).

We do currently lose prebuilt spidermonkey-wasi builds, which does mean a contributor will need to clone and compile gecko-dev. I've added a script which will install the required dependencies for gecko-dev, named `install-gecko-dependencies.sh`, and also updated our Makefile to compile gecko-dev.

`make` will compile gecko-dev and js-compute-runtime, both in release mode
`DEBUG=true make` will compile gecko-dev and js-compute-runtime, both in debug mode

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,8 @@
+[submodule "runtime/js-compute-runtime/gecko-dev"]
+	path = runtime/js-compute-runtime/gecko-dev
+	url = https://github.com/bytecodealliance/gecko-dev.git
+	shallow = true
+[submodule "tests/wpt-harness/wpt"]
+	path = tests/wpt-harness/wpt
+	url = https://github.com/web-platform-tests/wpt.git
+	shallow = true
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,8 +1,0 @@
-[submodule "runtime/spidermonkey"]
-	path = runtime/spidermonkey
-	url = https://github.com/fastly/spidermonkey-wasi-embedding.git
-	shallow = true
-[submodule "tests/wpt-harness/wpt"]
-	path = tests/wpt-harness/wpt
-	url = https://github.com/web-platform-tests/wpt.git
-	shallow = true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,27 +61,6 @@ jobs:
       with:
         path: runtime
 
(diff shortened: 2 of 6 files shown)


.gitmodules
```diff
@@ -1,8 +1,6 @@
-[submodule "runtime/spidermonkey"]
-	path = runtime/spidermonkey
+[submodule "runtime/js-compute-runtime/gecko-dev"]
+	path = runtime/js-compute-runtime/gecko-dev
+	url = https://github.com/bytecodealliance/gecko-dev.git
```
JakeChampion, line 3:
```suggestion
	url = https://github.com/bytecodealliance/gecko-dev.git
	shallow = true
```

.gitmodules
```diff
@@ -1,8 +1,6 @@
...
+[submodule "runtime/sm"]
+	path = runtime/sm
 	url = https://github.com/fastly/spidermonkey-wasi-embedding.git
```
JakeChampion, line 6:
```suggestion
	url = https://github.com/fastly/spidermonkey-wasi-embedding.git
	shallow = true
```

.gitmodules
```diff
@@ -1,8 +1,8 @@
...
+	path = runtime/js-compute-runtime/gecko-dev
+	url = https://github.com/bytecodealliance/gecko-dev.git
 	shallow = true
-[submodule "tests/wpt-harness/wpt"]
-	path = tests/wpt-harness/wpt
-	url = https://github.com/web-platform-tests/wpt.git
+[submodule "runtime/sm"]
+	path = runtime/sm
+	url = https://github.com/fastly/spidermonkey-wasi-embedding.git
 	shallow = true
```
JakeChampion, line 8:
```suggestion
[submodule "tests/wpt-harness/wpt"]
	path = tests/wpt-harness/wpt
	url = https://github.com/web-platform-tests/wpt.git
	shallow = true
```

elliottt (commented):
What do you think about keeping the runtime/spidermonkey directory so that the build happens in a completely different tree? I think we should still try to make it easy to keep the two codebases separate so that we stick to the embedder api.

PR #639, opened 2023-08-29
feat: add ability to import ECDSA JWK keys via crypto.subtle.importKey

JakeChampion:
This implements the JWK subset of ECDSA importKey operation as defined within https://w3c.github.io/webcrypto/#ecdh-operations


runtime/js-compute-runtime/builtins/crypto.cpp
```diff
@@ -27,17 +28,19 @@ bool Crypto::get_random_values(JSContext *cx, unsigned argc, JS::Value *vp) {
...
   size_t byte_length = JS_GetArrayBufferViewByteLength(typed_array);
   if (byte_length > MAX_BYTE_LENGTH) {
-    JS_ReportErrorUTF8(cx,
-                       "crypto.getRandomValues: input byteLength must be at most %u, "
-                       "but is %zu",
-                       MAX_BYTE_LENGTH, byte_length);
+    std::string message = "crypto.getRandomValues: input byteLength must be at most ";
+    message += std::to_string(MAX_BYTE_LENGTH);
+    message += ", but is ";
+    message += std::to_string(byte_length);
```
elliottt, line 42: We should add a submodule for https://fmt.dev to simplify stuff like this.

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -1105,6 +1221,260 @@ JSObject *CryptoAlgorithmHMAC_Import::toObject(JSContext *cx) {
...
+      // 2.6. If the "ext" field of jwk is present and has the value false and extractable is true, then throw a DataError.
+      if (jwk->ext && !jwk->ext.value() && extractable) {
+        DOMException::raise(cx, "Data provided to an operation does not meet requirements", "DataError");
+        return nullptr;
+      }
+      // 2.7. Let namedCurve be a string whose value is equal to the "crv" field of jwk.
+      auto namedCurve = toNamedCurve(jwk->crv.value());
+      // 2.8. If namedCurve is not equal to the namedCurve member of normalizedAlgorithm, throw a DataError.
+      if (!namedCurve.has_value() || namedCurve.value() != this->namedCurve) {
+        DOMException::raise(cx, "Oopsie", "DataError");
```
elliottt, line 1283: Should this error message be updated? (Though I do like "Oopsie" :D )

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -1105,6 +1221,260 @@ JSObject *CryptoAlgorithmHMAC_Import::toObject(JSContext *cx) {
...
+
+  KeyData data;
+  switch (format) {
+    case CryptoKeyFormat::Jwk: {
+      // This handles step 2.3: If the "kty" field of jwk is not "EC", then throw a DataError.
+      auto jwk = JsonWebKey::parse(cx, key_data, "EC");
+      if (!jwk) {
+        return nullptr;
+      }
+      data = jwk.release();
```
elliottt, line 1408: No need to change it in this pr, but what do you think about making `KeyData` into variant that holds a `std::unique_ptr<JsonWebKey>` instead of a raw pointer? Otherwise we have to remember to free the key at some point in the future.

elliottt (approved):
This looks good to me!

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
JakeChampion: lol oh yes it really should

JakeChampion:
```suggestion
        DOMException::raise(cx, "The JWK's \"crv\" member specifies a different curve than requested", "DataError");
```

runtime/js-compute-runtime/core/event_loop.cpp
```diff
@@ -146,6 +146,7 @@ bool process_pending_request(JSContext *cx, JS::HandleObject request,
 
   auto res = pending.wait();
   if (auto *err = res.to_err()) {
+    // TODO: Add error.cause property with the underlying host error
```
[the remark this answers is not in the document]
JakeChampion: 👍 https://github.com/fastly/js-compute-runtime/issues/657

PR #641, opened 2023-08-31
fix: Add documentation for Request.prototype.clone()

JakeChampion:
This method was added in version 1.2.0 and never documented 🙈


elliottt (approved):
:tada:

PR #643, opened 2023-09-01
fix: update to latest version of gecko-dev which fixes a bug with the default ecma262 sorting algorithm

JakeChampion:


elliottt (approved):
:bug:

PR #649, opened 2023-09-07
fix: revert host_api implementation of the cache apis

JakeChampion:
the host_api implementation seems to have aan issue with regards to the fastly_cache::get_state hostcall whereby it will sometimes never return from the host

this commit is not a long-standing solution to the problem, but it does solve it in the short-term, allowing our customers to update to the latest js sdk and still able to use the caching apis we expose


elliottt (approved):
:+1:

PR #651, opened 2023-09-08
fix: use new host_api implementation for transactional lookups and inserts

JakeChampion:
This pull-request replaces the Simple Caches use of the direct caching host-calls with a cpp interface provided by host_api.
This also adds a missing CacheHandle::transaction_lookup method and fixes the CacheHandle::lookup to use a non-transactional lookup (which was the bug that causes the SimpleCache API to freeze an application in version 3.2.0 - 3.3.3). Thank you @acfoltzer for hunting this down 🙇


runtime/js-compute-runtime/builtins/cache-simple.cpp
```diff
@@ -213,6 +212,12 @@ JS::Result<std::string> createSurrogateKeysFromCacheKey(JSContext *cx, std::stri
   return JS::Result<std::string>(surrogate_keys);
 }
 
+#define CANCEL_TRANSACTION(cx, promise, handle)                                                    \
+  if (auto *err = handle.transaction_cancel().to_err()) {                                          \
+    host_api::handle_fastly_error(cx, *err, __LINE__, __func__);                                   \
+  }                                                                                                \
+  return RejectPromiseWithPendingError(cx, promise);
+
```
elliottt, line 220: It would add some code back in at the call-site, but what do you think about removing the return here and adding it back after each call to `CANCEL_TRANSACTION`? Using `return` in macros makes it hard to tell locally that control flow won't continue after the macro call.

PR #659, opened 2023-09-13
chore: Add fmtlib

elliottt:
Add a submodule for fmtlib 10.1.1 and port some uses of string concatenation to
`fmt::format` instead.

If we start using fmtlib in any of the shared builtins, we'll need to ensure that the `third_party/fmt/src/format.o` file gets included when we construct `shared.a`.


JakeChampion (approved):
Nice!

PR #664, opened 2023-09-20
clarify that the note is about Backend constructors and not the allowDynamicBackends function

JakeChampion:


kailan (approved):
so speedy!

PR #667, opened 2023-09-21
feat: add support for ECDSA keys to be used with SubtleCrypto.prototype.sign and SubtleCrypto.prototype.verify

### LINKED ISSUE #663: Incorrect implementation for ECDSA verification
When provided with a signature signed by a ECDSA CryptoKey pair, verifying the signature with the same key-pair should should work but currently does not

JakeChampion:
In the last release we added support for importing an ECDSA key with SubtleCrypto.prototype.importKey. This patch adds the ability to use these keys to sign data and to verify data was signed with the same underlying key.

With this patch, Fastly Compute customers are now able to validate signatures provided by Fastly Fanout requests

fixes #663 

Here is an example of how a Fastly Customer can verify Fastly Fanout requests:
<https://fanout-cf.edgecompute.app/>
```js

import { createFanoutHandoff } from "fastly:fanout";
import { env } from 'fastly:env';
import { importJWK, jwtVerify } from "jose";

// A JWK version of the public key documented on https://developer.fastly.com/learning/concepts/real-time-messaging/fanout/#validating-grip-requests
const fanoutJWK = importJWK({
  kty: "EC",
  crv: "P-256",
  x: "CKo5A1ebyFcnmVV8SE5On-8G81JyBjSvcrx4VLetWCg",
  y: "7gwJqaU6N8TP88--twjkwoB36f-pT3QsmI46nPhjO7M",
}, "ES256");

async function app(event) {
  console.log('FASTLY_SERVICE_VERSION', env('FASTLY_SERVICE_VERSION'));

  const request = event.request;

  // Request might be from Fanout
  if (request.headers.has('Grip-Sig')) {
    try {
      const { payload, protectedHeader } = await jwtVerify(request.headers.get('Grip-Sig'), await fanoutJWK)

      return Response.json({ payload, protectedHeader }, { 'headers': { 'content-type': 'application/json' } });
    } catch (error) {
      return Response.json({ error }, { status: 500 });
    }
  } else {
    // Not from Fanout, hand it off to Fanout to manage
    return createFanoutHandoff(request, 'self');
  }
}

addEventListener('fetch', event => event.respondWith(app(event)))
```

### DIFF
--- a/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
+++ b/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
@@ -33,8 +33,7 @@ convertToBytesExpand(JSContext *cx, const BIGNUM *bignum, size_t minimumBufferSi
   size_t paddingLength = bufferSize - length;
   if (paddingLength > 0) {
     uint8_t padding = BN_is_negative(bignum) ? 0xFF : 0x00;
-    for (size_t i = 0; i < paddingLength; i++)
-      bytes[i] = padding;
+    std::fill_n(bytes + i, paddingLength, padding);
   }
   BN_bn2bin(bignum, bytes.get() + paddingLength);
   return std::pair<mozilla::UniquePtr<uint8_t[], JS::FreePolicy>, size_t>(std::move(bytes),
--- a/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
+++ b/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
@@ -889,8 +889,8 @@ JSObject *CryptoAlgorithmECDSA_Sign_Verify::sign(JSContext *cx, JS::HandleObject
   auto rBytesSize = rBytesAndSize.second;
 
   auto sBytesAndSize = convertToBytesExpand(cx, s, keySizeInBytes);
-  auto sBytes = std::move(sBytesAndSize.first.get());
-  auto sBytesSize = std::move(sBytesAndSize.second);
+  auto *sBytes = sBytesAndSize.first.get();
+  auto sBytesSize = sBytesAndSize.second;
 
   auto resultSize = rBytesSize + sBytesSize;
   mozilla::UniquePtr<uint8_t[], JS::FreePolicy> result{
--- a/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
+++ b/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
@@ -894,7 +894,7 @@ JSObject *CryptoAlgorithmECDSA_Sign_Verify::sign(JSContext *cx, JS::HandleObject
 
   auto sBytesAndSize = convertToBytesExpand(cx, s, keySizeInBytes);
   if (!sBytesAndSize.has_value()) {}
-  auto sBytes = std::move(sBytesAndSize.value().first);
+  auto *sBytes = sBytesAndSize->first.get();
   auto sBytesSize = sBytesAndSize->second;
 
   auto resultSize = rBytesSize + sBytesSize;
--- a/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
+++ b/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
@@ -19,7 +19,7 @@ namespace builtins {
 namespace {
 int numBitsToBytes(int x) { return (x / 8) + (7 + (x % 8)) / 8; }
 
-std::optional<std::pair<mozilla::UniquePtr<uint8_t[], JS::FreePolicy>, size_t>>
+std::pair<mozilla::UniquePtr<uint8_t[], JS::FreePolicy>, size_t>
 convertToBytesExpand(JSContext *cx, const BIGNUM *bignum, size_t minimumBufferSize) {
   int length = BN_num_bytes(bignum);
 
@@ -30,7 +30,7 @@ convertToBytesExpand(JSContext *cx, const BIGNUM *bignum, size_t minimumBufferSi
   size_t paddingLength = bufferSize - length;
   if (paddingLength > 0) {
     uint8_t padding = BN_is_negative(bignum) ? 0xFF : 0x00;
-    std::fill_n(bytes + i, paddingLength, padding);
+    std::fill_n(bytes.get(), paddingLength, padding);
   }
   BN_bn2bin(bignum, bytes.get() + paddingLength);
   return std::pair<mozilla::UniquePtr<uint8_t[], JS::FreePolicy>, size_t>(std::move(bytes),
--- a/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
+++ b/runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
@@ -895,7 +895,7 @@ JSObject *CryptoAlgorithmECDSA_Sign_Verify::sign(JSContext *cx, JS::HandleObject
   auto sBytesAndSize = convertToBytesExpand(cx, s, keySizeInBytes);
   if (!sBytesAndSize.has_value()) {}
   auto sBytes = std::move(sBytesAndSize.value().first);
-  auto sBytesSize = std::move(sBytesAndSize.value().second);
+  auto sBytesSize = sBytesAndSize->second;
 
   auto resultSize = rBytesSize + sBytesSize;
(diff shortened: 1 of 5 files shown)


runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +827,167 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+  auto sBytesSize = std::move(sBytesAndSize.value().second);
+
+  auto resultSize = rBytesSize + sBytesSize;
+  mozilla::UniquePtr<uint8_t[], JS::FreePolicy> result{
+      static_cast<uint8_t *>(JS_malloc(cx, resultSize))};
+
+  std::memcpy(result.get(), rBytes.get(), rBytesSize);
+  std::memcpy(result.get() + rBytesSize, sBytes.get(), sBytesSize);
+  static_cast<void>(rBytes.release());
+  static_cast<void>(sBytes.release());
```
JakeChampion, line 908: @elliottt I feel like this part could likely be improved but not too sure what I'd do different

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -16,6 +17,53 @@
...
+
+  size_t bufferSize = std::max<size_t>(length, minimumBufferSize);
+  mozilla::UniquePtr<uint8_t[], JS::FreePolicy> bytes{
+      static_cast<uint8_t *>(JS_malloc(cx, bufferSize))};
+
+  size_t paddingLength = bufferSize - length;
+  if (paddingLength > 0) {
+    uint8_t padding = BN_is_negative(bignum) ? 0xFF : 0x00;
+    for (size_t i = 0; i < paddingLength; i++)
+      bytes[i] = padding;
```
elliottt, line 37:
```suggestion
    std::fill_n(bytes + i, paddingLength, padding);
```

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
elliottt: You could modify `convertToBytesExpand` to accept the output pointer to write to, and do the realloc there. However, I think what you've done is totally reasonable, and I wouldn't worry about being that careful with memory for a function that shouldn't be called on a hot path.

One change that I would make is to change the uses of `std::move` to the `.get()` method on the `UniquePtr`, as that will give you access to the pointer for copying. That way you can remove the `.release()` calls, as the memory will still be managed by the `UniquePtr`s.

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +827,167 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+  ECDSA_SIG_get0(sig, &r, &s);
+  auto keySize = curveSize(cx, key);
+  if (keySize.isErr()) {
+    return nullptr;
+  }
+
+  size_t keySizeInBytes = numBitsToBytes(keySize.unwrap());
+
+  auto rBytesAndSize = convertToBytesExpand(cx, r, keySizeInBytes);
+  if (!rBytesAndSize.has_value()) {}
```
elliottt, line 892: Did you mean to return an error here?

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +827,167 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+  auto rBytes = std::move(rBytesAndSize.value().first);
+  auto rBytesSize = std::move(rBytesAndSize.value().second);
+
+  auto sBytesAndSize = convertToBytesExpand(cx, s, keySizeInBytes);
+  if (!sBytesAndSize.has_value()) {}
+  auto sBytes = std::move(sBytesAndSize.value().first);
```
elliottt, line 898: You can use `->` to do an unchecked access the contents of an optional, and `.get()` to get a pointer to the memory managed by the `UniquePtr`. Since you're checking above that it has a value, you should be ok using `->` as long as you error appropriately in the conditional body.

```suggestion
  auto *sBytes = sBytesAndSize->first.get();
```

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +827,167 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+  auto sBytesSize = std::move(sBytesAndSize.value().second);
```
elliottt, line 899: You can copy here, as the value of `.second` is just a `size_t`.

```suggestion
  auto sBytesSize = sBytesAndSize->second;
```

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -16,6 +17,53 @@
...
 
 namespace {
+int numBitsToBytes(int x) { return (x / 8) + (7 + (x % 8)) / 8; }
+
+std::optional<std::pair<mozilla::UniquePtr<uint8_t[], JS::FreePolicy>, size_t>>
+convertToBytesExpand(JSContext *cx, const BIGNUM *bignum, size_t minimumBufferSize) {
+  int length = BN_num_bytes(bignum);
+  if (length < 0) {
+    return std::nullopt;
+  }
```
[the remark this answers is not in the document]
JakeChampion:
```suggestion
  int length = BN_num_bytes(bignum);
```

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
JakeChampion: I did, but with your understanding that BN_num_bytes would never return negatives, it meant that we'd never return a nullopt_t, so now convertToBytesExpand does not use a std::optional in it's returned type -- [`e76c046` (#667)](https://github.com/fastly/js-compute-runtime/pull/667/commits/e76c046e2b87871fd0682aedfea224e4b51904f7)

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
JakeChampion: Good idea, I applied what I think you mean in this commit -- [`c78b288` (#667)](https://github.com/fastly/js-compute-runtime/pull/667/commits/c78b288f4f36aaf382407f66ebc2ab763ec4ac86)

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +823,163 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+  auto keySize = curveSize(cx, key);
+  if (keySize.isErr()) {
+    return nullptr;
+  }
+
+  size_t keySizeInBytes = numBitsToBytes(keySize.unwrap());
+
+  auto rBytesAndSize = convertToBytesExpand(cx, r, keySizeInBytes);
+  auto rBytes = std::move(rBytesAndSize.first.get());
+  auto rBytesSize = std::move(rBytesAndSize.second);
```
elliottt, line 889: You don't need to move these: there's no transfer of ownership because you're copying the pointer value, or the `size_t` value out.
```suggestion
  auto *rBytes = rBytesAndSize.first.get();
  auto rBytesSize = rBytesAndSize.second;
```

runtime/js-compute-runtime/builtins/crypto-algorithm.cpp
```diff
@@ -755,12 +823,163 @@ JSObject *CryptoAlgorithmHMAC_Sign_Verify::toObject(JSContext *cx) {
...
+
+  auto sBytesAndSize = convertToBytesExpand(cx, s, keySizeInBytes);
+  auto sBytes = std::move(sBytesAndSize.first.get());
+  auto sBytesSize = std::move(sBytesAndSize.second);
```
elliottt, line 893: Same with these two.
```suggestion
  auto *sBytes = sBytesAndSize.first.get();
  auto sBytesSize = sBytesAndSize.second;
```

elliottt (approved):
:rocket:

PR #672, opened 2023-09-25
chore: Build with `-mmultivalue`

elliottt:
Build spidermonkey and the js-compute-runtime with `-mmultivalue` to improve codegen for 128-bit math.

Interestingly, `wasm-opt` seems to fail to strip the runtime when it's built with `-mmultivalue`, and additionally `wasm-tools validate -f multi-value` indicates that there's a function that does not typecheck.


elliottt:
Closing, as it's not clear if we also need wasi-libc to be built with `-mmultivalue` for this to work.

PR #673, opened 2023-09-25
refactor: consolidate all test applications into one test application

JakeChampion:
by merging the test applications into one, we are able to run all the test suites locally much faster than before because it involves compiling to wasm once and spinning up viceroy once, compared to the ~50 wasms and viceroys we were using before.

It should also help reduce the amount of flakes we see when publishing to Fastly and reduce some constraints on internal Fastly systems.


JakeChampion:
we should also save on GitHub Actions minutes with this new approach
Before this PR, E2E and SDK Tests used ~2505 seconds on GitHub Actions
After this PR ~258 seconds on GitHub Actions

runtime/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -280,7 +280,7 @@ const uint8_t base64DecodeTable[128] = {
   /* 64 */   nonAlphabet,           0,           1,           2,           3,           4,           5,           6,
   /* 72 */             7,           8,           9,          10,          11,          12,          13,          14,
   /* 80 */            15,          16,          17,          18,          19,          20,          21,          22,
-  /* 88 */            23,          24,          25, nonAlphabet, nonAlphabet, nonAlphabet, nonAlphabet,          63,
+  /* 88 */            23,          24,          25, nonAlphabet, nonAlphabet, nonAlphabet, nonAlphabet, nonAlphabet,
```
JakeChampion, line 283: This PR also includes a commit which fixes our atob implementation to throw an error if the supplied input contains an underscore

elliottt (approved):
Amazing! :tada:

PR #683, opened 2023-10-05
fix: improve fetch error messages

JakeChampion:
This updates the SDK to use the new pending-req-wait-v2 function which supercedes pending-req-wait by providing a struct with rich error details (if the send has errored).

Instead of always saying "NetworkError when attempting to fetch resource." - we now will say things such as "NetworkError: DNS timeout", "NetworkError: TLS configuration error", or "NetworkError: HTTP response status invalid".


runtime/js-compute-runtime/host_interface/host_api.h
```diff
@@ -13,6 +14,8 @@
 #include "core/allocator.h"
 #include "js/TypeDecls.h"
 
+#include "host_interface/fastly.h"
```
JakeChampion, line 17: @elliottt - I feel like I shouldn't be pulling this in here but couldn't think of anything better to do...

elliottt: I pushed a commit to this branch to address that. Basically, instead of having an explicit constructor, have a private function in host_api.cpp that makes a value of `FastlySendError` from the type that comes from fastly.h. As the only place the error is constructed is in host_api.cpp, it works out fine to have that as a private function in there.

elliottt (approved):
Looks good to me!

PR #685, opened 2023-10-06
docs: rename to Fastly Compute

JakeChampion:


elliottt (approved):
Skimming through, this all looks good to me!

PR #689, opened 2023-10-13
fix: raise an error during wizening for async functions given to addEventListener

### LINKED ISSUE #687: Raise an error during wizening for async functions given to addEventListener
`addEventListener` must not be given an async function, as this will terminate execution at the first `await` encountered. Instead, the function given as the callback must be rewritten as `e => e.respondWith(async_function(e))`. We can detect this during wizening, and instead raise an error that indicates how to fix the problem, rather than delay the error to when a request is being handled.

JakeChampion:
resolves #687

Below is what what it looks like if you try and register an async fetchevent handler

>❯ fastly compute serve
>✓ Verifying fastly.toml
>✓ Identifying package name
>✓ Identifying toolchain
>✗ Running [scripts.build]
>
>INFO: Command output:
>--------------------------------------------------------------------------------
>Error: addEventListener: Argument 2 should not be async function. Please rewrite this to be a synchronous function which returns a Response or a Promise of a Response.
>Error: the `wizer.initialize` function trapped

The fetch event handler function has to be sync and not async because the Event Model itself is synchronous and you can register multiple FetchEventHandlers.
If you pass an async function as the FetchEventHandler and await  before your call to event.responseWith - then the FetchEventHandler will return a Promise, when the FetchEventHandler returns anything ([step 24.3.12 of Handle Fetch algorithm](https://w3c.github.io/ServiceWorker/#handle-fetch)), the specification states we need to check if the the respond-with flag has been set, and it would not have been set - we then need to call the next registered FetchEventHandler - doing the same steps, until one calls event.respondWith or we have called them all.

This thread has some more context: https://github.com/w3c/ServiceWorker/issues/836


runtime/js-compute-runtime/js-compute-runtime.cpp
```diff
@@ -361,6 +362,15 @@ static bool addEventListener(JSContext *cx, unsigned argc, Value *vp) {
     exit(1);
   }
 
+  auto func_class = JS::GetClass(&val.toObject());
+
+  if (func_class != js::ProtoKeyToClass(JSProtoKey::JSProto_AsyncFunction)) {
```
elliottt, line 367: Should this be `==` instead of `!=`?

PR #694, opened 2023-10-17
✏ Tighten up docs for `SimpleCache.getOrSet`

acfoltzer:
Hopefully clarifies that the function argument is called if and only if the key is not present in the cache.


JakeChampion (approved):
Thank you for this 🙏

PR #697, opened 2023-10-23
Improve our spec compliance for Request and Response classes and the Response.redirect static method

JakeChampion:
This patch correct a bunch of error types within our Request and Response classes to match those within the specifications. You should see that our web-platform-tests results now have more assertions passing!

This also includes a fix for our Response.redirect static method, previously it was creating a Headers instance which could be modified, which is incorrect, it's meant to be an immutable Headers instance and should throw errors if any modifications are attempted such as deleting or adding a header.


elliottt (commented):
The changes look good, but there are a few tests that flipped from `PASS` to `FAIL`, was that expected?

JakeChampion:
> The changes look good, but there are a few tests that flipped from `PASS` to `FAIL`, was that expected?

It wasn't, the tests caught a regression (woohoo!) and I've fixed that in [`43f2ec6` (#697)](https://github.com/fastly/js-compute-runtime/pull/697/commits/43f2ec684983bbd07531d87ade3f542f84ffa990)

elliottt (approved):
Awesome!

JakeChampion:
I'm going to rebase this because one commit message has a typo of 'fox:' instead of 'fix:' 😂 
>[fox: Make Response.redirect headers be immutable](https://github.com/fastly/js-compute-runtime/pull/697/commits/8cc821ee16532050b5c12ddb30f4e2f528b031ed)

PR #701, opened 2023-10-25
fix: Make the underlying KVStore.prototype.get implementation be async

JakeChampion:
We model the JS API for KVStore methods as async promise-returning methods however, the internal implementation is currently synchronous due to the host-call itself being synchronous.

This patch updates our implementation to use the new asynchronous host-call for KVStore lookups.

The great thing about this change, is it is not a user-facing API change as we were already returning a Promise. This PR makes that Promise resolution happen asynchronously ☺️


elliottt (approved):
Looks great!

PR #705, opened 2023-11-15
feat: Add `manualFramingHeaders` on RequestInit and ResponseInit, and add Request.prototype.setManualFramingHeaders and Response.prototype.setManualFramingHeaders

### LINKED ISSUE #5: Support overriding `Content-Length` header
Client set up a new JS C@E service and everything is working correctly with the exception that our content-length header is not being returned. Turns out in practice this doesn’t matter a ton because clients are looking for the content-range which is being returned correctly. But we would definitely want this fixed before completely switching over to C@E in production. 

@tschneidereit

elliottt:
This adds the ability for the JavaScript application to control how Request/Response Framing Headers should be determined.

By default the framing headers are set to "automatic" mode, which means they are created based on the body of the associated Request/Response instance.
In "automatic" mode, a `Content-Length` is used when the size of the body can be determined before it is sent.
Requests/Responses sent in streaming mode, where headers are sent immediately but the content of the body is streamed later, will receive a `Transfer-Encoding: chunked` to accommodate the dynamic generation of the body.
In "manual" mode, any `Content-Length` or `Transfer-Encoding` headers will be honored.
You must ensure that those headers have correct values permitted by the [HTTP/1.1 specification](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.1).
If the provided headers are not permitted by the specification, the headers will revert to "automatic" mode and a diagnostic message will be logged about what was wrong.
If a `Content-Length` is permitted by the specification, but the value does not match the size of the actual body, the body will either be truncated (if it is too long), or the connection will be hung up early (if it is too short).

Fixes #5


runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -1655,14 +1655,26 @@ JSObject *Request::create(JSContext *cx, JS::HandleObject requestInstance, JS::H
...
         !JS_GetProperty(cx, init, "body", &body_val) ||
         !JS_GetProperty(cx, init, "backend", &backend_val) ||
         !JS_GetProperty(cx, init, "cacheOverride", &cache_override) ||
-        !JS_GetProperty(cx, init, "fastly", &fastly_val)) {
+        !JS_GetProperty(cx, init, "fastly", &fastly_val) ||
+        !JS_GetProperty(cx, init, "overrideContentLength", &overrideContentLength)) {
       return nullptr;
     }
+
+    if (overrideContentLength.isBoolean() && overrideContentLength.toBoolean()) {
```
JakeChampion, line 1669: If we use `JS::ToBoolean`, then the value will be cast to a boolean, which is usually how people expect JavaScript APIs to work - for example [we use that here](https://github.com/fastly/js-compute-runtime/blob/4f399434c0959e902df03262dfceefdc16592afe/runtime/js-compute-runtime/builtins/request-response.cpp#L1963)

runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2687,6 +2700,16 @@ bool Response::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
       statusText = JS_NewStringCopyZ(cx, status_text.c_str());
     }
 
+    JS::RootedValue overrideContentLength(cx);
+    if (!JS_GetProperty(cx, init, "overrideContentLength", &overrideContentLength)) {
+      return false;
+    }
+
+    // `overrideContentLength: true` indicates that we want to set the framing mode manually.
+    if (overrideContentLength.isBoolean() && overrideContentLength.toBoolean()) {
```
JakeChampion, line 2709: Same here as https://github.com/fastly/js-compute-runtime/pull/705/files#r1439668102

elliottt (commented):
Looks good to me! I can't approve, because it's my own PR :sweat_smile:

runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -2735,6 +2845,29 @@ bool Response::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+      manualFramingHeaders.setBoolean(false);
+    }
+  }
+  JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::ManualFramingHeaders),
+                      JS::BooleanValue(JS::ToBoolean(manualFramingHeaders)));
+
+  // `manualFramingHeaders: true` indicates that we want to set the framing mode manually.
+  if (JS::ToBoolean(manualFramingHeaders)) {
+    mode = host_api::FramingHeadersMode::ManuallyFromHeaders;
+  }
```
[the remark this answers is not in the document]
JakeChampion: I'm not sure on a utility function for this as one place is dealing with Request instances and RequestInit instances, and the other place is dealing with Response and ResponseInit instances. If you think their is a nice way to package that up into a utility function, could we do that as a separate refactoring pull-request?

JakeChampion (approved):

PR #715, opened 2024-01-11
fix: parse latin-1 encoded field values correctly

JakeChampion:
This updates our field-value definition in wit from `string` to `list<u8>` as field-values are byte-strings. Finally, this update our field-value parsing implementation to make it understand how to parse latin-1 encoded field-values.


runtime/js-compute-runtime/builtins/headers.cpp
```diff
@@ -286,9 +286,26 @@ bool retrieve_value_for_header_from_handle(JSContext *cx, JS::HandleObject self,
...
+
+    JS::RootedString val_str(cx);
+    switch (encoding) {
+    case JS::SmallestEncoding::ASCII: {
+      val_str.set(JS_NewStringCopyN(cx, reinterpret_cast<char *>(str.ptr.get()), str.len));
+      break;
+    }
+    case JS::SmallestEncoding::Latin1: {
+      val_str.set(JS_NewStringCopyUTF8N(
+          cx, JS::UTF8Chars(reinterpret_cast<char *>(str.ptr.get()), str.len)));
```
JakeChampion, line 301:
```suggestion
      val_str.set(JS_NewStringCopyUTF8N(cx, chars));
```

runtime/js-compute-runtime/js-compute-builtins.cpp
```diff
@@ -1176,3 +1176,19 @@ bool print_stack(JSContext *cx, FILE *fp) {
     return false;
   return print_stack(cx, stackp, fp);
 }
+
+JS::Result<JS::HandleString> host_string_to_js_string(JSContext *cx, host_api::HostString &str) {
+  JS::RootedString val_str(cx);
+  val_str = JS_NewStringCopyUTF8N(cx, JS::UTF8Chars(str.ptr.get(), str.len));
+  if (!val_str) {
+    if (JS_IsExceptionPending(cx)) {
+      JS_ClearPendingException(cx);
```
elliottt, line 1185: Should this function use `JS::FindSmallestEncoding` as well?

elliottt (approved):
This all looks great to me! Thank you for fixing the host interface -- field values should never have been host strings.

PR #718, opened 2024-01-18
update to rust 1.75.0

JakeChampion:


elliottt (approved):
Very reasonable to me, but it looks like we might be using a cached spidermonkey somewhere, given the linker errors in the debug build.

PR #719, opened 2024-01-18
fix: ensure we honor first-byte-timeout and between-bytes-timeout for dynamically registered backends

JakeChampion:
Previously we were not setting the fields in the bitmask to indicate to Fastly Compute that these fields should be read


elliottt (approved):
Looks good!

PR #725, opened 2024-02-16
Add step to publish to GitHub package registry

bekahs4:
### Summary

Add a step to the `release-please.yml` GitHub Actions workflow that publishes to the Fastly internal GItHub package registry [as discussed in Slack](https://fastly.slack.com/archives/C022F76SJ83/p1707421784242279).

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -151,6 +151,19 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
 
+      - name: github package registry publish
+        run: |
+          cat << EOF > .npmrc
+          //npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}
+          @fastly:registry=https://npm.pkg.github.com
+          registry=https://registry.npmjs.org/
+          always-auth=true
+          EOF
+          npm publish
+          rm .npmrc
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - run: yarn
         working-directory: ./documentation
 


guybedford (approved):
Looks good, happy to merge and try this out for the next release. I suppose we should just run with this approach for all `@fastly` packages until GitHub come up with a solution.

bekahs4:
Thanks @guybedford. Looks like I still have a failing test after a re-run. Do you have any insight on what's going on with that? I don't expect this PR to be related to the test failure.

guybedford:
@bekahs4 the CI issue is unrelated, so I've gone ahead and landed this in the mean time - will apply for the next release.

bekahs4:
Thanks @guybedford!

guybedford:
@bekahs4 I see the first release just failed here - https://github.com/fastly/js-compute-runtime/actions/runs/7979097116/job/21785698469 due to credentials. Any ideas at what level we need to address that? Happy to take follow-ups as necessary here.

bekahs4:
@guybedford I'll try to take a look at this later with @psbanka. On first glance I would be surprised if the `release` job failure is related to this change because this PR only made changes to the `publish` job.

guybedford:
Taking a look at that further, I believe this is also unrelated to this PR, and rather a separate issue with the release check. We can verify this properly when we tag the next release though, and if there are any issues with the GitHub package publish we can follow-up at that time.

bekahs4:
Thanks for taking another look @guybedford. For planning purposes on our end, is there a timeline for the next tagged release?

guybedford:
@bekahs4 v3.8.3 just went out and it looks like the GitHub release worked correctly.

PR #726, opened 2024-02-16
fix: do not use colon character in types for windows support

guybedford:
This removes the use of `:` in the types files for Windows compatibility.


JakeChampion (commented):
I think this would be a breaking change for our types as projects which import the types right now would have to update their import locations as the old locations no longer exist. To indicate a breaking fix, I believe the commit message convention would be "fix!:" instead of "fix:"

guybedford:
I guess it would only affect users reaching into the types by reference `node_modules/@fastly/js-compute/types/fastly:kv-store`? Strictly speaking that is public API though. But most users probably get the `index.d.ts` reference right?

guybedford:
I've moved back to a `fix` from `fix!`commit as I'm fairly sure this is non-breaking for most users. But please correct me if I've missed something here.

PR #738, opened 2024-03-08
feat: add fastly:device module which allows applications to detect a device based on a user-agent

JakeChampion:
An example application looks like this:

``` javascript
import { Device } from "fastly:device";
    event.respondWith(
      Response.json(
        Device.lookup(
          event.request.headers.get('user-agent')
        )
      )
    )
})

```


elliottt (approved):
Awesome! :tada:

PR #739, opened 2024-03-12
docs: deprecate the Dictionary Class and advice applications to use the ConfigStore class

JakeChampion:
Dictionary has been deprecated on Fastly Compute for a long time, we forgot to document that in the JavaScript Compute SDK

We intended to do this change in 0.5.4 (https://github.com/fastly/js-compute-runtime/blob/main/CHANGELOG.md#054-2022-09-28) when we added ConfigStore but we forgot to update the documentation

### DIFF
--- a/types/config-store.d.ts
+++ b/types/config-store.d.ts
@@ -4,58 +4,6 @@ declare module "fastly:config-store" {
    * Class for accessing [Fastly Edge Dictionaries](https://docs.fastly.com/en/guides/about-edge-dictionaries).
    *
    * **Note**: Can only be used when processing requests, not during build-time initialization.
-   * 
-   * @example
-   * <script async defer src="https://fiddle.fastly.dev/embed.js"></script>
-   * In this example we have an Edge Dictionary named 'animals' and we return the 'cat'
-   * entry as the response body to the client.
-   * 
-   * <script type="application/json+fiddle">
-   * {
-   *   "type": "javascript",
-   *   "title": "ConfigStore Example",
-   *   "origins": [
-   *     "https://http-me.glitch.me"
-   *   ],
-   *   "src": {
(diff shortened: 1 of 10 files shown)


types/config-store.d.ts
```diff
@@ -4,58 +4,6 @@ declare module "fastly:config-store" {
    * Class for accessing [Fastly Edge Dictionaries](https://docs.fastly.com/en/guides/about-edge-dictionaries).
    *
    * **Note**: Can only be used when processing requests, not during build-time initialization.
-   * 
-   * @example
-   * <script async defer src="https://fiddle.fastly.dev/embed.js"></script>
-   * In this example we have an Edge Dictionary named 'animals' and we return the 'cat'
-   * entry as the response body to the client.
-   * 
-   * <script type="application/json+fiddle">
```
JakeChampion, line 13: This big example comment is not useful in IDE's -- this is a legacy example comment from when our documentation website was generated from the type definitions

types/dictionary.d.ts
```diff
@@ -3,71 +3,24 @@ declare module "fastly:dictionary" {
...
    *
    * **Note**: Can only be used when processing requests, not during build-time initialization.
-   * 
-   * @example
-   * <script async defer src="https://fiddle.fastly.dev/embed.js"></script>
-   * In this example we have an Edge Dictionary named 'animals' and we return the 'cat'
-   * entry as the response body to the client.
-   * 
-   * <script type="application/json+fiddle">
-   * {
```
JakeChampion, line 13: Same here -- 
This big example comment is not useful in IDE's -- this is a legacy example comment from when our documentation website was generated from the type definitions

kphillips: Looks good. Thank you for taking care of this.

kphillips (approved):
No objections here. Thanks!

elliottt (approved):
:tada:

Issue #729, opened 2024-02-26
Support top-level await


guybedford:
To support top-level await, it should be enough simply to configure esbuild to transpile it into an immediately executing promise function, where the error handler throws a global error and the completion handler is a noop.

With that, top-level await would correspond to running at preinitialization time since we already drive the event loop for preinitialization.

Later on, we could have more native level support if we support ES modules, but the semantics would remain the same, even on top of an esbuild virtualization.

guybedford:
The blocker here is esbuild not supporting TLA at all for bundling unfortunately.

guybedford:
Actually it turns out that per `Initial support for bundling with top-level await` in https://github.com/evanw/esbuild/releases/tag/v0.10.0 this is supported only for the `esm` output format in esbuild.

We currently use the `iife` output, so switching to the `esm` output sounds like the correct way to achieve this.

harmony7:
This has been implemented and can be enabled an experimental flag `--experimental-enable-top-level-await`, and is available starting v3.11.0.

Should we leave this issue open until it leaves experimental status?

PR #742, opened 2024-03-13
feat: add new flag --experimental-enable-top-level-await

harmony7:
This PR is a fix for #729

These fixes were enough to get top-level-await in both my main module and dependencies working for me.

I would love to add a test suite, but wasn't able to figure out how to do it. Does the setup require a linux machine like it says in the DEVELOPING readme file? I'm on a Mac...

The following minimal case works (requires `"type": "module"` in `package.json`):

src/index.js
```javascript

import { foo } from './dep.js';
const value = await foo();

  return new Response(value, { status: 200 }); // returns 'bar'
}
```

src/dep.js
```javascript
export async function wait() {
  return Promise.resolve();
}

await wait();

export async function foo() {
  await wait();
  return 'bar';
}
```

These changes also enabled me to use top level await in my upcoming version of compute-js-opentelemetry that no longer uses Webpack.

I've also tested and seen that `"type": "commonjs"` in `package.json` still works.

### DIFF
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -62,13 +62,13 @@ export const TransactionCacheEntry = globalThis.TransactionCacheEntry;
   },
 }
 
-export async function bundle(input) {
+export async function bundle(input, enableExperimentalTopLevelAwait = false) {
   return await build({
     conditions: ['fastly'],
     entryPoints: [input],
     bundle: true,
     write: false,
-    format: 'esm',
+    format: enableExperimentalTopLevelAwait ? 'esm' : 'iife',
     tsconfig: undefined,
     plugins: [fastlyPlugin],
   })
--- a/src/bundle.js
+++ b/src/bundle.js
@@ -68,6 +68,7 @@ export async function bundle(input) {
     entryPoints: [input],
     bundle: true,
     write: false,
+    format: 'esm',
     tsconfig: undefined,
     plugins: [fastlyPlugin],
   })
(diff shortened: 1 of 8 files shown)


harmony7:
Also, WRT to error handling, it seems that with this implementation, if the top level await throws an exception, this happens:

```javascript
export async function foo() {
  throw new Error('foo');
}
```

```
Promise rejected but never handled: foo
Stack:
  foo@<stdin>:7:9
  @<stdin>:13:19
  async*@<stdin>:18:3
```

If more robust handling is required, please advise.

src/bundle.js
```diff
@@ -68,6 +68,7 @@ export async function bundle(input) {
     entryPoints: [input],
     bundle: true,
     write: false,
+    format: 'esm',
```
JakeChampion, line 71: From [esbuild docs](https://esbuild.github.io/api/#format-esm):
>The esm format stands for "ECMAScript module". It assumes the environment supports import and export syntax.

We don't support import and export syntax, do you know if any issues will arise from this change?

harmony7:
I've tried putting it behind a flag in its own changeset

harmony7:
I've pushed a little more code.

The first changeset enables TLA in two steps:

* `bundle()` uses `format: 'esm'` when calling esbuild (also updated `precompile()` to use `sourceType: 'module'`)
* Added `enableTopLevelAwait()` that:
  * removes export declarations from the bundle (taking care to leave the declaration part intact for exports that have them)
  * wraps the entire thing in an async iife

In the second changeset:
All this is only enabled if `--experimental-enable-top-level-await` is specified. Otherwise the behavior is identical to the prior behavior.

JakeChampion (approved):

PR #744, opened 2024-03-27
update to gecko-dev 123.0.1 with our js streams and pbl patches

JakeChampion:
bytecodealliance/gecko-dev#41

This also involved updating our uses of `JS::NewArrayBufferWithContents` as in the newer version of gecko-dev we need to tell it what we want to happen if an out-of-memory error occurs. For our application, all uses of this function want to free the memory themselves so we pass `JS::NewArrayBufferOutOfMemory::CallerMustFreeMemory`

This also involved explicitly including mozilla/Try.h so we can use the MOZ_TRY macro


elliottt (approved):
Awesome work!

PR #751, opened 2024-04-09
feat: Add KVStore.prototype.delete method

### LINKED ISSUE #746: Feature Request: KVStore.delete
Currently `KVStore` only has a `get` and a `put` method exposed to JS: https://github.com/fastly/js-compute-runtime/blob/f15b6807bbb4d5273e50b6ca16170827dae1c31d/runtime/js-compute-runtime/builtins/kv-store.cpp#L401-L405

The Rust SDK [supports the delete method](https://docs.rs/fastly/latest/src/fastly/kv_store.rs.html#143-148) as well.

Is there a plan to add a `delete` method to `KVStore` in the JS SDK as well? Or is there any undocumented way to delete entries (e.g. by calling `put` with an empty value etc)?

JakeChampion:
Resolves #746 

In this example we connect to an KV Store named `'files'` and save an entry to the store under the key `'hello'` and then delete the entry.

```js

import { KVStore } from "fastly:kv-store";

async function app(event) {
  const files = new KVStore('files')

  await files.put('hello', 'world')
  await files.delete('hello')

  const entry = await files.get('hello')
  if (entry) {
    return new Response(await entry.text())
  } else {
    return new Response('no file named hello exists')
  }

}

addEventListener("fetch", (event) => event.respondWith(app(event)))

```

Note: It is perfectly fine to delete an entry that does not exist 👍


runtime/js-compute-runtime/core/event_loop.cpp
```diff
@@ -301,7 +305,11 @@ bool EventLoop::process_pending_async_tasks(JSContext *cx) {
   if (builtins::Request::is_instance(ready_obj)) {
     ok = process_pending_request(cx, ready_obj, host_api::HttpPendingReq{ready_handle});
   } else if (builtins::KVStore::is_instance(ready_obj)) {
-    ok = builtins::KVStore::process_pending_kv_store_lookup(cx, ready_obj);
+    if (builtins::KVStore::has_pending_delete_handle(ready_obj)) {
+      ok = builtins::KVStore::process_pending_kv_store_delete(cx, ready_obj);
+    } else {
+      ok = builtins::KVStore::process_pending_kv_store_lookup(cx, ready_obj);
+    }
```
JakeChampion, line 312: Note: Something needs to be thought about here -- the KVStore instance could have both a pending lookup and a pending delete, currently this implementation does not correctly allow for that

elliottt: Could you add a comment on this, or make an issue to track that we need to handle this better?

JakeChampion: > Could you add a comment on this, or make an issue to track that we need to handle this better?

I've got a second pull request open (in draft right now) which will solve this and fix two other bugs in our async kvstore operations 🙈

PR #752, opened 2024-04-09
 fix: ensure multiple KVStore lookups and deletes can happen at the same time 

JakeChampion:
refactor our async task implementation to be a generic AsyncTask class instead of separate implementations for each async operation

This implementation enables our event loop to no longer have to be aware of all the different async tasks and how to process them, instead it understands AsyncTask and to call the `process` function contained within an AsyncTask.

A side-effect of this refactoring is that KVStore.prototype.lookup and KVStore.prototype.delete now work when being called multiple times in parallel - previously this would not work as the KVStore class contained a slot for the Promise of the async work, which meant if you called either method more than once, the previous method's Promise would have been overwritten within the Slot. The new implementation creates a new AsyncTask instance for each lookup and each delete invocation.

### DIFF
--- a/runtime/js-compute-runtime/builtins/request-response.cpp
+++ b/runtime/js-compute-runtime/builtins/request-response.cpp
@@ -1,11 +1,11 @@
 #include "builtins/request-response.h"
-
 #include "builtins/cache-override.h"
 #include "builtins/client-info.h"
 #include "builtins/fastly.h"
 #include "builtins/fetch-event.h"
 #include "builtins/kv-store.h"
 #include "builtins/native-stream-source.h"
+#include "builtins/shared/dom-exception.h"
 #include "builtins/shared/url.h"
 #include "builtins/transform-stream.h"
 #include "core/encode.h"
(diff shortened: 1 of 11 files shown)


runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -29,8 +29,68 @@
 namespace builtins {
 
 namespace {
+bool error_stream_controller_with_pending_exception(JSContext *cx, JS::HandleObject controller) {
```
JakeChampion, line 32: This used to be within event_loop.cpp but I moved it into request-response as it is only used on requests

runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -29,8 +29,68 @@
...
+  JS_ClearPendingException(cx);
+
+  JS::RootedValueArray<1> args(cx);
+  args[0].set(exn);
+  JS::RootedValue r(cx);
+  return JS::Call(cx, controller, "error", args, &r);
+}
 constexpr size_t HANDLE_READ_CHUNK_SIZE = 8192;
 
+bool process_body_read(JSContext *cx, int32_t handle, JS::HandleObject context,
```
JakeChampion, line 45: This used to be within event_loop.cpp but I moved it into request-response as it is only used on requests

runtime/js-compute-runtime/builtins/request-response.cpp
```diff
@@ -120,6 +180,39 @@ bool enqueue_internal_method(JSContext *cx, JS::HandleObject receiver,
 
 } // namespace
 
+bool RequestOrResponse::process_pending_request(JSContext *cx, int32_t handle,
```
JakeChampion, line 183: This used to be within event_loop.cpp but I moved it into request-response as it is only used on requests

runtime/js-compute-runtime/core/event_loop.cpp
```diff
@@ -256,20 +199,9 @@ bool EventLoop::process_pending_async_tasks(JSContext *cx) {
...
-      } else {
-        handles.push_back(builtins::KVStore::pending_lookup_handle(pending_obj).async_handle());
-      }
-    } else {
-      MOZ_ASSERT(builtins::NativeStreamSource::is_instance(pending_obj));
-      JS::RootedObject owner(cx, builtins::NativeStreamSource::owner(pending_obj));
-      handles.push_back(builtins::RequestOrResponse::body_handle(owner).async_handle());
-    }
+    auto pending_obj = (*pending_async_tasks)[i];
+    fprintf(stderr, "oooh %i\n", AsyncTask::get_handle(pending_obj));
```
elliottt, line 203: Could you remove this printf?

elliottt (approved):
This looks good to me!

It would be nice if we could avoid storing everything in the `AsyncTask` class by keeping two parallel queues for gc managed data and plain old data, but I don't think that needs to block this PR from merging :+1:

PR #756, opened 2024-04-11
fix: correct type definition for the global BackendConfiguration type - there is no `checkCertificate` field

JakeChampion:
This field never existed in the released version of our Backend implementation, it is a remnant of a previous approach I was taking, I somehow forgot to remove the tests and type definition

The tests were all checking to see that the constructor did not throw, which it wouldn't because it has zero knowledge about this made up field...


elliottt (approved):
:fire:

PR #757, opened 2024-04-11
ci: group test failures separately to make it easier to see all the test failures without having to scroll thousands of lines of output

JakeChampion:


elliottt (approved):
Excellent!

PR #763, opened 2024-04-19
Initial StarlingMonkey Build

guybedford:
This implements an initial port of js-compute-runtime to be built on top of StarlingMonkey. The new runtime build is entirely self-contained in `runtime/fastly` and the build scripts are located there, also runnable via `npm run build:starlingmonkey:debug` and `npm run build:starlingmonkey`.

In the CLI a new `--starlingmonkey` flag is added, but not documented, and the StarlingMonkey build is not included in the release.

In testing, we branch both the main `index.js` file into a StarlingMonkey-specific `index-starlingmonkey.js` in the integraiton tests and also the `tests.json` file into a `tests-starlingmonkey.json`. By extending the lists in these files, to eventually include all tests, we should eventually be able to hit parity with the main test suite and then remove these changes.

The main `test.js` now takes a `--starlingmonkey` flag along with the former `--local` flag, which then performs the branching into the above and also updates the build script to use the starlingmonkey build.

In CI, a new StarlingMonkey workflow will run the StarlingMonkey debug build and tests, using the above new flag.

Current test status is just the `btoa` test passing. But this is a good start from which in theory it is just a matter of extending support and coverage.


guybedford:
Moved to https://github.com/fastly/js-compute-runtime/pull/764.

PR #764, opened 2024-04-23
Initial StarlingMonkey Build

guybedford:
This implements an initial port of js-compute-runtime to be built on top of StarlingMonkey. The new runtime build is entirely self-contained in `runtime/fastly` and the build scripts are located there, also runnable via `npm run build:starlingmonkey:debug` and `npm run build:starlingmonkey`.

Most existing StarlingMonkey builtins are used, apart from the `fetch-event` and `fetch` builtins which are directly ported from Fastly's existing implementation to ensure exact implementation consistency.

The Fastly host-api is integrated into StarlingMonkey, along with the existing component model support in that integration so that a StarlingMonkey component can be easily enabled in future similarly to the current runtime.

In the CLI, a new `--starlingmonkey` flag is added, but not documented, and the StarlingMonkey build is not included in the release.

In testing, some changes were necessary to the test files to support the mocked builtins currently implemented in StarlingMonkey. Then a new `tests-starlingmonkey.json` JSON file is added that is an array of the test keys that are tested for StarlingMonkey. By extending this list to include every test key in `tests.json`, we should eventually be able to hit parity with the main test suite and then be able to remove this filter.

The main `test.js` now takes a `--starlingmonkey` flag along with the former `--local` flag, which then performs the branching into the above and also updates the build script to use the starlingmonkey build.

In CI, a new StarlingMonkey workflow will run the StarlingMonkey debug build and tests, using the above new flag.

**Integration Test status: 159 / 727 = 21.9% passing.**


integration-tests/js-compute/fixtures/app/src/btoa.js
```diff
@@ -737,7 +737,7 @@ routes.set('/btoa', () => {
     if (error) { return error; }
     error = assertThrows(() => atob("--"))
     if (error) { return error; }
-    error = assertThrows(() => atob("__"))
+    error = assertThrows(() => atob("--"))
```
JakeChampion, line 740: This is a duplicate of the above assertion I think

guybedford: This is actually a StarlingMonkey bug can you believe it....

guybedford: I've split this test out into it's own route to isolate this as a separate upstream issue.

src/parseInputs.js
```diff
@@ -21,8 +22,18 @@ export async function parseInputs(cliInputs) {
...
+      noStarlingMonkeyComponent();
+    }
     wasmEngine = join(__dirname, "../js-compute-runtime-component.wasm");
   };
+  let useStarlingMonkey = () => {
+    starlingMonkey = true;
+    if (component) {
+      noStarlingMonkeyComponent();
+    }
+    wasmEngine = wasmEngine = join(__dirname, "../starling.wasm");
```
elliottt, line 35: Could we check for the existence of the starling monkey wasm, and give a more specific error message if it's not present?

runtime/fastly/builtins/fetch/fetch.h
```diff
@@ -0,0 +1,8 @@
+#include "headers.h"
+#include "request-response.h"
+
+namespace fastly::fetch {
+namespace {
+api::Engine *ENGINE;
```
elliottt, line 6: Could you mark this `extern` so that we don't end up with multiple definitions of `ENGINE`?

runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -0,0 +1,655 @@
...
+    if (!Fastly::baseURL)
+      return false;
+  }
+
+  return true;
+}
+
+// bool FetchEvent::init_incoming_request(JSContext *cx, JS::HandleObject self,
+//                                        host_api::HttpIncomingRequest *req) {
+//   JS::RootedObject request(
```
elliottt, line 200: Is this in-progress, or can we remove the commented-out code?

runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -0,0 +1,655 @@
...
+    return false;
+  }
+
+  // // Step 3
+  if (state(self) != State::unhandled) {
+    JS_ReportErrorUTF8(cx, "FetchEvent#respondWith can't be called twice on the same event");
+    return false;
+  }
+
+  // // Step 4
```
elliottt, line 390: There are a few cases where the implementation differs from the original in `runtime/js-compute-runtime/builtins/fetch-event.cpp` in comments only. Could we clean up these comments so that they match a bit closer?

runtime/fastly/builtins/fastly.h
```diff
@@ -0,0 +1,65 @@
...
+  static JS::PersistentRooted<JSObject *> baseURL;
+  static JS::PersistentRooted<JSString *> defaultBackend;
+  static bool allowDynamicBackends;
+
+  static const JSPropertySpec properties[];
+
+  // static bool createFanoutHandoff(JSContext *cx, unsigned argc, JS::Value *vp);
+  // static bool now(JSContext *cx, unsigned argc, JS::Value *vp);
+  // static bool dump(JSContext *cx, unsigned argc, JS::Value *vp);
+  // static bool enableDebugLogging(JSContext *cx, unsigned argc, JS::Value *vp);
```
elliottt, line 50: Could we get some TODOs on the commented-out code to make it obvious that it's part of an in-progress migration?

elliottt (commented):
I checked the diff for files between the `runtime/fastly` and `runtime/js-compute-runtime` dirs when I could find the originals. Overall this looks fine, but I would like two things before approving:

1. The commented-out code in all of the new modules in `runtime/fastly` should be either removed, or given explicit `TODO` comments indicating that they're part of an in-progress port
2. Let's make sure that if someone uses the `--starlingmonkey` flag on a released version of the compute sdk that they get either an error that indicates the flag doesn't exist, or a really clear explanation about how starling monkey isn't yet available.

I'd also like if you could make it really explicit in the existing build/release workflow that we do not include starling monkey in any sdk releases right now. That will make it easier to have an in-tree WIP for the port, but also make it really obvious where we're going to need to make changes in the future once we're looking to switch.

runtime/fastly/builtins/fetch/fetch.h
guybedford: This is supposed to be internal and duplicated - following the pattern from https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/fetch/fetch_event.cpp#L18.

If there's a better pattern here perhaps we can get alignment with StarlingMonkey first? All builtins need to maintain the engine in a builtin-local variable under the current hook-based extension model.

src/parseInputs.js
guybedford: Added.

guybedford:
Thanks for this review, I'll work through the commented out sections and post back when that's ready.

> I'd also like if you could make it really explicit in the existing build/release workflow that we do not include starling monkey in any sdk releases right now. That will make it easier to have an in-tree WIP for the port, but also make it really obvious where we're going to need to make changes in the future once we're looking to switch.

The GitHub workflow for StarlingMonkey is a completely separate workflow. As for ensuring we never inadvertantly expose it to users, I've explicitly added `"starling.wasm"` to the package.json `ignore` so that there is no risk of mistakenly publishing it to npm.

The cross-tree points remaining as mentioned are the CLI and test integration points, but otherwise we have comprehensive separation. I'm unsure what more we can do here, but am open to suggestions further.

runtime/fastly/builtins/fetch/fetch.h
elliottt: When defined in the `.cpp` file, that will be a constant. When defined in the header, the definition will introduce a new `ENGINE` constant for every include. I think this is probably unintentional, so I would mark this declaration `extern`, and put the real definition in `fetch.cpp`.

guybedford: The problem I was trying to solve here, is that this is shared between all the `fetch/*.cpp` files, as a kind of local state, where the fetch builtin is treated as one builtin across many files. I was under the impression that since the anonymous namespace is under the `fastly::fetch` builtin namespace we get that sharing without spilling it to other builtins in other builtin namespaces.

runtime/fastly/builtins/fetch-event.cpp
guybedford: I'ce gone ahead and removed this.

runtime/fastly/builtins/fastly.h
guybedford: Added TODOs to all code that should be uncommented.

guybedford:
CI is finally green here. If land and iterate works for all here, I'm happy to go with this dev process and to start on the next PR. Let me know if there's any further feedback.

runtime/fastly/builtins/fastly.cpp
```diff
@@ -0,0 +1,461 @@
...
+  if (errorNumber > 0 && errorNumber < JSErrNum_Limit) {
+    return &fastly_ErrorFormatString[errorNumber];
+  }
+
+  return nullptr;
+}
+
+namespace {
+
+api::Engine *ENGINE;
```
elliottt, line 36: Here's another `ENGINE` reference that doesn't originate in the `fetch.h` declaration. Are all the `ENGINE` pointers meant to be to the same value, or should they be different?

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "encode.h"
+#include "extension-api.h"
+
+using fastly::backend::Backend;
+using fastly::fastly::Fastly;
+using fastly::fetch::Request;
+
+namespace fastly::fetch {
+
+api::Engine *ENGINE;
```
elliottt, line 16: If all the other anonymous declaration of `ENGINE` are meant to point to this, it might be better to move this to a different header, and keep the declaration in one `cpp` file. If it's intentional that they're all different, then perhaps this one should move out of the header.

elliottt (approved):
I think this looks good, thank you for taking steps to isolate it from the existing runtime build!

It looks like there are still a few different declarations of `ENGINE` throughout the `runtime/fastly` tree. I'd be happy to help out with putting them all in a single place, if they're all supposed to be the same.

runtime/fastly/builtins/fetch/fetch.cpp
guybedford: I think if we want to change this extension model, we should address that upstream in StarlingMonkey rather.

PR #767, opened 2024-05-06
CacheOverride StarlingMonkey port

guybedford:
This ports the cache-override builtin to StarlingMonkey, fixing the TODOs on this subsystem and enabling the tests.

A regression in the ehe error lookup function is also fixed.

**Integration Test status: 245 / 726 = 34% passing.**


integration-tests/js-compute/fixtures/app/src/btoa.js
```diff
@@ -737,16 +737,9 @@ routes.set('/btoa', () => {
     if (error) { return error; }
     error = assertThrows(() => atob("--"))
     if (error) { return error; }
+    error = assertThrows(() => atob("__"))
+    if (error) { return error; }
```
guybedford, line 741: This was fixed upstream in StarlingMonkey, so no longer needs separating.

integration-tests/js-compute/fixtures/app/src/dynamic-backend.js
```diff
@@ -1527,7 +1522,7 @@ routes.set("/backend/timeout", async () => {
     routes.set("/backend/health/called-as-constructor-function", async () => {
       let error = assertThrows(() => {
         new Backend.health()
-      }, TypeError, `Backend.health is not a constructor`)
+      }, TypeError)
```
guybedford, line 1525: This is now `(intermediate value).health is not a constructor` (but still pointing to the line number source of the error in the original source, so the stack retains the context). This is due to the ad-hoc nature of the bytecode name tracing in https://searchfox.org/mozilla-central/source/js/src/vm/BytecodeUtil.cpp#1590, which doesn't carry through to the expression naming for the builtin module. The same thing happens for arbitrary JS expressions though anyway (eg `new getObj().health()`).

Since `new Backend.health()` is a pretty extreme edge case, as most JS developers won't attempt to call a function as a class constructor, so I don't think this error message change would be considered breaking.

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -11,6 +11,8 @@ using fastly::backend::Backend;
 using fastly::fastly::Fastly;
 using fastly::fetch::Request;
 
+using fastly::fastly::FastlyGetErrorMessage;
```
elliottt, line 14: Why did this namespace end up as `fastly::fastly`?

elliottt (approved):
This looks good, nice work!

runtime/fastly/builtins/fetch/fetch.cpp
guybedford: This is basically `fastly::core` since each builtin has its own name under the `fastly::` namespace. We could do a rename to `fastly::core` in a follow-up.

Issue #579, opened 2023-07-10
Feature request: SubtleCrypto support for RSA-PSS, ECDSA, ED25519, pkcs8 and spki


JakeChampion:
If I could lobby for a few more algorithms to be supported 🙏 

- `RSA-PSS`
- `ECDSA` (at least P-256, but ideally all three curves)
- `Ed25519` from https://wicg.github.io/webcrypto-secure-curves/
- almost forgot, the goto JWS algorithm - `HMAC`

Also `pkcs8` and `spki` key format import/export are not to be overlooked. Developers LOOOVE to use in favour of JWKs.

I have a comprehensive [test suite](https://github.com/panva/jose#supported-runtimes) for different runtimes' webcrypto implementations that I could unload on js-compute-runtime eventually.

_Originally posted by @panva in https://github.com/fastly/js-compute-runtime/issues/440#issuecomment-1535339296_

jfctfl:
Would this also include `RS256`? 🙂 We need `RS256` to be able to verify JWTs from Fanout in a c@e js runtime - for more context: https://developer.fastly.com/learning/concepts/real-time-messaging/fanout/#validating-grip-requests

JakeChampion:
@jfctfl We support RS256, but that is not what Fanout uses, Fanout uses EC P-256, which we do not support and is already listed in this issue 👍

guybedford:
Moved to https://github.com/bytecodealliance/StarlingMonkey/issues/123.

PR #769, opened 2024-05-07
Add EcKeyImportParams to SubtleCrypto.importKey

jsocol:
In #639 we added support for ECDSA keys in SubtleCrypto. This adds the
interface to the type defintions for SubtleCrypto.importKey to allow
consuming TypeScript applications to use this type of key. Partly
addresses #579 and brings the library in sync with [our Fanout docs][1], and the library's [reference docs][2].

[1]: https://www.fastly.com/documentation/guides/concepts/real-time-messaging/fanout/#validating-grip-requests
[2]: https://js-compute-reference-docs.edgecompute.app/docs/globals/SubtleCrypto/prototype/importKey#parameters

### DIFF
--- a/types/globals.d.ts
+++ b/types/globals.d.ts
@@ -1705,14 +1705,15 @@ declare class SubtleCrypto {
   // generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
   // generateKey(algorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKeyPair | CryptoKey>;
   // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
-  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
+  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
+  // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
   importKey(
     format: Exclude<KeyFormat, "jwk">,
     keyData: BufferSource,
     algorithm:
       AlgorithmIdentifier
       | RsaHashedImportParams
-      // | EcKeyImportParams
+      | EcKeyImportParams
       | HmacImportParams
       // | AesKeyAlgorithm
     , extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKey>;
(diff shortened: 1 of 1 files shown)


types/globals.d.ts
```diff
@@ -1705,14 +1705,15 @@ declare class SubtleCrypto {
   // generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
   // generateKey(algorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKeyPair | CryptoKey>;
   // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
-  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
+  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
+  // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
```
guybedford, line 1709: Do we need this line or can we just remove it?

types/globals.d.ts
```diff
@@ -1705,14 +1705,15 @@ declare class SubtleCrypto {
   // generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
   // generateKey(algorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKeyPair | CryptoKey>;
   // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
-  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
+  importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
```
guybedford, line 1708: I'm not sure we implemented `Hmac` for `jwk`? Do you have a reference for this? Otherwise let's just add `EcKeyImportParams` for now?

types/globals.d.ts
```diff
@@ -1705,14 +1705,15 @@ declare class SubtleCrypto {
...
+  // importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
   importKey(
     format: Exclude<KeyFormat, "jwk">,
     keyData: BufferSource,
     algorithm:
       AlgorithmIdentifier
       | RsaHashedImportParams
-      // | EcKeyImportParams
+      | EcKeyImportParams
```
guybedford, line 1716: It sounds like in https://github.com/fastly/js-compute-runtime/pull/639 it only implemented EcKey for the jwk subset, so that it shouldn't be added here as well, so this diff should probably be excluded? Unless you have a reference for non-jwk key support for ECDSA?

guybedford (commented):
Thanks for spotting this, I'm still picking up context here in understanding what we do and don't implement, so have more questions than answers right now.

I'm just tending towards wanting to land the most conservative change so we don't advertise features we don't support, until we can verify that we absolutely do support further cases.

guybedford (approved):

guybedford:
That's due to the remote branch on the PR, we're good to merge though.

Issue #768, opened 2024-05-07
bug: Setting cacheKey on Request constructor or during fetch does not work


harmony7:
In #281, Request.prototype.setCacheKey() was implemented, which sets the cache key for an outgoing request.

```javascript
const req = new Request("https://example.com/path/to/resource");
req.setCacheKey("foo");
const res = await fetch(req, { backend: "example" });
```

The above example works as expected.
(Header `Fastly-Xqd-Cache-Key: 2C26B46B68FFC68FF99B453C1D30413413422D706483BFA0F98A5E886266E7AE` is present.)

However, setting the cache key directly on the Request or during the fetch call does not work:
```javascript
const req = new Request("https://example.com/path/to/resource", { cacheKey: "foo" }); // Adding cacheKey here
const res = await fetch(req, { backend: "example" });
```

```javascript
const req = new Request("https://example.com/path/to/resource");
const res = await fetch(req, { backend: "example", cacheKey: "foo" }); // Adding cacheKey here
```

(No header `Fastly-Xqd-Cache-Key` is present)

harmony7:
I compared this to `cacheOverride`, which does work.

`setCacheOverride` works by setting the Slot `CacheOverride` on the request object:

```c++
bool Request::set_cache_override(JSContext *cx, JS::HandleObject self,
                                 JS::HandleValue cache_override_val) {
  /* ... */
  JS::SetReservedSlot(self, static_cast<uint32_t>(Slots::CacheOverride),
                      JS::ObjectValue(*override));
  return true;
}
```

and then by applying it during a fetch
```c++
bool fetch(JSContext *cx, unsigned argc, Value *vp) {
  CallArgs args = CallArgsFromVp(argc, vp);
  /* ... */
  if (!builtins::Request::apply_cache_override(cx, request)) {
    return false;
  }
  /* ... */
}
```

This means that cache_override can come from either the Request constructor or the fetch parameter, or be set by the setCacheKey, and they would all be applied the same.

Whereas `setCacheKey` goes ahead and applies the `cacheKey` value directly:
```javascript
bool Request::set_cache_key(JSContext *cx, JS::HandleObject self, JS::HandleValue cache_key_val) {
  /* ... */
  JSObject *headers = RequestOrResponse::headers<builtins::Headers::Mode::ProxyToRequest>(cx, self);
  /* ... */
  JS::RootedObject headers_val(cx, headers);
[...]

PR #770, opened 2024-05-07
fix: support cacheKey in Request init

guybedford:
Adds support for `cacheKey` in `Request` initialization, which was previously omitted, and fixes the test case that was supposed to test for this.

We don't use an internal slot for `cacheKey` since it is effectively a header-manipulating function, where that header manipulation is observable to users. If we want to use an internal slot, that would be a breaking change at this point to provide a header alteration from such a slot that exists only as observable to the host.

Note that this is implemented in the parallel StarlingMonkey branch while this port is in progress... a fact of life during this port process!

Resolves https://github.com/fastly/js-compute-runtime/issues/768.


elliottt (approved):
Looks great!

PR #772, opened 2024-05-08
Simple cache StarlingMonkey port

guybedford:
Ports the simple cache subsystem. This does not yet port core cache so the `fastly:cache` namespace remains incomplete.

A CI matrix is added for testing both `varnish` and `compute` platforms for StarlingMonkey, so that this can be properly tested.

As with the cache override PR, the "SimpleCache.x is not a constructor" error messages become `(intermediate value).x is not a constructor` errors due to the nature of the ESM builtins, but will still point to the correct JS source position for debugging.

This uses StarlingMonkey from https://github.com/bytecodealliance/StarlingMonkey/pull/43, thanks to @lpereira for helping figure out the OpenSSL CMake dependency here.

This is also based to https://github.com/fastly/js-compute-runtime/pull/776 to allow the StarlingMonkey e2e tests to run simultaneously with the runtime ones, by using distinct kv store names.


runtime/fastly/builtins/fastly.cpp
```diff
@@ -463,8 +494,59 @@ bool install(api::Engine *engine) {
...
+// We currently support five types of body inputs:
+// - byte sequence
+// - buffer source
+// - USV strings
+// - URLSearchParams
+// After the other other options are checked explicitly, all other inputs are
+// encoded to a UTF8 string to be treated as a USV string.
+// TODO: Support the other possible inputs to Body.
+JS::Result<std::tuple<JS::UniqueChars, size_t>> convertBodyInit(JSContext *cx,
+                                                                JS::HandleValue bodyInit) {
```
elliottt, line 510: Is there an equivalent of `js-compute-builtins.cpp` in the `fastly` tree? I remember this function being needed in a bunch of other places, so it might be worth moving it in the future.

runtime/fastly/builtins/cache-simple.cpp
```diff
@@ -0,0 +1,786 @@
+#include "cache-simple.h"
```
elliottt, line 1: There were pretty minor differences between the existing simple-cache.cpp, nice!

elliottt (approved):
This looks great, nice work!

runtime/fastly/builtins/fastly.cpp
guybedford: A lot of what was there previously ended up going into StarlingMonkey as separate builtin extensions / in the core. I think in general having helper functions associated with their spec builtins would be nicer than maintaining the generic common when possible.

runtime/fastly/builtins/cache-simple.cpp
guybedford: Yeah the process is fairly straightforward at this point, will try to get a few similar subsystems together for the next batch.

Issue #775, opened 2024-05-08
Support for checking the runtime version


guybedford:
It would be useful to have support for a `fastly.version` string which returns the runtime version of the runtime engine build itself.

Specifically this is useful for the StarlingMonkey builds to be able to say it is a `starlingmonkey-dev` tag on these builds.

PR #776, opened 2024-05-08
feat: fastly.sdkVersion implementation

guybedford:
This implements a new `fastly.sdkVersion` property as well as an equivalent `import { sdkVersion } from 'fastly:experimental'` version property.

This is immediately useful in the StarlingMonkey builds to be able to easily inspect if we are in a StarlingMonkey build by checking `fastly.sdkVersion === 'starlingmonkey-dev'`, which allows smoothing over any test handling edge cases more easily.

It's included in the types as a hidden property that is experimental. The engine build for the main runtime is set to the next patch version with a `-dev` suffix.

We will need to ensure we update this definition for each version release.

Resolves https://github.com/fastly/js-compute-runtime/issues/775.


guybedford:
@elliottt as discussed today, I've renamed this to `sdkVersion`. Ready for review.

runtime/js-compute-runtime/js-compute-builtins.h
```diff
@@ -32,6 +32,8 @@ const JSErrorFormatString js_ErrorFormatString[JSErrNum_Limit] = {
 
 #include "host_interface/host_api.h"
 
+#define RUNTIME_VERSION "3.13.2-dev"
```
elliottt, line 35: We talked about this briefly earlier, but it would be great to figure out how to get `release-please` to keep this in sync.

elliottt (approved):
Looks great to me!

runtime/js-compute-runtime/js-compute-builtins.h
guybedford: Will follow up as soon as I've worked out the release process.

PR #779, opened 2024-05-14
StarlingMonkey port of kv-store, client info

guybedford:
Ports the kv-store and client info subsystems.


runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -24,6 +29,283 @@ namespace fastly::fetch_event {
 
 namespace {
 
+JSString *address(JSObject *obj) {
+  JS::Value val = JS::GetReservedSlot(obj, static_cast<uint32_t>(ClientInfo::Slots::Address));
+  return val.isString() ? val.toString() : nullptr;
+}
```
elliottt, line 35: Why inline the ClientInfo implementation instead of copying over individual files from the existing builtin?

runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -24,6 +29,283 @@ namespace fastly::fetch_event {
...
+    JS::SetReservedSlot(self, static_cast<uint32_t>(ClientInfo::Slots::Cipher),
+                        JS::StringValue(result));
+  }
+
+  args.rval().setString(result);
+  return true;
+}
+
+bool ClientInfo::tls_ja3_md5_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  fprintf(stderr, "tlsJA3MD5");
```
elliottt, line 176:
```suggestion
```

runtime/fastly/host-api/host_api_fastly.h
```diff
@@ -107,11 +108,33 @@ class FastlyAsyncTask final : public AsyncTask {
       abort();
     handle_ = static_cast<int32_t>(handle);
   }
+  FastlyAsyncTask(uint32_t handle, JSContext *cx, JS::HandleObject context,
+                  JS::HandleObject promise, ProcessAsyncTask *process) {
+    if (static_cast<int32_t>(handle) < 0)
```
elliottt, line 113: Shouldn't this be a comparison against the `INVALID_HANDLE` constant? The conversion to a signed type is a little concerning, and I'd be more comfortable with a direct comparison against the known invalid handle value (if that's what this conditional is implementing).

elliottt (approved):
This looks good! I just had a few questions that don't really block merging.

runtime/fastly/builtins/fetch-event.cpp
guybedford: The argument here is that clientInfo isn't its own builtin as it is part of the "fastly fetch event" at this point.

Perhaps in future we could refactor our fetch event to be shared with StarlingMonkey in future, and treat the Fastly-specific stuff as monkey patching on top, but even then treating each feature as its own builtin may likely be overkill, and instead we may still have a single "fastly" builtin that does the Fastly patches.

runtime/fastly/builtins/fetch-event.cpp
guybedford: Well-spotted, thanks!

runtime/fastly/host-api/host_api_fastly.h
guybedford: The StarlingMonkey handles are signed ints https://github.com/bytecodealliance/StarlingMonkey/blob/main/include/extension-api.h#L27, so since we are casting our Fastly handles, it seemed like asserting we never go into the signed range would be the stronger assertion. We could just make this an invalid handle assertion though, yes.

elliottt: I think it's reasonable to check that it's always greater than zero, but it might violate assumptions that compute makes about the valid range of handles. I'm reasonably sure we'll never see handles that have the top-bit set, but it would be easiest to assume that we only care about the invalid handle sentinel.

PR #780, opened 2024-05-16
changelog: version links, disable checker for version links

guybedford:
The release is currently failing because the CHANGELOG generation is outputting the version headings as URLs.

Release please seems to default to outputting the version headings as URLs now which seems useful as well.

This PR updates the existing CHANGELOG to use version heading URLs and also disables the changelog checker based on the assumption that we have enough evidence at this point to trust the output from release please.

In addition, the changelog checker doesn't seem to have an option to disable version URL checks.

For now this leaves the changelog checker in the repo in case we want to change this back in future, if it can support this or if we find a release please option to disable the version URL links (which I was unable to), but we could remove it entirely as well.

This PR should fix the current release at least.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ env:
 jobs:
 
   check-changelog:
-    if: github.ref != 'refs/heads/main'
+    if: false && github.ref != 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 # Changelog
 
 
-## 3.13.1 (2024-04-12)
+## [3.13.1](https://github.com/fastly/js-compute-runtime/compare/v3.13.0...v3.13.1) (2024-04-12)
 
 
 ### Fixed
 
 * remove debugging message which got commited ([4219a0a](https://github.com/fastly/js-compute-runtime/commit/4219a0ac87d68d9a9fc57aaea43994a867f5dd0e))
 
-## 3.13.0 (2024-04-11)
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ env:
 jobs:
 
   check-changelog:
-    if: github.ref != 'refs/heads/main'
+    if: false && github.ref != 'refs/heads/main'
```
elliottt, line 20: Was this for debugging, or is it necessary for the fix to the release process?

elliottt (approved):
Nice!

.github/workflows/main.yml
guybedford: Yeah the checker itself is no longer in sync with the output. And the checker doesn't seem to expose options to change that - https://www.npmjs.com/package/cae-release-notes-format-checker.

This at least allows us to keep the infrastructure so that if @JakeChampion updates the checker we can then reenable it.

JakeChampion:
The checker ensures that the format is correct for consumption by the developer portal, the release-please format is not compatible unfortunately, I manually edit the changelog before each release to pass the checker, otherwise the RSS feed on developer portal will not get updates about our releases. Cc @doramatadora @kpfleming

guybedford:
I did a revert here but unfortunately:

1. This manual change is not documented anywhere in process
2. Release please will not let me make manual changes to the release PR
3. Manually creating a tag does not result in the release workflow running

I simply do not know how to create a release at this point.

The failed 3.14.2 release is not published, and sitting now on main. I will resolve this on Tuesday when I get back from holiday, unless someone else wants to look at it first or knows how to fix this further.

guybedford:
To update on this:

* The issue I was hitting was that the documentation action was force pushing over my changes, when I attempted to manually update the release
* The new update to the documentation action in https://github.com/fastly/js-compute-runtime/pull/789 will now automatically update the release
* Manual editing can be applied, it must just be done after this action completes and not during to avoid racing

PR #790, opened 2024-05-22
Port remaining StarlingMonkey builtins

guybedford:
This ports the remaining js-compute-runtime builtins to the StarlingMonkey port.

Test status with this PR gets to 716 / 729 tests passing (98%).


runtime/fastly/builtins/device.h
```diff
@@ -0,0 +1,62 @@
+#ifndef FASTLY_DEVICE_H
+#define FASTLY_DEVICE_H
+
+#include "builtin.h"
+#include "extension-api.h"
+
+using builtins::BuiltinImpl;
```
elliottt, line 7: It's good to keep `using` declarations out of headers, as they change the namespace everywhere they're included. Could you move this into the cpp file, and use the fully qualified name in the header?

elliottt (approved):
This looks great! My only suggestion would be that `using` declarations that manipulate the top-level namespace should be kept out of header files, as they will have an effect on whatever context they're included in. It's more verbose but ultimately easier to deal with if you restrict use of `using` to `.cpp` files only.

runtime/fastly/builtins/device.h
guybedford: Sure, added.

Issue #783, opened 2024-05-17
Dynamic backends default timeouts


guybedford:
When enabling dynamic backends, it would be nice to be able to also provide default values for the timeouts to be associated with the dynamic backends.

Perhaps we can extend the API of the form:

```js
allowDynamicBackends({
    connectTimeout?: number;
    firstByteTimeout?: number;
    betweenBytesTimeout?: number;
});
```

PR #792, opened 2024-05-23
feat: support default timeout configurations for dynamic backends

guybedford:
This implements https://github.com/fastly/js-compute-runtime/issues/783, supporting an object variant of the `allowDynamicBackends` call taking the default timeout configurations.

No other dynamic backend configurations are currently supported in this, but it could be extended to support further options in due course.

### DIFF
--- a/runtime/fastly/builtins/fastly.cpp
+++ b/runtime/fastly/builtins/fastly.cpp
@@ -43,6 +43,9 @@ JS::PersistentRooted<JSObject *> Fastly::env;
 JS::PersistentRooted<JSObject *> Fastly::baseURL;
 JS::PersistentRooted<JSString *> Fastly::defaultBackend;
 bool Fastly::allowDynamicBackends = false;
+host_api::BackendConfig Fastly::defaultDynamicBackendConfig = host_api::BackendConfig{
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt,
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt};
 
 bool Fastly::dump(JSContext *cx, unsigned argc, JS::Value *vp) {
   JS::CallArgs args = CallArgsFromVp(argc, vp);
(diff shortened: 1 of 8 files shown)


runtime/js-compute-runtime/builtins/fastly.cpp
```diff
@@ -26,6 +26,9 @@ JS::PersistentRooted<JSObject *> Fastly::env;
 JS::PersistentRooted<JSObject *> Fastly::baseURL;
 JS::PersistentRooted<JSString *> Fastly::defaultBackend;
 bool Fastly::allowDynamicBackends = false;
+host_api::BackendConfig Fastly::defaultDynamicBackendConfig = host_api::BackendConfig{
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt,
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt};
```
elliottt, line 31: All those fields will default-initialize with nullopt, so you can omit them.
```suggestion
host_api::BackendConfig Fastly::defaultDynamicBackendConfig;
```

runtime/js-compute-runtime/builtins/backend.cpp
```diff
@@ -1120,6 +1121,31 @@ JSObject *Backend::create(JSContext *cx, JS::HandleObject request) {
...
   JS::SetReservedSlot(backend, Backend::Slots::DontPool, JS::BooleanValue(false));
 
+  if (Fastly::defaultDynamicBackendConfig.connect_timeout.has_value()) {
+    JS::RootedValue connect_timeout_val(
+        cx, JS::NumberValue(Fastly::defaultDynamicBackendConfig.connect_timeout.value()));
+    if (!Backend::set_timeout_slot(cx, backend, connect_timeout_val, Backend::Slots::ConnectTimeout,
+                                   "connectTimeout")) {
+      return nullptr;
+    }
+  }
```
elliottt, line 1131: Why does this unconditionally set the values if the defaults are present? Will they be overridden by user code later?

runtime/fastly/builtins/fastly.cpp
```diff
@@ -43,6 +43,9 @@ JS::PersistentRooted<JSObject *> Fastly::env;
 JS::PersistentRooted<JSObject *> Fastly::baseURL;
 JS::PersistentRooted<JSString *> Fastly::defaultBackend;
 bool Fastly::allowDynamicBackends = false;
+host_api::BackendConfig Fastly::defaultDynamicBackendConfig = host_api::BackendConfig{
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt,
+    std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt, std::nullopt};
```
elliottt, line 48: Same comment here, you don't need to explicitly initialize these.

elliottt (approved):
I think this makes sense, but I had one question about how the defaults are used. Happy for you to merge if I misunderstood what's going on.

runtime/fastly/builtins/fastly.cpp
guybedford:
```suggestion
host_api::BackendConfig Fastly::defaultDynamicBackendConfig;
```

runtime/js-compute-runtime/builtins/backend.cpp
guybedford: This function creates a dynamic backend for a request's host, and this function doesn't currently support setting timeouts at all, since timeouts are only currently supported for explicitly created backends that are passed in.

The default timeouts are therefore always applied when present, providing a way to set them for this use case.

PR #795, opened 2024-05-28
StarlingMonkey port body streaming and event loop lifetime fixes

guybedford:
This fixes the fetch event body streaming and lifetime handling for the StarlingMonkey port. Lifetimes are handled properly now by carefully ensuring that event loop interest increments and decrements directly correspond to being equivalent to the fetch_event::is_active state.


elliottt (approved):
Awesome!

Issue #794, opened 2024-05-28
Support mTLS for dynamic backends


kphillips:
Given that the Fastly Rust SDK supports mutual TLS for dynamic backends via [BackendBuilder::provide_client_certificate()](https://docs.rs/fastly/latest/fastly/backend/struct.BackendBuilder.html#method.provide_client_certificate), the Fastly Compute platform is able to support mTLS backend connections.

Please extend this functionality to the JavaScript Compute SDK.

guybedford:
Yes, we can add support for this in the JavaScript compute runtime, it shouldn't be too tricky.

In terms of the API, it can be as simple as adding a new `clientCert` field to the [BackendConfiguration type](https://github.com/fastly/js-compute-runtime/blob/main/types/backend.d.ts#L2), of the form:

```ts
interface ClientCert {
  cert: String,
  key: SecretStoreEntry
}
```

Then we would also need to implement `SecretStore.fromBytes` of the form:

```ts
class SecretStore {
  static fromBytes (bytes: Uint8Array) -> SecretStoreEntry;
}
```

I will aim to implement and land this by early next week, to go out with the next release already adding features to dynamic backends as well.

guybedford:
Support landed in https://github.com/fastly/js-compute-runtime/commit/7d2b7b781ed808d9bcf1fe9584aa31f788b980a2, I'll get a release cycle going now.

PR #796, opened 2024-06-01
feat: dynamic backends clientCertificate with SecretStore fromBytes, rawbytes

guybedford:
Implements https://github.com/fastly/js-compute-runtime/issues/794. API as per `types/backend.d.ts`.

SecretStore is extended to support `SecretStore.fromBytes` and `SecretStoreEntry.prototype.rawbytes()` to get the raw buffer for a secret instead of the UTF8 encoding.

The new client cert option is then implemented for dynamic backends taking an object with a `certificate` string and `key` `SecretStoreEntry` properties, with error handling.

This PR also updates to the latest wit-bindgen, regenerating the world, and reenables the secret store tests which got dropped in the refactoring PR in https://github.com/fastly/js-compute-runtime/pull/673.


runtime/fastly/builtins/backend.cpp
```diff
@@ -1441,16 +1479,57 @@ bool Backend::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+    if (!JS_HasProperty(cx, client_cert_obj, "certificate", &found)) {
       return false;
     }
-    if (JS_GetStringLength(sni_hostname) == 0) {
+    if (!found) {
       JS_ReportErrorNumberASCII(cx, FastlyGetErrorMessage, nullptr,
-                                JSMSG_BACKEND_SNI_HOSTNAME_EMPTY);
+                                JSMSG_BACKEND_CLIENT_CERTIFICATE_NO_CERTIFICATE);
+      return false;
+    }
```
elliottt, line 1510: Isn't this dead code, as the `if (!found)` case is underneath the `if (found)` above on line 1491? Perhaps this should be reverted back to checking for an empty string, as that case would be distinct from the string being a `nullptr`.

runtime/fastly/builtins/secret-store.cpp
```diff
@@ -46,7 +81,8 @@ const JSPropertySpec SecretStoreEntry::static_properties[] = {
 };
 
 const JSFunctionSpec SecretStoreEntry::methods[] = {
-    JS_FN("plaintext", plaintext, 0, JSPROP_ENUMERATE), JS_FS_END};
+    JS_FN("plaintext", plaintext, 0, JSPROP_ENUMERATE),
+    JS_FN("rawbytes", rawbytes, 0, JSPROP_ENUMERATE), JS_FS_END};
```
elliottt, line 85: We have `fromBytes` and `rawbytes`: would it be reasonable to match the camel-case of `fromBytes`, or is there a js convention for using `rawbytes` for these sorts of methods?

runtime/fastly/host-api/component/fastly_world_adapter.cpp
```diff
@@ -24,6 +24,7 @@ static_assert(URI_MAX_LEN < HOSTCALL_BUFFER_LEN);
 static bool convert_result(int res, fastly_compute_at_edge_types_error_t *err) {
   if (res == 0)
     return true;
+  fprintf(stderr, "HOST ERROR %d", res);
```
elliottt, line 27: Should this be removed?

elliottt (commented):
This all looks good to me! Just a few questions before merging.

runtime/fastly/builtins/backend.cpp
guybedford: I believe not since `found` is reused on the stack throughout this routine.

runtime/fastly/builtins/secret-store.cpp
guybedford: My thinking here was to follow in the `plaintext` convention which isn't camel case either. In the name of acting like this API shape wasn't a mistake and was always planned :P We can certainly switch to `rawBytes` too.

runtime/fastly/builtins/backend.cpp
```diff
@@ -801,6 +803,17 @@ JS::Result<mozilla::Ok> Backend::register_dynamic_backend(JSContext *cx, JS::Han
     backend_config.sni_hostname.emplace(std::move(sni_hostname_chars));
   }
 
+  auto client_cert_slot = JS::GetReservedSlot(backend, Backend::Slots::ClientCert);
+  if (!client_cert_slot.isNullOrUndefined()) {
+    JS::RootedString client_cert_string(cx, client_cert_slot.toString());
+    auto client_cert_chars = core::encode(cx, client_cert_string);
```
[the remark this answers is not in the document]
guybedford: We always set this as a string in the slot. I can try use a `MOZ_ASSERT` to do an extra check.

runtime/fastly/builtins/secret-store.cpp
guybedford: Oh `plaintext` is a noun and not a compound word! Okay then `rawBytes` makes much more sense, thanks.

runtime/fastly/builtins/backend.cpp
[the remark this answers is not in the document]
elliottt: Ah, I missed that it's a slot that we manage. A debug assert would be fine here, thanks!

runtime/fastly/builtins/backend.cpp
elliottt: Ah, I missed that it was being overwritten. Thanks!

elliottt (approved):

PR #798, opened 2024-06-05
Starlingmonkey timers

guybedford:
This integrates Fastly support for timers on top of the new `deadline()` hook on AsyncTask for StarlingMonkey, which is supported by StarlingMonkey timers.

Timers are represented by a handle that is never used in Fastly's host call model. When encountering a timer in the select host hook we check its deadline and progress it if it is ready. We also bound the select call timeout now by the least deadline when timer tasks are present, as we previously implemented.

This architecture also exposes an implementation gap in that setting timers when no other async tasks were pending would never resolve. Without any other primitives to work with a busy loop is implemented in this case to properly support real timeouts and a new test is added. Alternatively, we could throw an error here but I strongly believe we should focus on supporting the code our users write instead of throwing for standard JS patterns.

### DIFF
--- a/runtime/fastly/host-api/host_api.cpp
+++ b/runtime/fastly/host-api/host_api.cpp
@@ -75,34 +75,37 @@ size_t api::AsyncTask::select(std::vector<api::AsyncTask *> *tasks) {
   fastly_world_list_handle_t hs{.ptr = handles.data(), .len = handles.size()};
   fastly_world_option_u32_t ret;
   fastly_compute_at_edge_types_error_t err = 0;
-  if (!fastly_compute_at_edge_async_io_select(&hs, (soonest_deadline - now) / MILLISECS_IN_NANOSECS,
-                                              &ret, &err)) {
-    abort();
-  } else if (ret.is_some) {
-    // The host index will be the index in the list of tasks with the timer tasks filtered out.
-    // We thus need to offset the host index by any timer tasks appearing before the nth
-    // non-timer task.
-    size_t task_idx = 0;
-    for (size_t idx = 0; idx < tasks_len; ++idx) {
-      if (tasks->at(idx)->id() != NEVER_HANDLE) {
-        if (ret.val == task_idx) {
-          return idx;
+
+  while (true) {
+    if (!fastly_compute_at_edge_async_io_select(
+            &hs, (soonest_deadline - now) / MILLISECS_IN_NANOSECS, &ret, &err)) {
+      abort();
+    } else if (ret.is_some) {
+      // The host index will be the index in the list of tasks with the timer tasks filtered out.
+      // We thus need to offset the host index by any timer tasks appearing before the nth
+      // non-timer task.
+      size_t task_idx = 0;
+      for (size_t idx = 0; idx < tasks_len; ++idx) {
+        if (tasks->at(idx)->id() != NEVER_HANDLE) {
+          if (ret.val == task_idx) {
+            return idx;
+          }
+          task_idx++;
         }
-        task_idx++;
       }
+      abort();
+    } else {
+      // No value case means a timeout, which means soonest_deadline_idx is set.
+      MOZ_ASSERT(soonest_deadline > 0);
+      MOZ_ASSERT(soonest_deadline_idx != -1);
+      // Verify that the task definitely is ready from a time perspective, and if not loop the host
+      // call again.
+      now = MonotonicClock::now();
+      if (soonest_deadline > now) {
+        continue;
+      }
+      return soonest_deadline_idx;
     }
-    abort();
-  } else {
-    // No value case means a timeout, which means soonest_deadline_idx is set.
-    MOZ_ASSERT(soonest_deadline > 0);
-    MOZ_ASSERT(soonest_deadline_idx != -1);
-    // Verify that the task definitely is definitely ready from a time perspective, and if not add
-    // an extra sleep.
-    now = MonotonicClock::now();
-    if (soonest_deadline > now) {
-      sleep_until(soonest_deadline, now);
-    }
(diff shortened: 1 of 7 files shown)


runtime/fastly/host-api/host_api.cpp
```diff
@@ -12,24 +12,86 @@
...
 
+#define NEVER_HANDLE 0xFFFFFFFE
+
 size_t api::AsyncTask::select(std::vector<api::AsyncTask *> *tasks) {
   size_t tasks_len = tasks->size();
   fastly_compute_at_edge_async_io_handle_t *handles =
       new fastly_compute_at_edge_async_io_handle_t[tasks_len];
-  for (int i = 0; i < tasks_len; i++) {
-    handles[i] = tasks->at(i)->id();
+  uint64_t now = 0;
```
elliottt, line 24: The allocation of `handles` looks like it leaks here, as I'm not finding a corresponding `delete`. What do you think about making it a `std::vector<fastly_compute_at_edge_async_io_handle_t>` instead? You could use the `reserve` method with `tasks_len` to pre-allocate the full buffer, and avoid the manual index tracking with `task_idx` by using `emplace_back` instead. Additionally, you could use the `data()` method to fetch the raw pointer when constructing the `fastly_world_list_handle_t` value lower down.

As there are a lot of early returns in this function, avoiding the need to `delete` the `handles` pointer at each would be good.

runtime/fastly/host-api/host_api.cpp
```diff
@@ -12,24 +12,86 @@
...
+      handles[task_idx] = handle;
+      task_idx++;
+    }
   }
-  fastly_world_list_handle_t hs{.ptr = handles, .len = tasks_len};
+
+  // When there are no async tasks, we have no choice to poll on time but to busy-loop.
+  if (task_idx == 0) {
+    MOZ_ASSERT(soonest_deadline > 0);
+    while (true) {
```
elliottt, line 57: We can implement this without busy waiting. rust implements `std::thread::sleep` by using `poll_oneoff` with a clock subscription, and we could do the same here: https://github.com/rust-lang/rust/blob/7ebd2bdbf6d798e6e711a0100981b0ff029abf5f/library/std/src/sys/pal/wasi/thread.rs#L137-L170

runtime/fastly/host-api/host_api.cpp
guybedford: Added a vector.

runtime/fastly/host-api/host_api.cpp
guybedford: Updated to use `nanosleep`.

runtime/fastly/host-api/host_api.cpp
```diff
@@ -10,26 +10,92 @@
...
-      new fastly_compute_at_edge_async_io_handle_t[tasks_len];
-  for (int i = 0; i < tasks_len; i++) {
-    handles[i] = tasks->at(i)->id();
+  std::vector<fastly_compute_at_edge_async_io_handle_t> handles;
+  handles.reserve(tasks_len);
+  uint64_t now = 0;
+  uint64_t soonest_deadline = 0;
+  size_t soonest_deadline_idx = -1;
+  for (size_t idx = 0; idx < tasks_len; ++idx) {
+    auto task = tasks->at(idx);
```
elliottt, line 30: I like `auto *` to make it obvious that the named value is a pointer vs something that implements `operator->`.

```suggestion
    auto *task = tasks->at(idx);
```

runtime/fastly/host-api/host_api.cpp
```diff
@@ -10,26 +10,92 @@
...
+    }
+    return soonest_deadline_idx;
   }
-  fastly_world_list_handle_t hs{.ptr = handles, .len = tasks_len};
+
+  fastly_world_list_handle_t hs{.ptr = handles.data(), .len = handles.size()};
   fastly_world_option_u32_t ret;
   fastly_compute_at_edge_types_error_t err = 0;
-  if (!fastly_compute_at_edge_async_io_select(&hs, 0, &ret, &err)) {
+  if (!fastly_compute_at_edge_async_io_select(&hs, (soonest_deadline - now) / 1000000, &ret,
```
elliottt, line 71: It would be nice to factor the constants here out into conversion functions. You could use the `std::chrono::duration` conversion functions, but those get a bit wordy. I'd settle for naming `1000000` as `MS` and `1000000000` as `US` constants somewhere in this file :)

runtime/fastly/host-api/host_api.cpp
```diff
@@ -10,26 +10,92 @@
...
     abort();
+  } else {
+    // No value case means a timeout, which means soonest_deadline_idx is set.
+    MOZ_ASSERT(soonest_deadline > 0);
+    MOZ_ASSERT(soonest_deadline_idx != -1);
+    // Verify that the task definitely is definitely ready from a time perspective, and if not
+    // repeat the entire routine as we cannot progress a timer before it's actually ready.
+    now = MonotonicClock::now();
+    if (soonest_deadline > now) {
+      return api::AsyncTask::select(tasks);
```
elliottt, line 96: It might be worth reworking this to have a top-level loop instead of a recursive call. There could be a situation where the stars align and the stack is blown.

elliottt (approved):
I think this all looks reasonable! I had some suggestions for small tweaks.

runtime/fastly/host-api/host_api.cpp
guybedford: Added

runtime/fastly/host-api/host_api.cpp
guybedford: I've added an extra sleep here to correct it.

PR #805, opened 2024-06-18
feat: add out-of-memory callback with stderr log

guybedford:
This adds an out of memory callback to SpiderMonkey to log the error to stderr.

The message logged is `Critical Error: out of memory`.


LeslieCarr:
@jaskiratr

elliottt (approved):
Looks good! It's not clear how easily we could test this, as our test suite would only be able to observe a 500. Does the OOM callback also terminate the instance?

guybedford:
The error does not result in a panic unless the code handling it chooses to panic. For normal JS code it becomes a JS exception with the message being out of memory.

Yeah without a nice stdout filtering test handler it's a bit tricky, but I did verify it on a simple over-allocation case.

PR #809, opened 2024-06-20
chore: Switch to using the config-store host calls

### LINKED ISSUE #801: Config stores should use `fastly_config_store` hostcalls
The `ConfigStore` interface should use [these newer host interfaces](https://github.com/fastly/Viceroy/blob/main/lib/compute-at-edge-abi/config-store.witx) rather than the Edge Dictionary-compatible hostcalls that are currently being called for both config stores and dictionaries. Config stores are subject to looser runtime limits, but the newer hostcalls must be used to take advantage of them fully.

elliottt:
Fixes #801

Use the config-store host calls for the ConfigStore api, instead of the deprecated dictionary host calls.


runtime/js-compute-runtime/host_interface/component/fastly_world_adapter.cpp
```diff
@@ -785,6 +785,36 @@ bool fastly_compute_at_edge_dictionary_get(fastly_compute_at_edge_dictionary_han
...
+                                              fastly_compute_at_edge_types_error_t *err) {
+  return convert_result(
+      fastly::config_store_open(reinterpret_cast<char *>(name->ptr), name->len, ret), err);
+}
+
+bool fastly_compute_at_edge_config_store_get(fastly_compute_at_edge_config_store_handle_t h,
+                                             fastly_world_string_t *key,
+                                             fastly_world_option_string_t *ret,
+                                             fastly_compute_at_edge_types_error_t *err) {
+  ret->val.ptr = static_cast<uint8_t *>(cabi_malloc(CONFIG_STORE_ENTRY_MAX_LEN, 1));
```
guybedford, line 799: I couldn't see offhand directly where `CONFIG_STORE_ENTRY_MAX_LEN` is defined. Is it definitely set correctly?

guybedford (approved):
Thanks for picking this one up!

runtime/js-compute-runtime/host_interface/component/fastly_world_adapter.cpp
elliottt: Yep, it's in `fastly.h`:

https://github.com/fastly/js-compute-runtime/blob/1c21c9a7e1ee484c2fb35d614fcda8d373822307/runtime/js-compute-runtime/host_interface/fastly.h#L21

and

https://github.com/fastly/js-compute-runtime/blob/1c21c9a7e1ee484c2fb35d614fcda8d373822307/runtime/fastly/host-api/fastly.h#L21

PR #810, opened 2024-06-20
StarlingMonkey WPT tests

guybedford:
This enables the WPT tests for the StarlingMonkey port on CI.

In the process StarlingMonkey is updated to the latest version, including with the upstream fixes:

* https://github.com/bytecodealliance/StarlingMonkey/pull/69
* https://github.com/bytecodealliance/StarlingMonkey/pull/70
* https://github.com/bytecodealliance/StarlingMonkey/pull/71
* https://github.com/bytecodealliance/StarlingMonkey/pull/72

It also turns out StarlingMonkey was actually passing one extra WPT than our build. To resolve that this PR also updates our `rust-url` crate to the latest and to line up with StarlingMonkey's implementation, which replaces https://github.com/fastly/js-compute-runtime/pull/806. The remaining port TODOs are also completed here.


guybedford:
From 132 failures initially, there are now just 7 WPT tests left to pass:

* html/webappapis/timers/missing-timeout-setinterval.any.js
* html/webappapis/timers/clearinterval-from-callback.any.js
* url/urlsearchparams-delete.any.js
* url/url-setters.any.js (2)
* fetch/api/response/response-init-001.any.js
* fetch/api/response/response-error.any.js

.github/workflows/starlingmonkey.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
     - name: Run tests
       timeout-minutes: 20
-      run: node ./tests/wpt-harness/run-wpt.mjs --starlingmonkey -vv
+      run: node ./tests/wpt-harness/run-wpt.mjs -vv
```
elliottt, line 110: Was removing the `--starlingmonkey` flag intentional?

runtime/fastly/host-api/host_api.cpp
```diff
@@ -1819,11 +1819,13 @@ const std::optional<std::string> FastlySendError::message() const {
...
   case dns_error: {
-    return "DNS error (rcode={}, info_code={})" /*, this->dns_error_rcode,
-                        this->dns_error_info_code*/
-        ;
+    // allocate maximum len of error message
+    char *buf = (char *)malloc(34 + 10 + 1);
+    int written = sprintf(buf, "DNS error (rcode=%d, info_code=%d)", this->dns_error_rcode,
+                          this->dns_error_info_code);
+    MOZ_ASSERT(written > 34);
+    return std::string(buf, written);
```
elliottt, line 1828: I believe that this `std::string` constructor will copy bytes out of `buf`, not take it over. As such, this will leak the allocation for `buf`. Given that the size of the buffer is known, why not use a stack allocated array?
```suggestion
    // allocate maximum len of error message
    char buf[34 + 10 + 1];
    int written = snprintf(buf, sizeof(buf), "DNS error (rcode=%d, info_code=%d)", this->dns_error_rcode,
                          this->dns_error_info_code);
    MOZ_ASSERT(written > 34);
    return std::string(buf, written);
```

elliottt (commented):
Looks good! Just one comment about buffer allocation.

.github/workflows/starlingmonkey.yml
guybedford: Yes, I originally was going to use this flag to have separate expectations for StarlingMonkey and js-compute-runtime, but was able to get them to match up in the end.

elliottt (approved):
Looks great!

PR #815, opened 2024-06-26
include StarlingMonkey build for release

guybedford:
This makes the `--starlingmonkey` flag public and binary to the release build and publish (although still experimental).

It also removes the former `--component` flag and binary, so our tarball should remain roughly the same size.


src/parseInputs.js
```diff
@@ -71,12 +59,8 @@ export async function parseInputs(cliInputs) {
...
-      case "--component": {
-        useComponent();
-        break;
-      }
       case "--component-adapter": {
-        useComponent();
+        component = true;
         adapter = cliInputs.shift();
         break;
       }
```
elliottt, line 66: Should we remove this too?

guybedford: Added.

elliottt (approved):
:fire:

Issue #814, opened 2024-06-25
Implement two hostcalls for obtaining address/port used to connect to origin


kpfleming:
These [two hostcalls](https://github.com/fastly/ExecuteD/blob/30ce2686a4020ea896d6adc4bfc077485adf2a14/wasm_abi/compute-at-edge-abi/compute-at-edge.witx#L643) were recently added. Support for them has already been added in the Rust SDK, but should also be added in the JavaScript SDK.

guybedford:
Thanks for sharing, I will aim to get this together for a release either next week or the week of the 8th of July.

guybedford:
I have some work-in-progress code here, but a release is currently blocked on Viceroy support for these new hostcalls.

guybedford:
This was released in 3.18.0.

PR #817, opened 2024-06-28
feat: support for Response.prototype.ip and port via get_addr_dest_ip & get_addr_dest_port

guybedford:
This implements https://github.com/fastly/js-compute-runtime/issues/814 adding support for `response.ip` and `response.port` getters.

These values are `undefined` in two scenarios - for cached responses and also for locally created responses which we check via the `!Response::is_upstream()` call to skip the host call. I've added a note in the type docs that cacheOverride: pass should be used to ensure a value if required.


types/globals.d.ts
```diff
@@ -1128,6 +1128,14 @@ interface Response extends Body {
...
   // readonly type: ResponseType;
   readonly url: string;
+  /**
+   * Fastly-specific property - obtain the IP address used for the request
+   */
+  readonly ip: string;
+  /**
+   * Fastly-specific property - Obtain the port used for the request
+   */
+  readonly port: number;
```
JakeChampion, line 1138: Non-blocking comment:

As these are only valid for responses which have come from a request, we could create a new TypeScript Interface for that scenario and update our `fetch` type definition to return it, that way it will be clear in the types that these properties are not on synthetic responses - what do you think?

```ts
interface FetchedResponse extends Response {
  /**
   * Fastly-specific property - obtain the IP address used for the request
   */
  readonly ip: string;
  /**
   * Fastly-specific property - Obtain the port used for the request
   */
  readonly port: number;
}


declare function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<FetchedResponse>;
```

guybedford: It is also possible for these to not be set when the response comes from a request but that request is cached. Instead of providing an error I've just made both cases result in an `undefined` getter which I think better aligns with user expectations in that these values should typically be checked for truthiness before reading them in JS.

I think this design works well, and wiring through custom Response flavours would likely be unwieldy.

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -2513,6 +2514,73 @@ bool Response::bodyUsed_get(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+    args.rval().setUndefined();
+    return true;
+  }
+  if (ret->len == 4) {
+    char *out = (char *)malloc(INET_ADDRSTRLEN);
+    inet_ntop(AF_INET, ret->ptr.get(), out, 16);
+    JS::RootedString text(
+        cx, JS_NewStringCopyUTF8N(cx, JS::UTF8Chars(out, strnlen(out, INET_ADDRSTRLEN))));
+    if (!text) {
+      return false;
```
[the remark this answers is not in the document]
guybedford: It will always be valid ASCII at this point so the only really valid failure would be an OOM error during the copy. We could switch it to an assert instead.

JakeChampion (approved):

PR #818, opened 2024-06-28
fix: debug build & tests

guybedford:
This fixes and tests the debug build against the SDK integration test suite.

Some assertions are no longer valid and need to be disabled since the KV store work shared these methods for KV store objects.

### DIFF
--- a/runtime/fastly/builtins/fetch/request-response.cpp
+++ b/runtime/fastly/builtins/fetch/request-response.cpp
@@ -238,17 +238,14 @@ uint32_t RequestOrResponse::handle(JSObject *obj) {
 }
 
 bool RequestOrResponse::has_body(JSObject *obj) {
-  // MOZ_ASSERT(is_instance(obj));
   return JS::GetReservedSlot(obj, static_cast<uint32_t>(Slots::HasBody)).toBoolean();
 }
 
 host_api::HttpBody RequestOrResponse::body_handle(JSObject *obj) {
-  // MOZ_ASSERT(is_instance(obj));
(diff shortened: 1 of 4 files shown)


guybedford:
The requirement for `sdktest` is not passing here because of the rename due to adding both debug and release. Once this lands we should update the requirement to both `CI/sdktest (debug)` and `CI/sdktest (release)` instead of just `sdktest`.

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -238,17 +238,17 @@ uint32_t RequestOrResponse::handle(JSObject *obj) {
 }
 
 bool RequestOrResponse::has_body(JSObject *obj) {
-  MOZ_ASSERT(is_instance(obj));
+  // MOZ_ASSERT(is_instance(obj));
```
elliottt, line 241: Could you either add a TODO to each commented out assertion, or remove them? It's fine to leave them with a TODO if the intent is to fix them, but I'd rather we remove them entirely otherwise.

elliottt:
Let's update the set of required checks before this merges, so that we can merge without bypassing.

runtime/fastly/builtins/fetch/request-response.cpp
guybedford: Sure, added.

guybedford:
> Is it possible to update the set of required checks before this merges, so that we can merge without bypassing?

Yeah it is, this seems to be working now.

elliottt (approved):

Issue #820, opened 2024-07-05
Allow commands/arrgs in spawnSync() to contain whitespace


alharirihamzah:
Commands and arrgs in `spawnSync()` with paths that includes spaces in them are broken, for example:
`/Users/hamzah/dev/MY APP/src/filename...`

The input in [containsSyntaxErrors](https://github.com/fastly/js-compute-runtime/blob/main/src/containsSyntaxErrors.js#L9) might contain a whitespace "like in my case"
![Screenshot 2024-07-05 at 10 28 17](https://github.com/fastly/js-compute-runtime/assets/42102276/c2678e3d-0f52-44a7-94a0-b15f2de6e92d)

_Note that a similar fix has been done but for the command in [containsSyntaxErrors - L#6](https://github.com/fastly/js-compute-runtime/blob/main/src/containsSyntaxErrors.js#L6)_

The exact same issue occurs while compiling the app to wasm in the command in [compileApplicationToWasm - L#120](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L120) and in the arrgs in [compileApplicationToWasm - #L128-#L129](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L128-L129)

PR: #821

JakeChampion:
This was fixed by https://github.com/fastly/js-compute-runtime/pull/821 and release in version [v3.16.1](https://github.com/fastly/js-compute-runtime/releases/tag/v3.16.1)

PR #821, opened 2024-07-05
Allow commands/arrgs in spawnSync() to contain whitespace

alharirihamzah:
Issue: #820 

This PR fixes whitespaces in commands and arrgs when calling `spawnSync()`

A simple fix is to always wrap commands and arrgs with double quotes to avoid similar issues.

### DIFF
--- a/src/compileApplicationToWasm.js
+++ b/src/compileApplicationToWasm.js
@@ -117,7 +117,7 @@ export async function compileApplicationToWasm(
 
   try {
     let wizerProcess = spawnSync(
-      wizer,
+      `"${wizer}"`,
       [
         "--inherit-env=true",
         "--allow-wasi",
@@ -125,8 +125,8 @@ export async function compileApplicationToWasm(
         ...starlingMonkey ? [`--dir=${dirname(wizerInput)}`] : [],
         `--wasm-bulk-memory=true`,
         "-r _start=wizer.resume",
-        `-o=${output}`,
-        wasmEngine,
+        `-o="${output}"`,
+        `"${wasmEngine}"`,
       ],
       {
         stdio: [null, process.stdout, process.stderr],
--- a/src/containsSyntaxErrors.js
+++ b/src/containsSyntaxErrors.js
@@ -6,7 +6,7 @@ export function containsSyntaxErrors(input) {
         `"${process.execPath}"`,
         [
             "--check",
-            input,
+            `"${input}"`,
         ],
         {
             stdio: [null, null, null],


src/containsSyntaxErrors.js
```diff
@@ -6,7 +6,7 @@ export function containsSyntaxErrors(input) {
         `"${process.execPath}"`,
         [
             "--check",
-            input,
+            `"${input}"`,
```
guybedford, line 9: This seems to fix the bug that was reported, which might be caused by the fact that this is spawed with `shell: true`.

src/compileApplicationToWasm.js
```diff
@@ -117,16 +117,16 @@ export async function compileApplicationToWasm(
...
         "--inherit-env=true",
         "--allow-wasi",
         "--dir=.",
         ...starlingMonkey ? [`--dir=${dirname(wizerInput)}`] : [],
         `--wasm-bulk-memory=true`,
         "-r _start=wizer.resume",
-        `-o=${output}`,
-        wasmEngine,
+        `-o="${output}"`,
+        `"${wasmEngine}"`,
```
guybedford, line 129: Normally `spawnSync` doesn't require escaping of arguments as it automatically handles this (and this is also why `windowsVerbatimArguments` is an option.

Therefore unless we can prove there is a test case where this fails, I'm pretty sure this code path should support spaces okay in the command and arguments.

guybedford (commented):
Thanks for looking into the fix for this, that's really appreciated.

I haven't had a chance to test out the spawn behaviours properly, but have left some comments for an initial review.

guybedford:
I've posted https://github.com/fastly/js-compute-runtime/pull/822 with just the check changes, which I believe may be sufficient, but please let me know if I'm mistaken as well.

alharirihamzah:
Hi @guybedford!
Sorry I couldn't answer fast enough and really thanks for your time looking into this, much appreciated 😄

I have tried fixing only the check in [containsSyntaxErrors #L9](https://github.com/fastly/js-compute-runtime/blob/main/src/containsSyntaxErrors.js#L9) but then it broke in the other `spawnSync()` invocation in [compileApplicationToWasm - L#119](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L119)
Screenshot 2024-07-09 at 08 47 01

It still broken when including the fix for the input command for `wizerProcess` in [compileApplicationToWasm #L120](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L120) without the args
Screenshot 2024-07-09 at 08 48 14

And including the fix for the arg `wasmEngine` in [compileApplicationToWasm #L129](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L129) still not enough
Screenshot 2024-07-09 at 08 49 24

And finally by including the fix for the `output` arg in [compileApplicationToWasm #L128](https://github.com/fastly/js-compute-runtime/blob/main/src/compileApplicationToWasm.js#L128) fix the whole chain 🙃
Screenshot 2024-07-09 at 08 49 58

I hope this clear things out 😃
Let me know if you have any other questions and thanks again 🙏

src/compileApplicationToWasm.js
alharirihamzah: Unfortunately it still needs the double quotes around the input on L#120 and around the args on L#128-L#129

guybedford:
Okay, thank you for confirming all the usages are required. Yeah it seems this is because of the `shell: true`, although I do wonder why we even need that in the invocation.

Regardless will merge this for the next release, thank you!.

PR #822, opened 2024-07-08
fix: escaping for check function

guybedford:
This includes just the minimum fix for https://github.com/fastly/js-compute-runtime/issues/820 without the other changes.


alharirihamzah:
I left a [comment](https://github.com/fastly/js-compute-runtime/pull/821#issuecomment-2216826872) regarding the fix in #820

guybedford:
Landed in https://github.com/fastly/js-compute-runtime/pull/821.

PR #823, opened 2024-07-10
fix: use same rust version that StarlingMonkey uses

JakeChampion:


guybedford:
Strictly speaking this isn't necessary anymore since https://github.com/bytecodealliance/StarlingMonkey/pull/55 was added.

So I think this was actually a StarlingMonkey bug. I've posted a possible fix in https://github.com/bytecodealliance/StarlingMonkey/pull/76.

PR #829, opened 2024-07-11
fix: keep sdkVersion property always up-to-date with the correct version of the SDK

JakeChampion:


runtime/fastly/CMakeLists.txt
```diff
@@ -19,4 +19,6 @@ add_builtin(fastly::fetch SRC builtins/fetch/fetch.cpp builtins/fetch/request-re
 add_builtin(fastly::cache_override SRC builtins/cache-override.cpp)
 add_builtin(fastly::fetch_event SRC builtins/fetch-event.cpp DEPENDENCIES OpenSSL)
 
+add_compile_definitions(PUBLIC RUNTIME_VERSION=${RUNTIME_VERSION})
```
guybedford, line 22: Added this line and verified with a local build the string `starlingmonkey-3.16.2` is being correctly inlined.

guybedford (approved):

runtime/fastly/CMakeLists.txt
JakeChampion: Thank you 🙏

PR #834, opened 2024-07-12
fix: correct the documentation for the fastly:logger module

JakeChampion:
When https://github.com/fastly/js-compute-runtime/pull/804 landed, this functionality changed but we missed the note in the website documentation, this commit updates the notes.

The Logger constructor as it can now be used during build-time initialisation and the Logger.prototype.log method can not be used during build-time initialisation


guybedford (approved):

documentation/docs/fastly:logger/Logger/Logger.mdx
```diff
@@ -74,8 +72,8 @@ addEventListener("fetch", event => event.respondWith(app(event)));
 ```js
 import { Logger } from "fastly:logger";
+let logger = new Logger("splunk");
```
guybedford, line 75: Maybe const?

```suggestion
const logger = new Logger("splunk");
```

Issue #836, opened 2024-07-12
headers.getSetCookie support


guybedford:
We need to support this API to support the latest version of Hono - https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie, including testing with WPT.

Issue #221, opened 2022-08-23
Avoid normalizing headers in `maybe_add`


elliottt:
`maybe_add` already assumes that the arguments are normalized, so we shouldn't repeat that work.

guybedford:
This was included in https://github.com/fastly/js-compute-runtime/pull/844.

PR #844, opened 2024-07-19
feat: support getSetCookie on new StarlingMonkey headers implementation

guybedford:
This updates to share the StarlingMonkey headers implementation, based to the work in https://github.com/bytecodealliance/StarlingMonkey/pull/79.

Resolves 40 WPT cases, as well as implementing `getSetCookie` (https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie).

Resolves https://github.com/fastly/js-compute-runtime/issues/836 and https://github.com/fastly/js-compute-runtime/issues/221.


guybedford:
The remaining failures here are divergence between StarlingMonkey and js-compute-runtime now that we have new support for separate cookie headers.

I would like to suggest we land this after deprecating the js-compute-runtime build to avoid the need to refactor the js-compute-runtime implementation as well here, since it's a fairly large diff.

runtime/fastly/host-api/host_api.cpp
```diff
@@ -237,6 +420,182 @@ int32_t MonotonicClock::subscribe(const uint64_t when, const bool absolute) { re
...
+    names_res = generic_get_header_names<fastly::resp_header_names_get>(
+        this->handle_state_.get()->handle());
+  }
+  if (const auto err = names_res.to_err()) {
+    return Result<vector<tuple<HostString, HostString>>>::err(*err);
+  }
+
+  vector<tuple<HostString, HostString>> entries_vec;
+  for (auto &name : names_res.unwrap()) {
+    auto values_res = get(name);
```
elliottt, line 483: Which `get` function is this?

runtime/fastly/host-api/host_api.cpp
```diff
@@ -196,6 +198,187 @@ Response make_response(fastly::fastly_host_http_response &resp) {
...
+        continue;
+      if (str_cnt == str_max) {
+        strs = static_cast<fastly::fastly_world_string *>(
+            cabi_realloc(strs, str_max * sizeof(fastly::fastly_world_string), 1,
+                         (str_max + LIST_ALLOC_SIZE) * sizeof(fastly::fastly_world_string)));
+        str_max += LIST_ALLOC_SIZE;
+      }
+      strs[str_cnt].ptr = static_cast<uint8_t *>(cabi_malloc(i - offset + 1, 1));
+      strs[str_cnt].len = i - offset;
+      memcpy(strs[str_cnt].ptr, buf + offset, i - offset + 1);
```
elliottt, line 237: Could you compute the length separately and use it for both the allocation size and `len` field? It looks like there's one extra char allocated than is needed right now, as I'm pretty sure it's not required that a `HostString` be null-terminated.

elliottt (commented):
I had some comments, but overall this seems very reasonable!

runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -566,6 +566,10 @@ bool start_response(JSContext *cx, JS::HandleObject response_obj, bool streaming
   auto response = Response::response_handle(response_obj);
   auto body = RequestOrResponse::body_handle(response_obj);
 
+  // write all the headers
+  if (!RequestOrResponse::commit_headers(cx, response_obj))
+    return false;
```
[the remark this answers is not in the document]
guybedford: Yeah the time may have come to decide this... ideally I think we should follow the StarlingMonkey style guide though for conformity of the code, which doesn't include this. So maybe we should bring this up there instead? Posted https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces.

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -484,6 +517,43 @@ JSObject *RequestOrResponse::headers(JSContext *cx, JS::HandleObject obj) {
...
+// request or response being sent.
+bool RequestOrResponse::commit_headers(JSContext *cx, HandleObject self) {
+  JS::RootedObject headers(cx, RequestOrResponse::maybe_headers(self));
+  if (!headers) {
+    return true;
+  }
+  MOZ_ASSERT(Headers::mode(headers) == Headers::Mode::ContentOnly ||
+             Headers::mode(headers) == Headers::Mode::CachedInContent ||
+             Headers::mode(headers) == Headers::Mode::Uninitialized);
+  Headers::HeadersList *list = Headers::get_list(cx, headers);
```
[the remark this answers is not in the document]
guybedford: Well-spotted, this should not be using the `std::move` and instead it should just be passing the string_view as a borrow.

runtime/fastly/builtins/fetch/request-response.h
```diff
@@ -1,8 +1,10 @@
 #ifndef FASTLY_REQUEST_RESPONSE
 #define FASTLY_REQUEST_RESPONSE
 
+#include "../../../StarlingMonkey/builtins/web/fetch/headers.h"
 #include "../../host-api/host_api_fastly.h"
-#include "headers.h"
+
+using builtins::web::fetch::Headers;
```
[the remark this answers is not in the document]
guybedford: Thanks I always forget about this one.

runtime/fastly/host-api/host_api.cpp
guybedford: Added the namespace.

runtime/fastly/host-api/host_api.cpp
guybedford: This was in the host API work Jake did, just moved around, but yeah I can remove the unused null byte allocation.

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -484,6 +517,43 @@ JSObject *RequestOrResponse::headers(JSContext *cx, JS::HandleObject obj) {
...
+  host_api::HttpHeaders *headers_handle;
+  if (Request::is_instance(self)) {
+    headers_handle = Request::request_handle(self).headers_writable();
+  } else {
+    MOZ_ASSERT(Response::is_instance(self));
+    headers_handle = Response::response_handle(self).headers_writable();
+  }
+
+  for (const auto &tuple : *list) {
+    const auto &[name, value] = tuple;
```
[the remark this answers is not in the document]
guybedford: I had to refactor this to use `std::get` to remove the `std::move` by ensuring the `std::string_view` interpretation path.

runtime/fastly/host-api/host_api.cpp
```diff
@@ -202,6 +204,191 @@ Response make_response(fastly::fastly_host_http_response &resp) {
...
 }
 
+template <auto header_names_get>
+Result<std::vector<HostString>> generic_get_header_names(auto handle) {
+  Result<std::vector<HostString>> res;
+
+  fastly::fastly_world_list_string ret;
+  fastly::fastly_host_error err;
+  fastly::fastly_world_string *strs = static_cast<fastly::fastly_world_string *>(
+      cabi_malloc(LIST_ALLOC_SIZE * sizeof(fastly::fastly_world_string), 1));
```
elliottt, line 214: What do you think about pushing into the result vector in the main loop, instead of keeping the separate `strs` buffer that's free'd later? That would avoid a memory leak with `strs` when there's an error in the conditional on line 222.

elliottt (approved):
I think this looks good, but revisiting the places where we're consuming lists from the host would be good -- we can avoid additional allocation and avoid explicit frees if we build the vector while processing the return value.

runtime/fastly/host-api/host_api.cpp
guybedford: As discussed, let's handle this in a follow-up PR.

PR #845, opened 2024-07-19
feat: add support for Intl APIs

JakeChampion:
Note: This looks to only work for release builds and not debug builds


JakeChampion:
This changes causes a linker error and seg fault in llvm https://github.com/llvm/llvm-project/issues/100084

twhid:
@JakeChampion is this on the roadmap anywhere? It's a nice API to have available. Thanks for any info 🙏 !

PR #851, opened 2024-07-24
Allow docusaurus to build concurrently

### LINKED ISSUE #847: check_docusaurus CI step taking a long time
It could be worthwhile to investigate why the `check_docusaurus` CI step is taking much longer than the other checks.

In recent PRs it seems this check is now taking over 30 minutes, from around 10 minutes a month ago. It could be beneficial to resolve this to ensure we are able to maintain a fast merge iteration cycle.

JakeChampion:
Only keep committed the documentations for the latest version of each major -- note that these versions will still exist on the deployed website, as the statically generated pages are stored in a Fastly KV Store

Future work would be to only have one version of the documentation and list on each page which version a feature/method/class was added/changed in -- I've got this partially done in a local branch, but wanted to ship something now to speed up our CI builds

Resolves #847


guybedford:
Amazing thank you for looking into this!

guybedford (approved):
Should we add a step to the release workflow that deletes the last versioned docs? Can be a follow-up certainly.

Issue #853, opened 2024-07-24
Implement hostcall for obtaining server IP


kphillips:
A hostcall to expose the IP address on which the server received the HTTP request was recently added. Support has already been added in the Rust SDK (see https://docs.rs/fastly/latest/fastly/handle/fn.server_ip_addr.html), and needs to be added to the JS SDK.

guybedford:
Hi @kphillips. Thanks for checking this one. Support for this was added in 3.18.0, see https://js-compute-reference-docs.edgecompute.app/docs/globals/Response/prototype/ip. Feedback very welcome!

kphillips:
Hi @guybedford. The Rust docs characterize this as "the IP address on which the server _received_ the HTTP request," but this function appears to relate to an IP address where the backend response was received, because it explains an `undefined` case where there was no backend response, either from a cached or a synthetic response. I don't know whether those would always be the same address, and I don't think it's possible for the _receiving_ side to have an `undefined` case at this time.

guybedford:
Ahh right, so we implement the client IP and the request IP for fetch requests, but there's also a server IP.

Certainly we can add this.

It doesn't feel like it belongs on `clientInfo` on the fetch event since it is server metadata. Possible implementations might include:

* A function under `fastly:experimental`
* `fetchEvent.ip`
* `get Request.prototype.ip`

Marking as a feature, we'll add this as soon as we can.

PR #855, opened 2024-07-25
feat: Add FetchEvent.server object which contains information about the server which received the incoming HTTP request from the client.

JakeChampion:
Currently `FetchEvent.server` contains one property, `address`, which is the IP address of the server. In the future more properties may be added.

Example usage for this field:
```js

addEventListener("fetch", event => event.respondWith(app(event)))

async function app(event) {
  return new Response(`Received request on ${event.server.address}`);
}
```

Resolves https://github.com/fastly/js-compute-runtime/issues/853.


runtime/fastly/host-api/host_api.cpp
```diff
@@ -944,6 +944,20 @@ Result<HostBytes> HttpReq::downstream_client_ip_addr() {
...
+Result<HostBytes> HttpReq::downstream_server_ip_addr() {
+  Result<HostBytes> res;
+
+  fastly_world_list_u8_t octets;
+  fastly_compute_at_edge_types_error_t err;
+  if (!fastly_compute_at_edge_http_req_downstream_server_ip_addr(&octets, &err)) {
+    res.emplace_err(err);
+  } else {
+    res.emplace(make_host_bytes(octets));
+  }
```
guybedford, line 956: Are we sure it's never an optional none?

runtime/js-compute-runtime/builtins/server-info.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <arpa/inet.h>
+
+namespace builtins {
+
+namespace {
+
+JSString *address(JSObject *obj) {
+  JS::Value val = JS::GetReservedSlot(obj, static_cast<uint32_t>(ServerInfo::Slots::Address));
+  return val.isString() ? val.toString() : nullptr;
+}
```
guybedford, line 15: This seems like it doesn't need to be a dedicated function, and could just as well be inlined into `address_get` since it has a single caller.

runtime/js-compute-runtime/builtins/server-info.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+static JSString *retrieve_address(JSContext *cx, JS::HandleObject self) {
```
guybedford, line 17: Did you revert abstracting the IP bytes to JS string logic? I release we repeat the same routine for the request IP as well, does that mean we now have three functions doing the same thing? Perhaps reinvestigating sharing might be good.

runtime/fastly/host-api/host_api.cpp
JakeChampion: Yes we always have an IP for the servers in our pops which receive client requests 👍

runtime/js-compute-runtime/builtins/server-info.cpp
JakeChampion: Can do yes, I was keeping it similar to the implementation in Client info, happy to refactor it

runtime/js-compute-runtime/builtins/server-info.cpp
JakeChampion: I abstracted it in the starling build but not in this build, I can abstract here also

guybedford: Ah sorry I missed which one I was looking at! Starling-only abstraction is fine, but specifically perhaps we can make it a shared util shared by the `Request.ip` one too.

guybedford:
@JakeChampion I've pushed up a commit here to fix the CI build issue and also refactor the shared implementation. Feel free to make any further changes as necessary.

JakeChampion:
Thanks @guybedford - I have no other changes I'd make, if you could approve then we can land this 🙏

guybedford (approved):

PR #861, opened 2024-07-26
feat: use StarlingMonkey by default, --disable-starlingmonkey flag

guybedford:
This switches the default engine used by js-compute-runtime to be the StarlingMonkey engine, with a new `--disable-starlingmonkey` flag which can be used to revert back to the js-compute-runtime.wasm engine.

* When using `--disable-starlingmonkey`, the message `"Building with the js-compute-runtime.wasm engine"` is printed during the build.
* The `--starlingmonkey` flag continues to work as a noop, but no longer logs that the experimental StarlingMonkey engine is being used.
* The `fastly.sdkVersion` is now `x.y.z` for the StarlingMonkey build and `x.y.z-legacy` for the non-StarlingMonkey build under `--disable-starlingmonkey`.

This is marked as a feature commit, as the hope is that we can achieve this as a minor release.

Working through the risk analysis here:

1. All test coverage is passing and has been passing for a while.
2. Internal users have been using this flag without any issues reported so far.
3. If bugs are encountered, developers can use the disable flag, or lock a reversion of js-compute.
4. If we get any issues reported, we can immediately revert this PR changing the default back to js-compute-runtime.wasm until the issue is resolved.

The cost benefit here appears to me to work out. Further, even after landing, let's not rip out the js-compute-runtime.wasm engine for a period of at least say 4 weeks, so that we can remain in a position to perform step (4) at anytime.

Let's not land this until we've all discussed further and signed off on it though. //cc @JakeChampion @cfallin @elliottt

### DIFF
--- a/.github/actions/compute-sdk-test/action.yml
+++ b/.github/actions/compute-sdk-test/action.yml
@@ -31,5 +31,5 @@ runs:
       shell: bash
     - run: cd ./integration-tests/js-compute && yarn
       shell: bash
-    - run: cd ./integration-tests/js-compute && ./test.js --local
+    - run: cd ./integration-tests/js-compute && ./test.js --disable-starlingmonkey --local
       shell: bash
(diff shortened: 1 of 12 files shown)


JakeChampion (approved):
As discussed on the call, this looks good 👍

PR #864, opened 2024-07-29
refactor: remove the component adapter implementation

JakeChampion:
We no longer need this JS SDK component adapter implementation as we're pursuing a generic component adapter approach which will work for more SDKs than the JS one.

Let's take this time to simplify our implementation by removing this abstraction and inlining (plus simplifying where we can) the adapter functions into the host api.

This:
- Removes the js-compute-runtime-component.wasm file
- Removes all our vendored wit files and wit-bindgen commands as we are not dealing with wasm component functionality directly in this project anymore (... for the time being?)
- Inlines the component adapter functionality into the host_api abstraction


runtime/fastly/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.27)
 
 include("../StarlingMonkey/cmake/add_as_subproject.cmake")
 
-add_builtin(fastly::runtime SRC handler.cpp host-api/component/fastly_world_adapter.cpp)
+# add_builtin(fastly::runtime SRC handler.cpp host-api/component/fastly_world_adapter.cpp)
```
JakeChampion, line 5:
```suggestion
```

runtime/fastly/CMakeLists.txt
```diff
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.27)
 
 include("../StarlingMonkey/cmake/add_as_subproject.cmake")
 
-add_builtin(fastly::runtime SRC handler.cpp host-api/component/fastly_world_adapter.cpp)
```
guybedford, line 5: This shouldn't be deleted, as we do need `handler.cpp`, so perhaps just change it to:

```
add_builtin(fastly::runtime SRC host-api/component/fastly_world_adapter.cpp)
```

runtime/fastly/host-api/host_api.cpp
```diff
@@ -22,6 +22,67 @@ using host_api::Result;
 #define MILLISECS_IN_NANOSECS 1000000
 #define SECS_IN_NANOSECS 1000000000
 
+// Ensure that all the things we want to use the hostcall buffer for actually
+// fit into the buffer.
+#define HOSTCALL_BUFFER_LEN HEADER_MAX_LEN
+static_assert(DICTIONARY_ENTRY_MAX_LEN < HOSTCALL_BUFFER_LEN);
+static_assert(METHOD_MAX_LEN < HOSTCALL_BUFFER_LEN);
+static_assert(URI_MAX_LEN < HOSTCALL_BUFFER_LEN);
```
guybedford, line 30: Can we move these into `fastly.h` to be defined all together?

runtime/fastly/host-api/host_api.cpp
```diff
@@ -22,6 +22,67 @@ using host_api::Result;
...
+
+#define LIST_ALLOC_SIZE 50
```
guybedford, line 32: Again, this might be good to define in `fastly.h`?

guybedford (commented):
Will be great to go forward with this, we just need to ensure very careful review.

runtime/js-compute-runtime/host_interface/host_api.h
```diff
@@ -349,9 +345,6 @@ class HttpPendingReq final {
   explicit HttpPendingReq(Handle handle) : handle{handle} {}
   explicit HttpPendingReq(AsyncHandle async) : handle{async.handle} {}
 
-  /// Poll for the response to this request.
-  Result<std::optional<Response>> poll();
```
JakeChampion, line 353: This was never implemented in the adapter and is never used in the sdk

elliottt (approved):
Thanks for working through this!

As we discussed today, this was mostly copying the implementation out of the adapter and into the implementation of host_api.cpp. You've kept the other changes to a minimum, so I feel pretty confident that the tests passing means that everything is good here.

runtime/fastly/host-api/fastly.h
```diff
@@ -7,10 +7,39 @@ extern "C" {
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+namespace fastly {
 
-#include "./component/fastly_world.h"
+typedef struct fastly_world_string {
+  uint8_t *ptr;
+  size_t len;
+} fastly_world_string;
```
guybedford, line 15: It would be nice to see these removed entirely with less ceremony / better internal types, but we can always do that as a follow-up.

runtime/fastly/host-api/fastly.h
```diff
@@ -7,10 +7,39 @@ extern "C" {
...
-namespace fastly {
+typedef struct fastly_world_list_string {
+  fastly_world_string *ptr;
+  size_t len;
+} fastly_world_list_string;
+
+typedef struct fastly_world_list_u8 {
+  uint8_t *ptr;
+  size_t len;
+} fastly_world_list_u8;
```
guybedford, line 25: Eg this could probably be a span or similar?

guybedford (approved):
Checked various APIs and the overall approach seems of high quality.

This will definitely be a performance improvement, it would also be nice to measure and report it.

PR #868, opened 2024-07-31
CI: Always run the build job, never cancel it

JakeChampion:
We change the concurrency cancelling logic from applying to the whole workflow to instead being applied to individual jobs in the workflow, except for the build job, which we want to always run

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -166,7 +166,7 @@ jobs:
 
   run_wpt:
     concurrency:
-      group: ${{ github.head_ref }}-${{ github.workflow}}-run_wpt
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -386,8 +407,11 @@ jobs:
         path: starling.wasm
 
   starlingmonkey-run_wpt:
+    concurrency:
+      group: ${{ github.head_ref }}-${{ github.workflow}}
+      cancel-in-progress: true
     if: github.ref != 'refs/heads/main'
-    name: Run Web Platform Tests
+	name: Run Web Platform Tests (starlingmonkey)
```
JakeChampion, line 414:
```suggestion
    name: Run Web Platform Tests (starlingmonkey)
```

guybedford (approved):

PR #870, opened 2024-07-31
fix: perf: Use wasm-opt -O3 when making a release build

JakeChampion:


elliottt (approved):
Looks good! Are there changes we can make to the StarlingMnokey build to ensure that we're doing the same thing there as well?

JakeChampion:
> Looks good! Are there changes we can make to the StarlingMnokey build to ensure that we're doing the same thing there as well?

The StarlingMonkey build already does this

JakeChampion:
Yes StarlingMonkey has the same applied alreday

PR #871, opened 2024-07-31
 use clang-format included from wasi-sdk 

JakeChampion:
Ensures we use the clang-format that ships with the versioned wasi-sdk we use (and any other clang-format as different versions can format differently, which is a pain to deal with)

### DIFF
--- a/ci/clang-format.sh
+++ b/ci/clang-format.sh
@@ -48,7 +48,7 @@ for file in $(git ls-files | grep '\.\(cpp\|h\)$'); do
   fi
 
   formatted="${file}.formatted"
-  clang-format "$file" > "$formatted"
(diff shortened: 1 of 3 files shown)


ci/clang-format.sh
```diff
@@ -48,7 +48,7 @@ for file in $(git ls-files | grep '\.\(cpp\|h\)$'); do
   fi
 
   formatted="${file}.formatted"
-  clang-format "$file" > "$formatted"
+  /opt/wasi-sdk/bin/clang-format "$file" > "$formatted"
```
elliottt, line 51: I'm a little undecided about this. It might be nice to accept the clang-format program as an environment variable, but we also require wasi-sdk to build, so this path should always exist.

PR #874, opened 2024-07-31
perf: re-order the http methods so the most often requested is first and the least requested is last

JakeChampion:
this is to help us exist the loop sooner for the more common scenarios


guybedford (approved):
Focusing on unnecessary computation and copying might be a good focus for micro-optimization further.

PR #875, opened 2024-07-31
perf: remove the ciphersuite list parser and instead rely on the host

JakeChampion:


runtime/fastly/builtins/backend.cpp
```diff
@@ -1466,11 +879,6 @@ bool Backend::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
       return false;
     }
     std::string cipher_spec(ciphers_chars.begin(), ciphers_chars.len);
-    if (!is_cipher_suite_supported_by_fastly(cipher_spec)) {
-      JS_ReportErrorNumberASCII(cx, FastlyGetErrorMessage, nullptr,
-                                JSMSG_BACKEND_CIPHERS_NOT_AVALIABLE);
```
guybedford, line 1471: Can we test this error with this code path removed?

JakeChampion: The host does not yet validate the ciphersuite list, waiting on https://github.com/fastly/ExecuteD/issues/2390 and https://github.com/fastly/ExecuteD/issues/4205

PR #881, opened 2024-08-01
Experimental: add ahead-of-time compilation via weval and PBL.

cfallin:
This PR pulls in my work to use "weval", the WebAssembly partial evaluator, to perform ahead-of-time compilation of JavaScript using the PBL interpreter we previously contributed to SpiderMonkey. This work has been merged into the BA fork of SpiderMonkey in
bytecodealliance/gecko-dev#45,  bytecodealliance/gecko-dev#46, bytecodealliance/gecko-dev#47, bytecodealliance/gecko-dev#48, bytecodealliance/gecko-dev#51, bytecodealliance/gecko-dev#52, bytecodealliance/gecko-dev#53, bytecodealliance/gecko-dev#54, bytecodealliance/gecko-dev#55, and then integrated into StarlingMonkey in bytecodealliance/starlingmonkey#91.

The feature is off by default; it requires a `--enable-experimental-aot` flag to be passed to `js-compute-runtime-cli.js`. This requires a separate build of the engine Wasm module to be used when the flag is passed.

This should still be considered experimental until it is tested more widely. The PBL+weval combination passes all jit-tests and jstests in SpiderMonkey, and all integration tests in StarlingMonkey; however, it has not yet been widely tested in real-world scenarios.

Initial speedups we are seeing on Octane (CPU-intensive JS benchmarks) are in the 3x-5x range. This is roughly equivalent to the speedup that a native JS engine's "baseline JIT" compiler tier gets over its interpreter, and it uses the same basic techniques -- compiling all polymorphic operations (all basic JS operators) to inline-cache sites that dispatch to stubs depending on types. Further speedups can be obtained eventually by inlining stubs from warmed-up IC chains, but that requires warmup.

Important to note is that this compilation approach is *fully ahead-of-time*: it requires no profiling or observation or warmup of user code, and compiles the JS directly to Wasm that does not do any further codegen/JIT at runtime. Thus, it is suitable for the per-request isolation model (new Wasm instance for each request, with no shared state).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -394,7 +394,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        profile: [debug, release]
+        profile: [debug, release, weval]
     steps:
     - uses: actions/checkout@v3
       with:
@@ -409,10 +409,18 @@ jobs:
     - name: Build
       if: ${{ matrix.profile == 'debug' }}
       run: npm run build:starlingmonkey:debug
+    - name: Build
+      if: ${{ matrix.profile == 'weval' }}
+      run: npm run build:starlingmonkey:weval
     - uses: actions/upload-artifact@v3
       with:
         name: starling-${{ matrix.profile }}
         path: starling.wasm
+    - uses: actions/upload-artifact@v3
+      if: ${{ matrix.profile == 'weval' }}
+      with:
+        name: starling-${{ matrix.profile }}-ic-cache
+        path: starling-ics.wevalcache
 
   starlingmonkey-run_wpt:
     concurrency:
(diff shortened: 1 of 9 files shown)


cfallin:
@guybedford @JakeChampion I haven't quite worked out the CI flow yet -- the separate build of the engine and the wevalcache artifact both need to be packaged as well I guess. Any help appreciated there :-) I've verified a local build does work:

```
% ./js-compute-runtime-cli.js --enable-experimental-aot  ./richards.js richards.wasm
% viceroy richards.wasm
Log: Richards: 480
Log: ----
Log: Score (version 9): 480
^C%
% ./js-compute-runtime-cli.js ./richards.js richards-base.wasm
% viceroy richards-base.wasm
Log: Richards: 138
Log: ----
Log: Score (version 9): 138
^C%
```

cfallin:
I think I've re-discovered that some CI jobs require a branch in the main repo rather than a fork (e.g. [this CI failure](https://github.com/fastly/js-compute-runtime/actions/runs/10204047493/job/28237315838)); closing this PR to re-open another from such a branch.

PR #885, opened 2024-08-05
perf: improve performance of CoreCache.lookup and CoreCache.transactionLookup by creating the host request handle directly instead of via the WHATWG Request Constructor

JakeChampion:
We used to use Request::create to create the host request handle for the headers, which involved doing lots of WHATWG Request spec steps that were not required for this code path. We now create the host request handle directly, avoiding those not required WHATWG spec steps.


runtime/js-compute-runtime/builtins/cache-core.cpp
```diff
@@ -12,13 +13,181 @@ namespace builtins {
 
 namespace {
 
+JS::Result<host_api::HttpReq> extractHeadersIntoReq(JSContext *cx, JS::HandleValue headers_val,
```
JakeChampion, line 16: need to pull out into a new file in the common folder

guybedford:
I've integrated this change into https://github.com/fastly/js-compute-runtime/pull/844 now.

PR #887, opened 2024-08-07
update to latest wpt and update our assertion files

JakeChampion:


guybedford (approved):
Looks like the js-compute expectations may need a re-update still.

PR #894, opened 2024-08-16
refine task assertions, add test for task timeout

guybedford:
This adds a test for a timed out fetch request, that relies on the timeout feature of `async_select`.

This is currently failing on compute due to the recently fixed bug at https://github.com/fastly/ExecuteD/pull/4243, which should then pass once that rollout occurs.

This PR also refines the event loop task assertions and errors per debugging for this error.


integration-tests/js-compute/fixtures/app/src/timers.js
```diff
@@ -342,6 +342,35 @@ import { routes } from "./routes.js";
...
+        let timedOut = false
+        const first = fetch('https://httpbin.org/delay/1', {
+            backend: 'httpbin'
+        })
+        const second = Promise.race([
+            fetch('https://httpbin.org/delay/1', {
+                backend: 'httpbin'
+            }),
+            new Promise(resolve => setTimeout(resolve, 5)).then(() => {
+                console.error('TIMEOUT');
```
guybedford, line 355: These logs should be removed once the test is passing!

runtime/fastly/host-api/host_api.cpp
```diff
@@ -127,13 +127,16 @@ size_t api::AsyncTask::select(std::vector<api::AsyncTask *> &tasks) {
   fastly::fastly_host_error err = 0;
 
   while (true) {
-    if (!convert_result(fastly::async_select(handles.data(), handles.size(),
-                                             (soonest_deadline - now) / MILLISECS_IN_NANOSECS,
-                                             &ret),
+    MOZ_ASSERT(soonest_deadline == 0 || soonest_deadline > now);
+    uint32_t timeout = soonest_deadline > 0 ? (soonest_deadline - now) / MILLISECS_IN_NANOSECS : 0;
```
elliottt, line 131: Is it worth adding a comment here the `0` means no timeout?

elliottt (approved):
This looks great!

runtime/fastly/host-api/host_api.cpp
guybedford: Sure, how about this:

```suggestion
    // timeout value of 0 means no timeout for async_select
    uint32_t timeout = soonest_deadline > 0 ? (soonest_deadline - now) / MILLISECS_IN_NANOSECS : 0;
```

elliottt: Perfect!

PR #895, opened 2024-08-16
Remove --disable-starlingmonkey

guybedford:
Removes the js-compute-runtime build.


cfallin (approved):
Changes in toplevel driver and CI all look reasonable to me; momentous milestone when we merge this!

PR #914, opened 2024-08-27
fix: docs site kv-store name and toml definitions

guybedford:
It seems the PR in https://github.com/fastly/js-compute-runtime/pull/837 still needed some work, despite the passing tests.

Specifically tke kv-store name changed which is currently giving an error on the docs site deployment.

This PR changes the name back to the same kv-store name and corrects the kv-store name in the fastly.toml file.

This should then be a fast-tracked 3.21.2 to hopefully resolve that static docs site deployment.


guybedford:
This has now been rolled into https://github.com/fastly/js-compute-runtime/pull/917.

PR #915, opened 2024-08-27
fix: revert documentation website refactoring

guybedford:
This reverts https://github.com/fastly/js-compute-runtime/commit/e9114819350427b9fb6e6b58af291ded7c7d28fa which resulted in a failed documentation site publish.

Fast-tracking to fixup the deployment.


guybedford:
The dependency check is failing here because the dependencies before this PR landed did not pass review, therefore the revert fails dependency review. I'm temporarily disabling this check requirement to land the revert, then will reenable.

PR #917, opened 2024-08-27
Docs refactoring

guybedford:
Reposting this draft of the docs refactoring after the reverted deployment in https://github.com/fastly/js-compute-runtime/pull/915.

Hopefully we can then try this upgrade again with some more testing.


guybedford:
We need to check this new work resolves:

* Repeated publishes of unchanged files share the same KV entry.
* New versions.json updates have the right cache-control so that the new version publish is available immediately despite former caching.

PR #918, opened 2024-08-27
fix: documentation build for 3.21

guybedford:
The 3.21.2 publish successfully resolved the deployment issue, but it turns out there is still some manual process required in a docs deployment to bump the `versions.json` file versions listing.

This does that bump for a follow-up release to ensure a versioned page for the new version. We should ensure this step is automated for future deployments.


guybedford:
Turns out this is the same caching issue we've had in the past, and a refresh fixes it, so still another thing to look into with the refactoring.

PR #934, opened 2024-08-30
Update event loop stall error as no longer a panic

guybedford:
This adds a test for an event loop stall and ensures that it is just a warning instead of being a panic, so that it is distinguished from critical runtime errors / internal errors.

I also refactored the JS test runner to have some options for how to stream the response body while investigating some tests cases around this even if I didn't use this full functionality in the end.

This also includes a couple of recent Starlingmonkey patches that improve error outputs here - (https://github.com/bytecodealliance/StarlingMonkey/pull/132, https://github.com/bytecodealliance/StarlingMonkey/pull/139).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,7 +238,7 @@ jobs:
 
   sdktest:
     concurrency:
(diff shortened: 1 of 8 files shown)


elliottt (approved):
The handler.cpp changes look good to me :+1:

PR #936, opened 2024-08-31
Replace Yarn with Npm

guybedford:
Last Yarn release was 6 months ago, feels like npm may be better maintained at this point and there's not much speed difference left these days.


elliottt (approved):
Seems reasonable to me!

PR #937, opened 2024-09-03
fix: Device.toJSON() properties

guybedford:
Resolves a bug where `Device.toJSON()` would always return `null` for all `isDesktop()` style boolean properties as well as for `hardwareType`.

In the process I also refactored the test file to simplify the error handler hoping to do this as I touch each file going forward.


elliottt (approved):
This looks reasonable to me, but it looks like some test instability has been introduced.

guybedford:
The penaltybox tests seem to have been flakey on the latest XQD for some reason, but this has been an ongoing thing with those tests that seems to vary from time to time.

PR #939, opened 2024-09-03
AOT compilation: update to weval 0.2.12.

cfallin:
This includes a fix for irreducible control flow; see [this PR](https://github.com/cfallin/waffle/pull/13) in waffle.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@bytecodealliance/jco": "^1.3.1",
         "@bytecodealliance/wizer": "^3.0.1",
-        "@cfallin/weval": "^0.2.9",
+        "@cfallin/weval": "^0.2.11",
         "acorn": "^8.12.1",
         "acorn-walk": "^8.3.3",
         "esbuild": "^0.23.0",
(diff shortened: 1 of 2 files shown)


cfallin:
@guybedford @elliottt I've intentionally avoided a submodule update here as we have the open issue with the new headers implementation in StarlingMonkey interacting with weval somehow in debug builds; the version of weval pulled in by the `package.json` here is used by the js-compute-runtime toplevel driver, separately from the version of weval that the StarlingMonkey submodule uses.

cfallin:
Hmm, this is triggering some issues deep in the Wasm backend; I need to investigate further.

guybedford:
> @guybedford @elliottt I've intentionally avoided a submodule update here as we have the open issue with the new headers implementation in StarlingMonkey interacting with weval somehow in debug builds;

For what it is worth, the new headers work landed already in js-compute-runtime without any error, so we are already on the latest StarlingMonkey here. But fine to maintain separate versioning as well!

cfallin:
I've pushed a new version of weval (0.2.12) with some additional fixes that I believe should pass tests now (let's see how CI does); analogous update in StarlingMonkey at bytecodealliance/StarlingMonkey#141.

> For what it is worth, the new headers work landed already in js-compute-runtime without any error, so we are already on the latest StarlingMonkey here. But fine to maintain separate versioning as well!

Ah, interesting, and we're passing WPT here still -- I had been worried about hitting the same issue as bytecodealliance/StarlingMonkey#114 but I guess for whatever reason we don't see that here? In any case, I'm happy to do a separate PR to update to latest StarlingMonkey once above PR lands too, or land it first and update this one, either way.

guybedford (approved):

PR #942, opened 2024-09-04
Run Prettier

guybedford:
Runs Prettier against all the JS code, adding a new `npm run format` command to perform this.

Also fixes up one of the GitHub workflows in the process.


elliottt (approved):
Looks reasonable to me, thanks for pulling this out into a separate PR!

Is it easy to add a format checking step to CI to ensure that the codebase stays formatted?

guybedford:
Good point, added.

PR #945, opened 2024-09-05
fix: --experimental-top-level-await support

guybedford:
This fixes support for `--experimental-top-level-await` which was broken in the StarlingMonkey upgrade path.

The new implementation now uses full native ESM support in the StarlingMonkey engine, so we now implement this without needing special wrapping and rewriting tricks.

This PR also adds a test suite for this new mode, which in theory in future should become the main mode running JS modules natively, such that we should make this flag the default in future and migrate the full test suite to run with TLA enabled.

In the process a couple of test flakes are made less flakey too.


cfallin (approved):
LGTM!

elliottt (approved):

Issue #925, opened 2024-08-27
vcpu time hostcall


guybedford:
Support `get-vcpu-ms`.

PR #950, opened 2024-09-06
Add support for vcpu ms

guybedford:
Resolves https://github.com/fastly/js-compute-runtime/issues/925.

This is implemented as a new `fastly:runtime` interface, exporting a `vCpuTime()` function.


.vscode/c_cpp_properties.json
```diff
@@ -3,12 +3,11 @@
...
-                "${workspaceFolder}/runtime/spidermonkey/debug/include",
-                "${workspaceFolder}/runtime/spidermonkey/release/include",
+                "${workspaceFolder}/runtime/StarlingMonkey/include",
+                "${workspaceFolder}/runtime/StarlingMonkey/builtins/web",
+                "${workspaceFolder}/runtime/fastly/build-debug/",
                 "/opt/wasi-sdk/share/wasi-sysroot/include/",
-                "${workspaceFolder}/runtime/js-compute-runtime",
-                "${workspaceFolder}/runtime/js-compute-runtime/build/openssl-3.0.7/include",
-                "${workspaceFolder}/runtime/js-compute-runtime/third_party/fmt/include"
+                "${workspaceFolder}/runtime/fastly/host-api"
```
guybedford, line 10: I'm trying to work out how to get intellisense to configure properly, but am still not finding this working. Any suggestions very much welcome, but at least this is a step in the right direction.

integration-tests/js-compute/fixtures/app/src/runtime.js
```diff
@@ -0,0 +1,23 @@
...
+  ok(cpuTime > 0);
+  ok(cpuTime < 1000);
+  const arr = [];
+  for (let i = 0; i < 1000; i++) {
+    arr.push(i);
+  }
+  const cpuTime2 = vCpuTime();
+  ok(cpuTime2 > cpuTime);
+  ok(cpuTime2 - cpuTime > 1);
+  ok(cpuTime2 - cpuTime < 1000);
```
elliottt, line 21: The two comparisons around the upper bound of `1000` might end up being flaky tests. Does it make sense to bump it up a bit more?

elliottt (approved):
Looks good! Just one comment about the test possibly being flaky.

.vscode/c_cpp_properties.json
elliottt: `cmake` can generate a `compile_commands.json`, which might be more reliable. Then you can configure `clangd` to read that instead.

integration-tests/js-compute/fixtures/app/src/runtime.js
guybedford: It's denoted in milliseconds if that helps to clarify?

.vscode/c_cpp_properties.json
guybedford: Amazing thank you, this worked.

integration-tests/js-compute/fixtures/app/src/runtime.js
elliottt: Yep, I'm just worried that the test will drift in execution time, and cause failures if somehow it goes beyond one second. If you're confident that we won't see that, then we can go ahead and merge.

guybedford: I've bumped it to 3000.

PR #953, opened 2024-09-09
feat: expose purgeSurrogateKey hostcall

guybedford:
This resolves https://github.com/fastly/js-compute-runtime/issues/923 in adding support for a `purgeSurrogateKey` call which takes a key string and an optional soft purge boolean.

This is added to `fastly:runtime` in the spirit of this new import subsystem replacing the Fastly global, which currently mostly mirrors `fastly:experimental` anyway. The purge host call itself is refactored to move into the `host_api::Runtime` namespace as part of this process.

This PR also refactors the test.json usage to default the request and response and environment properties to avoid unnecessary repetition.


elliottt (approved):
Looks good!

PR #959, opened 2024-09-10
rename fastly:runtime to fastly:compute

guybedford:
A last minute change of heart on the addition of the two new hostcalls.

Instead of making them available in the `fastly:runtime` subsystem:

```js
import { vCpuTime, purgeSurrogateKey } from 'fastly:runtime';
```

this makes them under a `fastly:compute` import:

```js
import { vCpuTime, purgeSurrogateKey } from 'fastly:compute';
```

Feedback very welcome.


elliottt (approved):
Looks good!

PR #962, opened 2024-09-11
fix: windows support & windows ci

guybedford:
Finally hit the first major StarlingMonkey regression with Windows support inadvertently broken on the StarlingMonkey upgrade path.

This resolves that and also adds a Windows and Mac CI for the CLI tests to avoid future regressions.

One of the major issues here is cloning the repo itself on Windows, because of the use of `:` in paths. As a simple workaround for now I've added a script which will remove and add the fastly prefix across documentation pages, which is then run just before building the docs.


elliottt (approved):
:ship:

PR #972, opened 2024-09-13
Update to weval 0.3.2.

cfallin:
This update pulls in the latest changes, mainly revolving around weval's move to the Bytecode Alliance as a Hosted Project, in its new repository at [bytecodealliance/weval](https://github.com/bytecodealliance/weval).

### DIFF
--- a/integration-tests/js-compute/fixtures/app/src/compute.js
+++ b/integration-tests/js-compute/fixtures/app/src/compute.js
@@ -8,7 +8,7 @@ routes.set('/compute/get-vcpu-ms', () => {
   ok(cpuTime > 0);
   ok(cpuTime < 3000);
   const arr = [];
-  for (let i = 0; i < 10_000; i++) {
+  for (let i = 0; i < 1_000_000; i++) {
     arr.push(i);
   }
   const cpuTime2 = vCpuTime();
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@bytecodealliance/jco": "^1.5.0",
+        "@bytecodealliance/weval": "^0.3.2",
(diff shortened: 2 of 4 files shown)


cfallin:
@guybedford the `get-vcpu-ms` test is failing under weval [here](https://github.com/fastly/js-compute-runtime/actions/runs/10856556928/job/30131641655?pr=972); do you think we could update the logic to accept "arbitrarily fast" execution (i.e., [here](https://github.com/fastly/js-compute-runtime/blob/0dcff5831745053cbb6ec687a6d5435860126fc8/integration-tests/js-compute/fixtures/app/src/compute.js#L15-L16), accept the case where `cpuTime2 == cpuTime`) so it's not as flaky?

guybedford:
@cfallin it feels to me like we can only say we've tested the host call if we can verify the progression of time. Therefore I'd rather extend the test to write more performance intensive code to ensure we do use vcpu time, if you're okay with that approach?

guybedford:
That said, it is unfortunate this hostcall is in milliseconds and not microseconds.

cfallin:
Sure -- just upped the loop count again; let's see if this works!

cfallin:
Hmm, now it's flaking in the debug build. I'm still not sure that adjusting the knob *just right* to get runtime right for the test won't bite us later; timing-dependent tests in general are harder to keep working. Perhaps we can mock the vCPU-time passage somehow (test endpoint or switch in Viceroy to increment ms on every hostcall or something like that)?

guybedford:
I've added a change for this in https://github.com/fastly/js-compute-runtime/pull/971 which I'll land shortly so the rebase should no longer be flaky.

So for now, just revert the vcpu test change here and then as soon as we rebase this PR it should be okay.

Will aim to land and release still this afternoon if I can.

guybedford (approved):

PR #976, opened 2024-09-17
Do not panic on missing geo data

guybedford:
When running in Viceroy, missing geo data causes `len = 0` returns or `none` returns depending on the exact scenario, both of which are currently panics.

This resolves that in returning `null` when no geo data is available. ExecuteD always returns geo data, even if it is mostly stub data, so the test branches on the environment to handle both sides of this correctly.

### DIFF
--- a/integration-tests/js-compute/fixtures/app/src/geoip.js
+++ b/integration-tests/js-compute/fixtures/app/src/geoip.js
@@ -1,6 +1,6 @@
 /* eslint-env serviceworker */
 /* global fastly */
-import { assert, assertThrows } from './assertions.js';
+import { assert, assertThrows, strictEqual } from './assertions.js';
 import { getGeolocationForIpAddress } from 'fastly:geolocation';
 import { isRunningLocally, routes } from './routes.js';
 
@@ -127,9 +127,9 @@ routes.set(
 );
 
 let geoFields = [
-  'area_code',
   'as_name',
   'as_number',
+  'area_code',
   'city',
   'conn_speed',
   'conn_type',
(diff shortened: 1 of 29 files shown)


integration-tests/js-compute/fixtures/app/src/geoip.js
```diff
@@ -148,11 +148,29 @@ let geoFields = [
...
+  }
+  assertThrows(() => {
+    fastly.getGeolocationForIpAddress('999.999.999.999');
+  }, Error);
+});
+
 routes.set(
   '/fastly/getgeolocationforipaddress/parameter-ipv4-string',
   async () => {
+    let geo = fastly.getGeolocationForIpAddress('1.9.2.1');
```
cfallin, line 170: WHOIS shows `1.9.2.1` as part of an allocated network range:

```
inetnum:        1.9.2.0 - 1.9.2.255
netname:        MYREN-TMNET
descr:          Kementerian Pengajian Tinggi Malaysia
country:        MY
admin-c:        KK753-AP
tech-c:         KK753-AP
abuse-c:        AM2355-AP
status:         ASSIGNED NON-PORTABLE
mnt-by:         MAINT-MY-WEBE
mnt-irt:        IRT-MY-WEBE
last-modified:  2023-06-30T00:01:26Z
source:         APNIC
[ ... ]
```

Is the idea that we expect valid data for this address (so we test the `keys` on the object below)? If so, could we pick an address we control, or one that is well-known, to avoid any potential test flakes in the future? E.g. either a Fastly IP (`151.101.1.1` as the first IP in our main /16) or Google's DNS (`8.8.8.8`) or something?

cfallin: (Should match the assert error message below, and likewise same question for the IPv6 address below, which appears to be in a Softlayer IP range)

cfallin (approved):
Host API implementation details look good; some thoughts on the IPs used in the tests but otherwise LGTM 👍

integration-tests/js-compute/fixtures/app/src/geoip.js
guybedford: Good suggestion - I've gone with Fastly's IP here instead.

Issue #753, opened 2024-04-10
Reading some fields of FetchEvent.client causes a crash in Viceroy in js-compute >= 3.0


harmony7:
`FetchEvent.client` is of type `ClientInfo`, which contains information about the currently connected client.

One thing to note is, these crashes do not occur when deployed to Fastly Compute.

Reading some of these fields in Viceroy (including fiddle) causes crashes, starting with js-compute@3.x (the problem does not happen in js-compute@2.x): `'tlsJA3MD5', 'tlsCipherOpensslName', 'tlsProtocol', 'tlsClientCertificate', 'tlsClientHello'`.

The error I get is something like this:
```
Error while running request handler: tls_cipher_openssl_name_get: A `None` error. This status code is used to indicate when an optional value did not exist, as opposed to an empty value.
```

> NOTE: In Fiddle the error looks like this:
> ```
> tls_cipher_openssl_name_get: Generic error value. This means that some unexpected error occurred during a hostcall.
> ```

This means that simply trying to perform a `console.log(event.client)` in local development will cause a crash.

However, these crashes do not occur when deployed to a Fastly service. The reason for this I'm speculating may be because these fields always have values on Fastly.

I noticed this because a library I had written back in js-compute@2 started crashing in a debug logging statement after switching to js-compute@3, but only on local dev.

I have a repro here: https://fiddle.fastly.dev/fiddle/adac0871

![image](https://github.com/fastly/js-compute-runtime/assets/1396801/72395a5f-07ca-4e79-8e30-c5b1d4abae6a)

![image](https://github.com/fastly/js-compute-runtime/assets/1396801/0aa1444f-2037-431c-9ba9-7fa46514c81e)

JakeChampion:
Thanks for this.

We should probably return `undefined` if these host-calls return a None error.
Should we return `undefined` for any other type of error that the host-calls return?

@guybedford what do you think?

guybedford:
Perhaps we can have a more general fallback for all `None` errors like `No value returned for host call ...`?

But then for these specifically, returning `undefined` makes sense. I don't know which XQD functions may possibly return None, it's a problem of the typing overall, but a fallback none error might at least help catch those cases further.

guybedford:
I've gone ahead with the `null` return implementation here in https://github.com/fastly/js-compute-runtime/pull/979.

This means not getting an error anymore, but does require more careful type checks going forward.

PR #979, opened 2024-09-17
fix: never error for missing client data, return null instead

guybedford:
This is a follow-on to https://github.com/fastly/js-compute-runtime/pull/976 (and is based to), that resolves https://github.com/fastly/js-compute-runtime/issues/753 in always returning `null` instead of throwing an error for missing client data.

Documentation updates are included as well including a note that the null return is for Viceroy / testing environments.


cfallin (approved):
Second commit LGTM

Issue #977, opened 2024-09-17
TCP and HTTP keepalive parameters on dynamic backends


guybedford:
Per https://docs.rs/fastly/0.10.5/fastly/backend/struct.BackendBuilder.html.

guybedford:
Landed in #993.

PR #993, opened 2024-10-02
feat: setDefaultDynamicBackendConfig with all backend configuration options

### LINKED ISSUE #970: Support all backend configuration option defaults for dynamic backends
Currently the `enableDynamicBackends` call allows providing default configuration options for dynamic backends.

These defaults are restricted to only the three timeout values currently.

It would be beneficial to extend these defaults to all configuration options that make sense across all backend constructor calls generally.

Another way to frame this might even be as default construction options for all backends possibly even as fallbacks for options not provided to `new Backend()`. Under such a framing it might become `import { setDefaultBackendOptions } from 'fastly:backend'` say.

This would align well with eventually deprecating `enableDynamicBackends()` entirely, which could nice to get going for the next major release, to rather give a nicer error when dynamic backends are disabled.

guybedford:
This adds a new `import { setDefaultDynamicBackendConfig } from 'fastly:backend'`, which allows defining default values for all appropriate dynamic backend configurations.

These defaults then get applied to both automatically generated dynamic backends when using fetch, as well as the defaults to use when explicitly performing `new Backend(...)`.

In addition the new `httpKeepalive` and `tcpKeepalive` backend options are implemented and also supported in this default mechanism.

Resolves #970, #977.


runtime/fastly/host-api/host_api_fastly.h
```diff
@@ -360,6 +366,74 @@ struct BackendConfig {
   std::optional<HostString> sni_hostname;
   std::optional<ClientCert> client_cert;
   std::optional<bool> grpc;
+  std::optional<uint32_t> http_keepalive_time_ms;
+  std::optional<TcpKeepalive> tcp_keepalive;
+
+  BackendConfig clone() {
```
cfallin, line 372: It's too bad this code has to be written manually -- it looks like the struct could *almost* use the default copy constructor except that `HostString` deletes its copy-constructor. I don't know enough about modern C++ idioms to say why but this might be something to see about?

Alternately, perhaps we could write an initializer-list expression (`return BackendConfig { a, b, c, ... }`) and use a helper to clone `std::optional<HostString>` so that at least we'll get a compile error if we add a field and forget to update this?

cfallin (approved):
Seems reasonable to me overall! A few thoughts below but nothing major.

runtime/fastly/builtins/backend.cpp
```diff
@@ -1553,18 +1144,389 @@ bool Backend::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  }
+  std::string_view target_string(reinterpret_cast<char *>(target_string_slice.data),
+                                 target_string_slice.len);
+  if (!validate_target(cx, target_string)) {
+    return nullptr;
+  }
+  const jsurl::SpecString origin_specstring = jsurl::origin(url);
+  std::string_view origin((char *)origin_specstring.data, origin_specstring.len);
+
+  auto use_ssl = origin.rfind("https://", 0) == 0;
```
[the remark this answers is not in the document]
guybedford: This will be a parsed URL since it is originally extracted from the request internals.

runtime/fastly/host-api/host_api_fastly.h
guybedford: I believe we disabled the copy constructor to ensure a consistent ownership model for host strings. Good suggestion wrt the initializer, I've gone ahead and added that.

PR #1006, opened 2024-10-11
fix!: immutable headers guards for incoming response and downstream request

### LINKED ISSUE #992: Backend response headers set issue
We had a report that this was causing duplication when using `beresp.headers.set()`.

We should ensure we either properly guard or fix this case.

guybedford:
This implements immutable headers guards for the fetch event request, and any `fetch()` responses.

With this change, trying to set a header on one of these will throw a TypeError as is correct per the fetch and service worker specs.

Resolves https://github.com/fastly/js-compute-runtime/issues/992.


guybedford:
I will wait to land this until we make the next major release.

guybedford:
This should also reintroduce the `"fetch/api/response/response-headers-guard.any.js"` WPT test.

guybedford:
This can be done as a minor since there is no functional change, so long as it rebased in a way that respects the Candidate Response phase of the Response permitting mutations (i.e. the definition of is_upstream needs to be tweaked, and the guard needs to change during the lifecycle).

dgohman-fastly:
@cfallin I've now removed you as a reviewer.

PR #1008, opened 2024-10-15
perf: use -o2 optimization for starlingmonkey

guybedford:
We're in the process of investigating performance differences on StarlingMonkey, and this is one obvious difference between js-compute-runtime and StarlingMonkey that might be worth starting with.

I'd like to roll this out into the 3.24 release today to test it out more widely, and then use that information to inform further investigation here.


guybedford:
Closing, as we have another solution here for investigation.

Issue #991, opened 2024-10-01
HTTP Cache API: JS SDK


guybedford:
This is to track the JS SDK additions in Q4 to support HTTP Cache API

abharadwaj999:
Can we have this by EoY (Q4) as we do have JS customers waiting on HTTP Cache API support ?

guybedford:
Hostcall parity is exactly the Q4 goal. I haven't fully analyzed the surface area of the new HTTP Cache API yet, and KVStore v2 is next on the list, but it seems fairly likely we should be able to complete this in Q4.

abharadwaj999:
apologies, but just for my understanding, does "hostcall parity" equate to "JS SDK availability in guest for HTTP Cache API " or something else host specific only?

guybedford:
Yes we want to fully support the latest host APIs in the JS SDK.

PR #1018, opened 2024-10-22
DO NOT MERGE YET - HTTP Caching API

harmony7:
This PR tracks the implementation of the HTTP caching API.

Implements https://github.com/fastly/js-compute-runtime/issues/991

### DIFF
--- a/types/cache-override.d.ts
+++ b/types/cache-override.d.ts
@@ -98,6 +98,20 @@ declare module 'fastly:cache-override' {
      *
      * See the [Fastly PCI-Compliant Caching and Delivery documentation](https://docs.fastly.com/products/pci-compliant-caching-and-delivery)
      * for details.
+     *
+     * @param {void} [init.onBeforeSend]
+     * Set a [callback function](https://developer.mozilla.org/en-US/docs/Glossary/Callback_function) to be invoked if a
+     * request is going all the way to a backend, allowing the request to be modified beforehand.
+     *
+     * See [Modifying a request as it is forwarded to a backend](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/#modifying-a-request-as-it-is-forwarded-to-a-backend)
+     * in the Fastly cache interfaces documentation for details.
+     *
+     * @param {void} [init.onAfterSend]
+     * Set a [callback function](https://developer.mozilla.org/en-US/docs/Glossary/Callback_function) to be invoked after
+     * a response has been sent, but before it is stored into the cache.
+     *
+     * See [Controlling cache behavior based on backend response](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/#controlling-cache-behavior-based-on-backend-response)
+     * in the Fastly cache interfaces documentation for details.
      */
     constructor(
       mode: 'none' | 'pass' | 'override',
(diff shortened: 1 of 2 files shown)


harmony7:
Starting by committing proposed TypeScript types for the API surface.

guybedford:
I've started to put together a variation of this in https://github.com/fastly/js-compute-runtime/pull/1051. Using Response mutations instead of having a separate CandidateResponse, and also using a CacheOptions return.

Here are the same examples updated to this API.

Depending on how the implementation goes I will update this.

```javascript
// Example: inject headers before sending.
const request = event.request;
const response = await fetch(request, {
  backend: 'example',
  cacheOverride: new CacheOverride('override', {
    beforeSend(req) {
      // Assume buildAuthorizationHeader() may be expensive, so
      // we only want to call this when the request would actually reach the backend.
      req.headers.set('Authorization', buildAuthorizationHeader());
    },
  }),
});
```

```javascript
// Example: customizing caching based on content type.
const request = event.request;
const response = await fetch(request, {
  backend: 'example',
  cacheOverride: new CacheOverride('override', {
    beforeCache(resp) {
      const contentType = resp.headers.get('Content-Type') ?? '';
      switch(true) {
        case contentType.startsWith('image/'):
          resp.ttl = 67;
          break;
        case contentType === 'text/html':
          resp.ttl = 321;
          break;
        case contentType === 'application/json':
          // setUncacheable becomes returning { cache: false }
          return { cache: false };
          break;
        default:
          resp.ttl = 2;
      }
    },
  }),
});
```

```javascript
// Example: creating a hit-for-pass object.
const request = event.request;
const response = await fetch(request, {
  backend: 'example',
  cacheOverride: new CacheOverride('override', {
    afterSend(resp) {
      if (resp.headers.has('my-private-header')) {
        // setUncacheable() becomes 'uncacheable'
        return { cache: 'uncacheable' };
      }
    },
  }),
});
```

```javascript
// Example: expanding a template before caching.
const request = event.request;
const response = await fetch(request, {
  backend: 'example',
  cacheOverride: new CacheOverride('override', {
    beforeCache(resp) {
      resp.headers.set('Content-Type', 'text/html');
      const bodyTransform = new TransformStream({
        bytes: null,
        start() {
          this.bytes = new Uint8Array(0);
        },
        transform(chunk) {
          // The ideal transform would transform bytes as it goes over the wire,
          // but for a transformation whose input is JSON, we need to buffer the
          // bytes to memory, because we need the whole body before we can
          // deserialize it.
          const newBytes = new Uint8Array(this.bytes.length + chunk.length);
          newBytes.set(this.bytes, 0);
          newBytes.set(chunk, this.bytes.length);
          this.bytes = newBytes;
        },
        flush(controller) {
          const str = new TextDecoder().decode(this.bytes);
          // jsonToHtml applies a template to generate HTML from JSON
          const html = jsonToHtml(str);
          controller.enqueue(new TextEncoder().encode(html));
        },
      });
      return { bodyTransform };
    },
  }),
});

// The resulting cached object will have a body that is HTML
// despite that the backend returned a JSON response:
return response;
```

harmony7:
I am liking the shape of this.

I'm have one question.

One of the reasons my previous suggested API surface has a CandidateResponse type was to provide an interface that specifically does not provide a way to access certain parts of the Response (mainly the Body) during after-send. For example, if you consume Body here then the caller of fetch will be in trouble.  Also, revalidation responses will have an empty Body during after-send, to be filled in by the cached body before returning from fetch.

If you use a normal Response here, will you be able to enforce this?

guybedford:
Thanks for bringing that up. We could add a state bit to the response that treats it as a locked response somehow I'm sure, and to give an error on access to the body or something similar.

guybedford:
Closing as the final design is now fully tracking in https://github.com/fastly/js-compute-runtime/pull/1051. Any final comments with regards to the API surface area very much welcome.

PR #1021, opened 2024-10-22
feat: module mode

guybedford:
This introduces a new flag for `--module-mode` which disables bundling and runs the modules directly in StarlingMonkey.

This is especially useful for debugging workflows where getting a nice stack of the exact JS sources is very much needed, and I actually just implemented this myself exactly to make a debugging case easier.

In module mode, a new `--bundle` flag is also support to still bundle in module mode if desired, which then exactly corresponds to the previous top-level await support which is also then implemented as such and that flag is now deprecated and removed from docs for using this technique instead.

`--bundle` can be kept along on a 3.x timeframe, but should also then get removed on the 4.x path.

Taking module mode out of experimental is dependent on the StarlingMonkey issues https://github.com/bytecodealliance/StarlingMonkey/issues/168 and https://github.com/bytecodealliance/StarlingMonkey/issues/15.


src/printHelp.js
```diff
@@ -14,9 +14,10 @@ FLAGS:
 
 OPTIONS:
     --engine-wasm <engine-wasm>                             The JS engine Wasm file path
+    --module-mode                            [experimental] Run all sources as native modules,
+                                                            with full error stack support.
     --enable-experimental-aot                               Enable experimental AOT compilation for performance
     --enable-experimental-high-resolution-time-methods      Enable experimental high-resolution fastly.now() method
```
guybedford, line 20:
```suggestion
    --enable-experimental-high-resolution-time-methods      Enable experimental high-resolution fastly.now() method
    --enable-experimental-top-level-await                   Enable experimental top level await
```

integration-tests/cli/help.test.js
```diff
@@ -57,9 +58,10 @@ test('-h should return help on stdout and zero exit code', async function (t) {
     '-V, --version                                           Prints version information',
     'OPTIONS:',
     '--engine-wasm <engine-wasm>                             The JS engine Wasm file path',
+    '--module-mode                            [experimental] Run all sources as native modules,',
+    'with full error stack support.',
     '--enable-experimental-aot                               Enable experimental AOT compilation for performance',
     '--enable-experimental-high-resolution-time-methods      Enable experimental high-resolution fastly.now() method',
```
guybedford, line 64:
```suggestion
    '--enable-experimental-high-resolution-time-methods      Enable experimental high-resolution fastly.now() method',
    '--enable-experimental-top-level-await                   Enable experimental top level await',
```

integration-tests/cli/help.test.js
```diff
@@ -30,9 +30,10 @@ test('--help should return help on stdout and zero exit code', async function (t
...
```
guybedford, line 36:
```suggestion
    '--enable-experimental-high-resolution-time-methods      Enable experimental high-resolution fastly.now() method',
    '--enable-experimental-top-level-await                   Enable experimental top level await',
```

Issue #1063, opened 2024-12-04
Option to disable log prefixing


guybedford:
Currently all log messages are prefixed with `Log: ` - it could be worth providing an option to disable this prefixing.

jsocol:
I have an app using `js-compute-runtime` that's trying to output JSON-formatted logs, and parsing is broken by these prefixes. In some cases there are other places to handle it, but since this is atypical behavior for `console`, it'd be great to be able to switch it to behave as it would in Node or tests

guybedford:
Added in https://github.com/fastly/js-compute-runtime/pull/1065.

PR #1065, opened 2024-12-05
feat: configureConsole function for configuring log prefixing and stderr

guybedford:
Adds a new `configureConsole` method to `fastly:logging` to enable configuration of the `console` logging for removing prefixing and supporting stderr output.

In future, it might be nice to support routing console logs to logging endpoints here as well, but that is out of scope for this initial PR.


guybedford:
Resolves https://github.com/fastly/js-compute-runtime/issues/1063.

PR #1079, opened 2025-01-07
Fanout: Use redirect_to_grip_proxy_v2 and pass request handle

harmony7:
This PR updates `createFanoutHandoff()` to call `redirect_to_grip_proxy_v2` instead of `redirect_to_grip_proxy`, passing the Request handle to it.

### DIFF
--- a/runtime/fastly/builtins/fetch-event.cpp
+++ b/runtime/fastly/builtins/fetch-event.cpp
@@ -668,12 +668,10 @@ bool response_promise_then_handler(JSContext *cx, JS::HandleObject event, JS::Ha
       return false;
   }
 
-  bool streaming = false;
-  if (Response::is_grip_upgrade(response_obj)) {
-    auto grip_upgrade_request = Response::grip_upgrade_request(response_obj);
+  if (auto grip_upgrade_request = Response::grip_upgrade_request(response_obj)) {
     auto backend = Response::backend_str(cx, response_obj);
 
-    auto res = grip_upgrade_request.redirect_to_grip_proxy(backend);
+    auto res = grip_upgrade_request->redirect_to_grip_proxy(backend);
     if (auto *err = res.to_err()) {
       HANDLE_ERROR(cx, *err);
       return false;
@@ -681,6 +679,7 @@ bool response_promise_then_handler(JSContext *cx, JS::HandleObject event, JS::Ha
     return true;
   }
 
+  bool streaming = false;
   if (!RequestOrResponse::maybe_stream_body(cx, response_obj, &streaming)) {
     return false;
   }
--- a/runtime/fastly/builtins/fetch-event.cpp
+++ b/runtime/fastly/builtins/fetch-event.cpp
@@ -670,9 +670,10 @@ bool response_promise_then_handler(JSContext *cx, JS::HandleObject event, JS::Ha
 
   bool streaming = false;
   if (Response::is_grip_upgrade(response_obj)) {
+    auto grip_upgrade_request = Response::grip_upgrade_request(response_obj);
     auto backend = Response::backend_str(cx, response_obj);
 
-    auto res = host_api::HttpReq::redirect_to_grip_proxy(backend);
+    auto res = grip_upgrade_request.redirect_to_grip_proxy(backend);
     if (auto *err = res.to_err()) {
       HANDLE_ERROR(cx, *err);
       return false;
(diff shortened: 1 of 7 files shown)


harmony7:
There are no updates to tests at the current moment, as I'm not able to find any exiting tests to redirect_to_grip_proxy. If there is a way to add tests for this I'd like help.

runtime/fastly/builtins/fetch-event.cpp
```diff
@@ -670,9 +670,10 @@ bool response_promise_then_handler(JSContext *cx, JS::HandleObject event, JS::Ha
 
   bool streaming = false;
   if (Response::is_grip_upgrade(response_obj)) {
+    auto grip_upgrade_request = Response::grip_upgrade_request(response_obj);
```
guybedford, line 673: So all this work can actually be simplified since we do have the request handle already available in this function via the Request slot on the fetch event itself:

```
JS::RootedObject request(
      cx, &JS::GetReservedSlot(self, static_cast<uint32_t>(Slots::Request)).toObject());
MOZ_ASSERT(request);
auto req_handle = Request::request_handle(request);
MOZ_ASSERT(req_handle.is_valid());
```

guybedford (changes requested):
Thanks for the PR! I think we can avoid the need for the extra book-keeping though, added a suggested to that effect.

guybedford:
Also do the tests at `integration-tests/js-compute/fixtures/app/src/fanout.js` not already provide some coverage here?

runtime/fastly/builtins/fetch-event.cpp
harmony7: Thanks for the comment!

The request handle passed in to `redirect_to_grip_proxy_v2()` doesn't necessarily have to be the request handle attached to the fetch event. This way you can use modified headers and/or url. So I want to save off the request that was passed in as the first parameter of `createFanoutHandoff()` and pass that.

harmony7:
> Also do the tests at `integration-tests/js-compute/fixtures/app/src/fanout.js` not already provide some coverage here?

Aha thanks! Yes I see them. Unfortunately, looks like it's hard to test the effects of this PR.

runtime/fastly/builtins/fetch-event.cpp
guybedford: Ahh right, sorry I traced the wrong Response::create here!

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -2285,7 +2284,13 @@ bool Response::is_upstream(JSObject *obj) {
 
 bool Response::is_grip_upgrade(JSObject *obj) {
   MOZ_ASSERT(is_instance(obj));
-  return JS::GetReservedSlot(obj, static_cast<uint32_t>(Slots::IsGripUpgrade)).toBoolean();
+  return JS::GetReservedSlot(obj, static_cast<uint32_t>(Slots::GripUpgradeRequest)).toInt32() != 0;
+}
+
+host_api::HttpReq Response::grip_upgrade_request(JSObject *obj) {
```
guybedford, line 2290: You could combine these two calls into a single `std::option<host_api::HttpReq> Response::grip_upgrade_request(JSObject *obj)` instead of needing two separate function calls.

If changing to support UndefinedValue per the next comment, then you'd also need to use an `isUndefined()` for this.

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -3196,8 +3202,10 @@ JSObject *Response::create(JSContext *cx, JS::HandleObject response,
   JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::Redirected), JS::FalseValue());
   JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::IsUpstream),
                       JS::BooleanValue(is_upstream));
-  JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::IsGripUpgrade),
-                      JS::BooleanValue(is_grip));
+  JS::SetReservedSlot(response, static_cast<uint32_t>(Slots::GripUpgradeRequest),
+                      JS::Int32Value(grip_upgrade_request
+                                         ? Request::request_handle(grip_upgrade_request).handle
+                                         : 0));
```
guybedford, line 3208: Better to use `UndefinedValue` for the empty case:

```
grip_upgrade_request ? JS::Int32Value(Request::request_handle(grip_upgrade_request).handle) : JS::UndefinedValue()
```

guybedford (commented):
Thanks for clarifying, a couple of points from code review, but otherwise looks great to land!

guybedford (approved):

PR #1097, opened 2025-01-28
Fix forbidden KV key special characters to match documentation

btsheehy:
Related:
https://github.com/fastly/cli/issues/1383
https://github.com/fastly/Viceroy/pull/447

Currently the Compute Javascript SDK enforces an incorrect set of special characters that are forbidden in KV key names. The current behavior is forbidding `[`, `]`, `*`, `?`, `#`, while the [documentation](https://docs.fastly.com/en/guides/working-with-kv-stores#limitations-and-considerations) lists `;`, `^`, `|`, `#`, and `?`. Square brackets and asterisks are specifically called out as allowed characters. 

I'm unfamiliar with this codebase, so it's entirely possible I've missed some files that need to be updated.

### DIFF
--- a/documentation/docs/kv-store/KVStore/prototype/delete.mdx
+++ b/documentation/docs/kv-store/KVStore/prototype/delete.mdx
@@ -29,7 +29,7 @@ Returns `undefined`
   - If the provided `key`:
     - Is any of the strings `""`, `"."`, or `".."`
     - Starts with the string `".well-known/acme-challenge/"`
-    - Contains any of the characters `"#?*[]\n\r"`
+    - Contains any of the characters `"#;?^|\n\r"`
     - Is longer than 1024 characters
     - Does not exist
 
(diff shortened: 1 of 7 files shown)


guybedford:
Thanks for posting. The other places we need to handle this are:

* The implementation at https://github.com/fastly/js-compute-runtime/blob/main/runtime/fastly/builtins/kv-store.cpp#L199
* The types file at https://github.com/fastly/js-compute-runtime/blob/main/types/kv-store.d.ts
* The documentation file at https://github.com/fastly/js-compute-runtime/blob/main/documentation/docs/kv-store/KVStore/KVStore.mdx

Let me know if you're able to update this PR to include those, or if I can assist further with this.

btsheehy:
> Thanks for posting. The other places we need to handle this are:
> 
> * The implementation at https://github.com/fastly/js-compute-runtime/blob/main/runtime/fastly/builtins/kv-store.cpp#L199
> * The types file at https://github.com/fastly/js-compute-runtime/blob/main/types/kv-store.d.ts
> * The documentation file at https://github.com/fastly/js-compute-runtime/blob/main/documentation/docs/kv-store/KVStore/KVStore.mdx
> 
> Let me know if you're able to update this PR to include those, or if I can assist further with this.

I believe my original commit handled the first one, but I just committed a change to update the types files and any doc files I could find containing the error

guybedford:
Not sure why this isn't running CI correctly, but the PR looks good to me, will land and we can follow up if there are further issues.

Issue #1029, opened 2024-10-31
KV store generation integer


guybedford:
We still don't support the generation integer in KV Store lookup and insert APIs, due to the bug fix that is still releasing in Viceroy for this. When that goes out we should update the API to expose this.

guybedford:
The generation integer must also be updated to be a U64.

computermouth:
Started a draft PR here: https://github.com/fastly/js-compute-runtime/pull/1069

guybedford:
Reopening to continue tracking the support for the generation integer in calls other than lookup.

computermouth:
Oh my bad, sorry

kpfleming:
Support is currently available for the `put` operation (upsert), but not for `delete`. I was planning to open a new issue for that, but this one will suffice with a summary change.

PR #1108, opened 2025-02-05
KV store generation integer follow up (#1029)

vcarvajal-sigsci:
follow up to https://github.com/fastly/js-compute-runtime/issues/1029

### DIFF
--- a/integration-tests/js-compute/setup.js
+++ b/integration-tests/js-compute/setup.js
@@ -11,6 +11,9 @@ const { DICTIONARY_NAME, CONFIG_STORE_NAME, KV_STORE_NAME, SECRET_STORE_NAME } =
   getEnv(serviceName);
 
 function existingStoreId(stores, existingName) {
+  if (stores == null) {
+    return false;
+  }
   const existing = stores.find(
     ({ Name, name }) => name === existingName || Name === existingName,
   );
--- a/integration-tests/js-compute/setup.js
+++ b/integration-tests/js-compute/setup.js
@@ -11,9 +11,6 @@ const { DICTIONARY_NAME, CONFIG_STORE_NAME, KV_STORE_NAME, SECRET_STORE_NAME } =
   getEnv(serviceName);
 
 function existingStoreId(stores, existingName) {
-  if (stores == null) {
-    return false;
-  }
   const existing = stores.find(
     ({ Name, name }) => name === existingName || Name === existingName,
   );
@@ -116,9 +113,14 @@ async function setupKVStore() {
 async function setupSecretStore() {
   let stores = await (async function () {
     try {
-      return JSON.parse(
+      res = JSON.parse(
         await zx`fastly secret-store list --quiet --json --token $FASTLY_API_TOKEN`,
       );
+      if (res == null) {
+        return [];
+      } else {
+        return res;
+      }
     } catch {
       return [];
     }
--- a/runtime/fastly/builtins/kv-store.cpp
+++ b/runtime/fastly/builtins/kv-store.cpp
@@ -482,13 +482,12 @@ bool KVStore::put(JSContext *cx, unsigned argc, JS::Value *vp) {
 
     if (!gen_val.isNullOrUndefined()) {
       if (gen_val.isNumber()) {
-          if (gen_val.toInt32()) {
-            if_gen.emplace(gen_val.toInt32());
+          if (gen_val.toNumber()) {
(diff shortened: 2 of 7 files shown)


integration-tests/js-compute/setup.js
```diff
@@ -11,6 +11,9 @@ const { DICTIONARY_NAME, CONFIG_STORE_NAME, KV_STORE_NAME, SECRET_STORE_NAME } =
   getEnv(serviceName);
 
 function existingStoreId(stores, existingName) {
```
vcarvajal-sigsci, line 13: in setting up my fastly [secret-store](https://github.com/fastly/js-compute-runtime/blob/main/integration-tests/js-compute/setup.js#L113), it got into a state where it was null and throw an exception.  Not sure if this is necessary or just something I encountered while trying to setup the integration tests.

runtime/fastly/builtins/kv-store.cpp
```diff
@@ -474,6 +475,24 @@ bool KVStore::put(JSContext *cx, unsigned argc, JS::Value *vp) {
       ttl = parsed;
     }
 
+    JS::RootedValue gen_val(cx);
+    if (!JS_GetProperty(cx, opts_val, "gen", &gen_val)) {
+      return false;
+    }
+
```
vcarvajal-sigsci, line 482: the generation match integer comes across as a uint64_t, however I'm not converting it to [uint64](https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/file.cpp#L16).  I noticed some functions in spidermonkey that had a comment of it being non-standard, could/should I be using those?

guybedford: Adding the support for BigInt can be useful for 64 bit integers effectively supporting `gen: 1000n` as well as `gen: 1000` say. The BigInt functions are exactly what you want to use and there should be existing examples in the codebase for this as well.

integration-tests/js-compute/setup.js
guybedford: Strange, it would be interesting to dig into the caller of this function, to see what that is happening and potentially resolve the issue at the root.

runtime/fastly/host-api/error_numbers.msg
```diff
@@ -65,6 +65,7 @@ MSG_DEF(JSMSG_KV_STORE_KEY_ACME,                               0, JSEXN_TYPEERR,
 MSG_DEF(JSMSG_KV_STORE_KEY_RELATIVE,                           0, JSEXN_TYPEERR, "KVStore key can not be '.' or '..'")
 MSG_DEF(JSMSG_KV_STORE_PUT_CONTENT_STREAM,                     0, JSEXN_TYPEERR, "Content-provided streams are not yet supported for streaming into KVStore")
 MSG_DEF(JSMSG_KV_STORE_PUT_OVER_30_MB,                         0, JSEXN_TYPEERR, "KVStore value can not be more than 30 Megabytes in size")
+MSG_DEF(JSMSG_KV_STORE_PUT_OPTIONS,                            1, JSEXN_TYPEERR, "KVStore insert option: {0}")
```
guybedford, line 68: This would throw the error `KVStore insert options: if generation match`. Maybe we can refine this to be the generic `api_error`, which StarlingMonkey has been migrating to with a nicer message like: `KVStore insert: "gen" must be a number` or something like that.

guybedford (commented):
Looks like a great start!

Would be worth adding a test for some of the error cases.

There's also the `docs/documentation/kv-store` documentation to update as well.

runtime/fastly/host-api/error_numbers.msg
vcarvajal-sigsci: good idea, i changed this in 8e0616f99a3d6b9426ad675b085005cdd960c50c

runtime/fastly/builtins/kv-store.cpp
vcarvajal-sigsci: I couldn't find a good example of converting a `RootedBigInt` and converting that to a uint64_t, so I left it as a `number`

vcarvajal-sigsci:
> Looks like a great start!
> 
> Would be worth adding a test for some of the error cases.
> 
> There's also the `docs/documentation/kv-store` documentation to update as well.

Thanks I updated the related documentation, I'd appreciate your input on the wording.  I'll look to add more error cases.

guybedford (approved):

Issue #1110, opened 2025-02-10
Ensure secret store cleanup


guybedford:
We don't seem to be properly disposing the secret stores created in the integration tests, and as a result we regularly hit the maximum number of secret stores limit.

Fixing this will improving CI reliability, instead of having to manually delete these.

PR #1112, opened 2025-02-12
Secret store cleanup

vcarvajal-sigsci:
https://github.com/fastly/js-compute-runtime/issues/1110

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -446,11 +446,11 @@ jobs:
       run: npm install && cd ./integration-tests/js-compute && npm install
 
     - name: Run Tests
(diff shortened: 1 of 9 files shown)


vcarvajal-sigsci:
> I'm not sure this is where the issue is - these skips are just because we are running two test suites side by side, so we only need to do setup once, and teardown once either side of those runs.
> 
> Rather I think the issue is within the teardown script itself that it is not disposing the secret store properly, which needs to be tested against the Fastly API.

The original error I was encountering was that we were exceeding our number of stores because we kept creating them
`
Output [Error]: ERROR: the Fastly API returned 400 Bad Request: Bad Request (limit exceeded: number of stores).
`
But now after creating the secret store, we're trying to retrieve the store id, we are getting a 404 stating the record isn't found.
`
 Creating new config store aZ2 __ 2__app_secret_store_cleanup__debug
  ERROR: the Fastly API returned 404 Not Found: Record not found.
`

But good point about further testing against the Fastly API, I'll continue on that path.

PR #1118, opened 2025-02-13
initial changes for compute workflows

vcarvajal-sigsci:


guybedford:
Some of this may be duplicated with the recent changes made in the Acl hostcall work PR. Could be worth rebasing, we just need to make sure we land with the debugging adapters disabled.

vcarvajal-sigsci:
actually i think it's worth deleting this branch! :)

PR #1129, opened 2025-02-26
fix: ready-based immediate task indexing

guybedford:
This fixes a bug found by @harmony7 in https://github.com/harmony7/compute-react19-readablestream-repro where a specific setup in async tasks with immediate timers would lead to a panic.

The panic in this case turns out to be in the Fastly event loop layer, where I previously implemented a bug in the optimized immediate handle checking case in not properly looking up the immediate index. The logic fix is self-explanatory for the most part.

Confirmed that with this change the linked repo no longer panics.


guybedford:
CI is currently failing because the `fastly config-store create` command doesn't seem to be working currently.

Since this is critical fix I'm going ahead with the release based on local & viceroy tests passing and the changes for this release being fully isolated to this subsystem.

harmony7:
Thanks for this fix!

Issue #1155, opened 2025-03-21
Missing "getSetCookie" Method on Headers Type Definition


dustinknudson-target:
Update the [Headers interface definition](https://github.com/fastly/js-compute-runtime/blob/3ffbdb8c7ae231c53068378c324cd8599333ec8f/types/globals.d.ts#L1918-L1933) to include the `getSetCookie` method [that is now implemented](https://github.com/fastly/js-compute-runtime/releases/tag/v3.21.0).

vcarvajal-sigsci:
Hi @dustinknudson-target , thanks for bringing catching this! It's included in release 3.33.3

PR #1162, opened 2025-04-02
fix: TypeScript declaration for Headers (#1155)

vcarvajal-sigsci:
https://github.com/fastly/js-compute-runtime/issues/1155


harmony7 (approved):
Thank you for this!

PR #1174, opened 2025-04-11
Add some documentation for CA certificate validation

cceckman-at-fastly:
We've gotten various questions over time about the certificate validation behavior. It might not be obvious that:
- Backend certificates are always validated as rooted-at-the-CA...
- ...which consists of a set of default CAs, unless otherwise stated...
- ...but that doesn't include validating the hostname, unless you ask.

Add some doc comments to try to make this behavior more visible.

Go SDK version [here](https://github.com/fastly/compute-sdk-go/pull/161).

### DIFF
--- a/types/backend.d.ts
+++ b/types/backend.d.ts
@@ -66,6 +66,15 @@ declare module 'fastly:backend' {
     betweenBytesTimeout?: number;
     /**
      * Whether or not to require TLS for connections to this backend.
+     *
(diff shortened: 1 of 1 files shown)


vcarvajal-sigsci (approved):
LGTM

cceckman-at-fastly:
Unfortunately CI doesn't like the branch name here; will retry shortly.

cceckman-at-fastly:
Same content, different name: https://github.com/fastly/js-compute-runtime/pull/1176

PR #1181, opened 2025-06-02
feat(config): add support for config store buffers longer than 8k

zkat:


posborne (approved):

runtime/fastly/host-api/fastly.h
```diff
@@ -47,8 +47,8 @@ typedef fastly_host_http_response fastly_world_tuple2_handle_handle;
 #define HEADER_MAX_LEN 69000
 #define METHOD_MAX_LEN 1024
 #define URI_MAX_LEN 8192
-#define CONFIG_STORE_ENTRY_MAX_LEN 8000
-#define DICTIONARY_ENTRY_MAX_LEN CONFIG_STORE_ENTRY_MAX_LEN
+#define CONFIG_STORE_INITIAL_BUF_LEN 512
+#define DICTIONARY_ENTRY_MAX_LEN 8000
```
cceckman-at-fastly, line 51: It looks like this constant is still used.

(a) Is this properly 8000, or 8192 (i.e. 8k or 8Ki)?
(b) Does this `Dict` family of calls need the same treatment? Or are those no-ops now?

zkat: 1. 8000 is the legacy value, and I'm told Dict is a legacy-ish feature. I _believe_ part of the reason this change was made was precisely because 8000 is not 8192, and failures were happening when generating 8192-length hashes as keys /cc @awilliams-fastly 
2. Aside from the legacy-ish bit about `Dict`, I did not want to make this change to `Dict` &co because they have additional length limitations related to _hostcalls_: https://github.com/fastly/js-compute-runtime/pull/1181/files#diff-e0e662c1f6c8ef664ac3da563c24304a60d163a81666a487db34b98aee8f4e3dL53-L58 and I just wanted to keep things as narrowly scoped as I could in this case.

cceckman-at-fastly: Got it, thanks for the explanations!

cceckman-at-fastly (approved):

PR #1184, opened 2025-06-12
Option to respect range during cache streaming

cceckman-at-fastly:
The core cache API allows for cache items to be concurrently streamed to / from the cache.

If:
- The size of the cached item's body was not provided by the writer
- The reader requests a specific range of the cached item's body (`to_stream_from_range`)
- The writer and reader are concurrent, i.e. the body is streamed from one to the other

then today, the core cache API will ignore the requested range and provide the whole body. There is no explicit notification that the whole body is provided instead of a range.

In this SDK release, this remains the default behavior. However, lookup calls (lookup, transaction lookup) now offer an `always_use_requested_range` option. If set to true, body reads conducted as part of this lookup/transaction/replacement will always use the requested range, even when streaming.

In a future (major) SDK release, the default behavior will change to `always_use_requested_range`.

### DIFF
--- a/runtime/fastly/builtins/cache-core.cpp
+++ b/runtime/fastly/builtins/cache-core.cpp
@@ -41,6 +41,7 @@ JS::Result<host_api::CacheLookupOptions> parseLookupOptions(JSContext *cx,
       return JS::Result<host_api::CacheLookupOptions>(JS::Error());
     }
     JS::RootedObject options_obj(cx, &options_val.toObject());
+
     JS::RootedValue headers_val(cx);
     if (!JS_GetProperty(cx, options_obj, "headers", &headers_val)) {
       return JS::Result<host_api::CacheLookupOptions>(JS::Error());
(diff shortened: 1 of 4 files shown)


zkat:
phew. Rebased this one and cleaned up the history.

runtime/fastly/builtins/cache-core.cpp
```diff
@@ -78,6 +79,20 @@ JS::Result<host_api::CacheLookupOptions> parseLookupOptions(JSContext *cx,
       }
       options.request_headers = host_api::HttpReq(request_handle);
     }
+
+    JS::RootedValue always_use_requested_range_val(cx);
+    if (!JS_GetProperty(cx, options_obj, "always_use_requested_range",
```
zkat, line 84: this should be camelCase

runtime/fastly/builtins/cache-core.cpp
zkat, line 29: remember to update this when you update the types.

zkat (changes requested):

PR #1193, opened 2025-08-29
HTML Rewriter

TartanLlama:
Adds an HTML rewriter feature with an interface mostly the same as [Akamai's html-rewriter](https://techdocs.akamai.com/edgeworkers/docs/htmlrewriter). Uses the [`lol-html`](https://github.com/cloudflare/lol-html) library under the hood.

A few potentially questionable design decisions which we may want to go a different way on:

- I copied the source of `lol-html`'s C API into this repo. This was done for a few reasons:
  - The upstream version uses a [single function call](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) that is from a newer `rustc` version that we build with, so I patched that change.
  -  The C API crate is in a [subdirectory of the main `lol-html` repo](https://github.com/cloudflare/lol-html/tree/main/c-api) and is not published on crates.io.
  - The upstream version specifies a different `lib.name` in its `Cargo.toml`, which is incompatible with StarlingMonkey's [`add_rust_lib`](http://github.com/bytecodealliance/StarlingMonkey/blob/main/cmake/build-crates.cmake) function, so I patched that too.
  - I think ideally we'd change the version of `rustc` we compile with (which is a change made in upstream StarlingMonkey already), patch `add_rust_lib` to support custom `lib.name`s, then grab the C API crate through `cargo` with a simple wrapper library, similar to the existing crates in StarlingMonkey. Alternatively, we could make a fork of `lol-html` and point at that instead.
- I did not implement the `insert_implicit_close` feature of Akamai's API, because it seems less important and I don't see a simple way to do this from `lol-html`. I also extended their API with a `escapeHTML` option for all insertion functions, which allows inserting HTML content as text.
- I added the types for the rewriter to the global object, mostly because I couldn't seem to get it working with solely having them in `fastly:html-rewriter`. Ideally someone who knows their way around the module system better can tell me what to do here 😄

### DIFF
--- a/documentation/docs/html-rewriter/Element/prototype/after.mdx
+++ b/documentation/docs/html-rewriter/Element/prototype/after.mdx
@@ -1,0 +1,35 @@
+---
+hide_title: false
+hide_table_of_contents: false
+pagination_next: null
+pagination_prev: null
+---
+
+# after
+
+The `after` method inserts content after the closing tag of the element.
+
+## Syntax
+
+```js
+element.after(content);
+element.after(content, options);
+```
+
+### Parameters
+
+- `content` _: string_
+  - The content to insert after the element's closing tag.
+
+- `options` _: ElementRewriterOptions_
+  - An optional object that can have the following properties:
+    - `escapeHTML` _: boolean_
+      - If `true`, any HTML markup in `content` will be escaped so it is safe to insert as text.
+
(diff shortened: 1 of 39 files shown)


TartanLlama:
@IGx89 great to hear, thank you!

PR #1237, opened 2025-12-02
docs: Update html-rewriter import in docs

bricksroo:
This PR updates the import statement in docs for `fastly:html-rewriter`. The previous `fastly/html-rewriter` causes a TS error: `Cannot find module 'fastly/html-rewriter' or its corresponding type declarations.`


TartanLlama:
Thank you!

PR #1280, opened 2026-01-22
feat: Map the FastlyStatus values into KVStoreError values the same way the Rust SDK does

computermouth:
In the Rust SDK, we check the hostcall error first, and map it into a KVStoreError before processing the KV context-specific error returned by the hostcall (https://github.com/fastly/fst-compute-sdk-rs/blob/main/fastly/src/kv_store/handle.rs#L440-L463). 
```
        status.result().map_err(|st| match st {
            FastlyStatus::BADF => KVStoreError::InvalidStoreHandle,
            FastlyStatus::INVAL => KVStoreError::ItemBadRequest,
            FastlyStatus::LIMITEXCEEDED => KVStoreError::TooManyRequests,
            _ => st.into(),
        })?;

        if kv_sys_error != KvSysError::Ok {
            return Err(kv_sys_error.into());
        }
```

### DIFF
--- a/documentation/docs/device/Device/prototype/isBot.mdx
+++ b/documentation/docs/device/Device/prototype/isBot.mdx
@@ -1,0 +1,15 @@
+---
+hide_title: false
+hide_table_of_contents: false
+pagination_next: null
+pagination_prev: null
+---
+# Device.prototype.isBot
+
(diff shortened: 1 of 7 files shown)


zkat:
Looks like you actually need to _use/handle_ this, not just add it :)

PR #1300, opened 2026-02-05
working on new shortcutting strategy

zkat:


runtime/fastly/builtins/fetch/fetch.cpp
zkat: this is definitely happening...

runtime/fastly/builtins/fetch/request-response.cpp
zkat: How can this be true if `ShortcuttedVia` is never set to anything???

zkat:
yes

PR #1307, opened 2026-02-12
feat: stale-if-error support

TartanLlama:
Based on the Rust SDK implementation


cceckman-at-fastly (commented):
Nice! One nit and one symbol fix

integration-tests/js-compute/fixtures/app/src/stale-if-error.js
```diff
@@ -0,0 +1,417 @@
...
+  routes.set(
+    '/stale-if-error/response/setter-throws-on-non-cached',
+    async () => {
+      const response = new Response('test body');
+      assertThrows(
+        () => {
+          response.staleIfError = 300;
+        },
+        TypeError,
+        'Response set: staleIfError must be set only on unsent cache transaction responses',
```
cceckman-at-fastly, line 71: nit: Is this the same criterion as "candidate responses", below?

In the Rust universe, we have separate types for the two, but I think they're the same in JS -- the restriction makes sense here, just might be clearer if it's using the same language.

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -910,6 +924,47 @@ bool stream_back_then_handler(JSContext *cx, JS::HandleObject request, JS::Handl
   // response process.
   auto cache_write_options = Response::override_cache_options(response_obj);
   MOZ_ASSERT(cache_write_options);
+
+  // Check if we should use stale-if-error response instead of this error response
+  auto state_res = cache_entry.get_state();
+  if (!state_res.is_err()) {
+    auto cache_state = state_res.unwrap();
+    auto status = Response::status(response_obj);
+    if (cache_state.is_usable_if_error() && (status >= 500 && status < 600)) {
```
cceckman-at-fastly: If I'm reading this right -- I don't think we should have the 5xx check here.

[Here](https://www.fastly.com/documentation/solutions/tutorials/stale/#deal-with-origins-that-are-erroring), we say about the Delivery behavior:

> Fastly will, by default, accept and serve any response that is syntactically valid HTTP, which includes error responses.

This is also how the Rust SDK works: if there's a stale-if-error response in cache, and we get a 5xx from the origin, we return the 5xx to the user by default. The user can override that (in `vcl_fetch`, or in `after_send` in Rust) by matching the 5xx and returning an error from that handler... which causes the `send` (Rust) to return the stale response.

If I'm understanding this code right -- this precludes the ability for the guest to _prefer_ the 5xx to the stale-if-error? Like, if the server returns a 5xx, they'll just always get SIE instead? I don't think we want that.

(Note: the only case in which Delivery "forces" the stale-if-error response is if the backend is failing a health check, according to [here](https://www.fastly.com/documentation/guides/concepts/cache/stale/); all other cases require an explicit choice of `deliver_stale`.)

cceckman-at-fastly: (I may be misreading where in the stack this code lives, though, so please correct me if so!)

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -1102,6 +1164,69 @@ namespace fastly::fetch {
...
+  if (!cache_state.is_usable_if_error()) {
+    return std::nullopt;
+  }
+
+  // We have a usable stale-if-error response
+  auto chose_stale_res = cache_entry.transaction_choose_stale();
+  if (auto *err = chose_stale_res.to_err()) {
+    HANDLE_ERROR(cx, *err);
+    return std::nullopt;
+  }
```
cceckman-at-fastly, line 1153: Hm, I need a bigger picture here...

`usable_if_error` can show up in two cases:
1. This client is a follower of a request collapse. Revalidation has already happened, and failed, so `lookup` returns the stale-if-error response as "immediately available" (`FOUND | USABLE` are also set).
2. This client is the leader of a request collapse. In this case the `MUST_INSERT_OR_UPDATE` flag is also set.

`transaction_choose_stale` should only be called in case (2). Are there any paths by which `try_serve_stale_if_error` can be hit from case (2)?

(In case (1), we shouldn't be performing the backend `send` or evaluating the `beforeSend` / `afterSend` hooks, so... probably OK?)

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -1102,6 +1164,69 @@ namespace fastly::fetch {
...
+
+  JS::RootedValue no_candidate(cx);
+  auto maybe_response = get_found_response(cx, cache_entry, request, no_candidate, false);
+  if (maybe_response.has_value() && !maybe_response.value()) {
+    return std::nullopt;
+  }
+
+  if (!maybe_response.has_value()) {
+    return std::nullopt;
+  }
```
cceckman-at-fastly: Just a nit-- seems backwards, as `has_value` is checked twice?

```suggestion
  if (!maybe_response.has_value()) {
    return std::nullopt;
  }
  if (!maybe_response.value()) {
    return std::nullopt;
  }
```

or given short-circuit "or"

```suggestion
  if (!maybe_response.has_value() || !maybe_response.value()) {
    return std::nullopt;
  }
```

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -1176,8 +1301,14 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
...
-  auto transaction_res =
-      host_api::HttpCacheEntry::transaction_lookup(request_handle, override_key_hash);
+  host_api::Result<host_api::HttpCacheEntry> transaction_res;
+  if (override_key_hash.empty()) {
+    transaction_res = host_api::HttpCacheEntry::transaction_lookup(request_handle, {});
+  } else {
+    transaction_res = host_api::HttpCacheEntry::transaction_lookup(
+        request_handle, std::span<uint8_t>{override_key_hash.data(), override_key_hash.size()});
+  }
+
```
cceckman-at-fastly, line 1262: Seems like this would be the same as

```suggestion
  transaction_res = host_api::HttpCacheEntry::transaction_lookup(
        request_handle, std::span<uint8_t>{override_key_hash.data(), override_key_hash.size()});
```

right? Since `.empty()` implies `.size() == 0` -- or is there something about whether the pointer has data or not that matters?

runtime/fastly/builtins/fetch/fetch.cpp
```diff
@@ -1272,19 +1416,45 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
...
-      RequestOrResponse::close_if_cache_entry(cx, request);
-      return false;
+      if (cache_state.is_usable_if_error()) {
+        // We've got a stale-if-error response, so swap it out for the error and notify any
+        // request collapse followers.
+        auto chose_stale_res = cache_entry.transaction_choose_stale();
+        if (auto *err = chose_stale_res.to_err()) {
+          HANDLE_ERROR(cx, *err);
+          return false;
+        }
```
cceckman-at-fastly, line 1381: Trying to understand the interaction between this and `try_serve_stale_with_error`... is it: `try_serve_stale_if_error` is only invoked as part of promise processing, and this bit runs if the error occurs "not in a promise"?

runtime/fastly/builtins/fetch/request-response.cpp
```diff
@@ -4021,6 +4043,35 @@ bool Response::swr_get(JSContext *cx, unsigned argc, JS::Value *vp) {
   return true;
 }
 
+bool Response::staleIfError_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  auto entry = RequestOrResponse::cache_entry(self);
+
+  // all caching paths should set the override options as the final options
+  // so if they aren't set we are in the undefiend cases of no caching API use / no hostcall support
```
cceckman-at-fastly:
```suggestion
  // so if they aren't set we are in the undefined cases of no caching API use / no hostcall support
```

cceckman-at-fastly (changes requested):
Mostly nits & clarifying questions -- but the thing on 5xx handling is a blocker.

runtime/fastly/builtins/fetch/fetch.cpp
TartanLlama: Got it, I've removed this behaviour

runtime/fastly/builtins/fetch/fetch.cpp
TartanLlama: Yeah, this bit was for synchronous failures, but you're right, we can reuse that function here

integration-tests/js-compute/fixtures/app/src/stale-if-error.js
TartanLlama: It is, the difference in working is because:
- The first error is given the same wording as errors thrown from other setters like `ttl`, which I thought was good for consistency across properties
- The second error uses more precise wording because there's not an existing corresponding error

I could make the two consistent, but I think this is okay

cceckman-at-fastly (approved):

zkat:
Looks like this needs conflicts resolved now, since things have moved along a bit.

zkat (approved):

cceckman-at-fastly (approved):

dgryski:
Merge conflicts need to be handled.

PR #1312, opened 2026-02-16
Allow passing custom weval binary path

tyler-goodwin:
We make use of bazel, which doesn't allow downloading the weval binary, and instead we need to vendor it, so we need a way to override how weval is called.


tyler-goodwin:
> Thanks! Looks reasonable, but would having this as a command line argument to the CLI work for you, for consistency with other customisation points? Something like `--weval-path`. 


Yep! That'd be perfect too.

PR #1353, opened 2026-03-04
Docs parity - backends

case-fastly:
This addresses #1328.


types/backend.d.ts
```diff
@@ -228,6 +260,10 @@ declare module 'fastly:backend' {
   /**
    * Class for dynamically creating new [Fastly Backends](https://developer.fastly.com/reference/api/services/backend/).
    *
+   * **Note**: Dynamic backends are by default disabled at the Fastly service level.
```
TartanLlama: I think this isn't true anymore, no?

PR #1425, opened 2026-04-17
fix: Check pending exceptions between requests in reusable sandbox mode

TartanLlama:
If an exception occurs outwith the main event loop (e.g. when sending a `500` for a request that has not been responded to), it can still be pending when the next request handler runs. This PR checks for one between requests and, if there is one, throws an error.

### DIFF
--- a/runtime/fastly/handler.cpp
+++ b/runtime/fastly/handler.cpp
@@ -177,7 +177,7 @@ int main(int argc, const char *argv[]) {
       return -1;
     }
 
-    // In case there was an exception pending from the previous request handling, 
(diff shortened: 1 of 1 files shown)


runtime/fastly/handler.cpp
zkat, line 124: for future reference: I think we should have this function actually return a true/false value, propagate things from any calls inside it, and break from this loop if we get a `false`, rather than rely solely on pending exceptions.

zkat (approved):
This is good for now, but I've noted a change I'd like to see in the future.

PR #1426, opened 2026-04-17
fix: Reset StarlingMonkey engine between requests

TartanLlama:
This PR attempts to clear out the StarlingMonkey engine as much as is reasonable between requests. We should submit these changes upstream so that when we eventually upgrade our StarlingMonkey dependency, we can get rid of the patching.


zkat (approved):

zkat:
In the future, we should move away from a shared ENGINE and just instantiate a new one for every new request, imo, but I think this works for now.

TartanLlama:
@zkat I think that would mean re-running the top-level parts of the guest JS, which is what reusable sandboxes are trying to avoid, no?

PR #1427, opened 2026-04-17
Add integration test for verifying when `FetchEvent` gets marked done

ulyssa:
This follows up on #1416 and adds tests to catch when the `FetchEvent` was not marked as finished after a WebSocket or Fanout handoff.


TartanLlama (approved):

TartanLlama:
Replaced with https://github.com/fastly/js-compute-runtime/pull/1435 so the token-based CI can run (we should work out some safe method of enabling this for fork PRs)

PR #1440, opened 2026-04-23
feat: Bot detection API

TartanLlama:
Implements the bot detection API, mostly passing directly through to the hostcalls.

Marked as draft while I determine the best way to test this, and update the docs and types.

I added some helpers to `host_api.cpp` for factoring out some of the boilerplate. We could apply these patterns to much of that file. Particularly, the buffer-resizing functions are a bit finnicky and I'd feel better about having that work centralised.


zkat:
superseded by #1501

PR #1441, opened 2026-04-23
Implemented checkSSL

pablodelolmo:
### Rationale behind this PR

One of the features currently present in our delivery services when defining backend is the possibility to disable TLS certificate validation when reaching backends (see `ssl_check_cert` in [our API docs](https://www.fastly.com/documentation/reference/api/services/backend/)). We currently support that in our Compute platform but the way of doing it is to not define the backend certificate fields such as `sniHostname`, `certificateHostname` and `caCertificate`. The intention of adding this is to have the same explicit field and handle the behind the scenes removal of those fields when dealing with the Compute runtime.

### What's in this PR

- The SDK & runtime changes to do that.
- Integration tests
- Updated documentation


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fastly/js-compute",
-  "version": "3.41.1",
+  "version": "3.41.2",
```
zkat: we'll bump this ourselves, no need to do it in your PR.

integration-tests/js-compute/fixtures/module-mode/src/dynamic-backend.js
```diff
@@ -1968,6 +1968,76 @@ routes.set('/backend/timeout', async () => {
...
+          assertDoesNotThrow(() => {
+            new Backend({
+              name: 'checkSSL-false-with-default-ssl-fields',
+              target: 'a',
+              checkSSL: false,
+            });
+          });
+        },
+      );
+    }
```
zkat, line 2039: Can you add a test that actually hits https://http-me.fastly.dev?

pablodelolmo:
Added three sets of tests for three different scenarios. The scenarios are:

- Expired cert
- Wrong hostname in the cert
- Self-signed cert

The set of tests for each one are:

- Without `checkSSL` (which defaults to `true`) and defining the SNI & CN -> Fetch fails (test successful)
- With `checkSSL` set to `true` and defining the SNI & CN -> Fetch fails (test successful)
- With `checkSSL` set to `false` and defining the SNI & CN -> Fetch goes through (test successful)

zkat:
Closing this for now, per discussion

PR #1447, opened 2026-04-30
fix: Protect against GC from `validate_bytes`

TartanLlama:
In the case that it is given an `ArrayBufferView`, `validate_bytes` returns a pointer to a buffer that may be GC'ed. The current uses of `validate_bytes` are careful to not add any potentially GC-triggering calls between the call to `validate_bytes` and hostcalls, but this may change in the future, and there is nothing in place to ensure the GC does not occur. As such, this PR changes `validate_bytes` to additionally return an optional GC guard, that calling code can hold on to to ensure there is no GC at runtime while the pointer is live.

Honestly I'm not sure this is all worth it and I'm tempted to just return a `std::vector` from that function, but I figured I'd avoid regressing performance in this case while adding some additional safety.

There is no test for this PR, because it is adding defensive code for potential future uses of `validate_bytes`, or changes to existing callers.


zkat (approved):

zkat:
This has some build errors

PR #1448, opened 2026-04-30
fix: Numeric overflow issue in Core Cache API

TartanLlama:
Numeric overflow of several parameters in the Core Cache API are checked *after* computation, so overflow could have already occurred. Furthermore, for `staleWhileRevalidate`, the wrong property was being checked for overflow. This PR fixes these issues and adds regression tests for them.

### DIFF
--- a/integration-tests/js-compute/fixtures/app/tests.json
+++ b/integration-tests/js-compute/fixtures/app/tests.json
@@ -1931,6 +1931,9 @@
   "GET /core-cache/insert/options-parameter-maxAge-field-negative-number": {
     "environments": ["compute"]
   },
+  "GET /core-cache/insert/options-parameter-maxAge-field-too-large": {
+    "environments": ["compute"]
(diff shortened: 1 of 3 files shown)


integration-tests/js-compute/fixtures/app/tests.json
```diff
@@ -1931,6 +1931,9 @@
   "GET /core-cache/insert/options-parameter-maxAge-field-negative-number": {
     "environments": ["compute"]
   },
+  "GET /core-cache/insert/options-parameter-maxAge-field-too-large": {
+    "environments": ["compute"]
```
zkat, line 1935: I believe cache tests can now run on viceroy. I will check.

zkat: we'll deal with this later. closing this out for now.

zkat (approved):

PR #1471, opened 2026-05-07
fix: revert validate_bytes change to get CI in a good place again

zkat:


zkat:
Closing in favor of #1475

PR #1474, opened 2026-05-07
fix: Correct GC guard type in KVStore::put to prevent wasm unreachable crashes

mcinerney:


zkat:
closing in favor of #1475

PR #1477, opened 2026-05-12
chore: upgrade node version in CI

zkat:
`setup-node@v3` uses `node@20`, which is now EOL. This bumps it to `node@24`, the latest LTS.


dgryski (approved):
LGTM

PR #1482, opened 2026-05-19
fix: Make headers of cloned requests independent

TartanLlama:
Headers for requests that are cloned through `req.clone()` are currently shared. This PR makes them independent.


zkat (commented):
I'm confused.. This PR only seems to include the test?

zkat:
oh I see. The fmt reverted your changes, but they still passed the tests, so maybe the tests need to be revisited.

zkat (approved):

PR #1488, opened 2026-05-21
fix: reset SDK state between requests

zkat:
This PR makes it so certain global shared state gets properly reset between different container sessions.


integration-tests/js-compute/fixtures/reusable-sandboxes/src/dynamic-backend.js
TartanLlama, line 1: Can we not just make two requests to the same endpoint?

runtime/fastly/handler.cpp
```diff
@@ -17,12 +18,32 @@ namespace fastly::runtime {
 
 api::Engine *ENGINE;
 
+int restore_builtin_state() {
```
TartanLlama: Why return `int` rather than `bool`?

zkat: because I brain farted and my brain was in `int main` mode lol

dgryski:
Looks reasonable.

dgryski (approved):
LGTM

PR #1394, opened 2026-03-26
feat: StarlingMonkey upgrade

TartanLlama:
- Bumps wasi-sdk to v30 from v20
- Bumps rustc to 1.88 from 1.81
- Moves from the old NPM wizer package to pulling down `wasmtime` and using the version of wizer bundled with `wasmtime`
- Uses the new policy-based way of defining finalizable builtins
- Stubs out parts of `AbortSignal` and `AbortController` that are usid in other places in StarlingMonkey
- Fixes compiler diagnostics that are triggered based on the wasi-sdk bump
- Fixes misplaced namespace in `fastly.h`
- Removes the now-unnecessary copy of `rust-toolchain.toml` 
- Updates wpt expectations

Docs still todo

PR #1490, opened 2026-06-11
feat: Upgrade StarlingMonkey

zkat:
This supersedes #1394, since we're moving to land this into `main` proper.


zkat:
wow this is an incredibly annoying dependency issue: the vulnerability is in NPM, which bundles its dependencies, which means `npm audit fix` can't touch it, but still detects it as a vulnerability, even with `--force`.

image

I can try using NPM overrides, but those might also not work on bundled dependencies, I'll that next.

Issue #1492, opened 2026-06-16
StarlingMonkeyFormBoundary does not conform to RFC 7231 and causes parse errors


sudheeshskrishna:
Noticed an issue in Fastly Compute JS runtime library. While handling multipart formdata, it uses StarlingMonkeyFormBoundary. But the boundary generated sometimes contain **/** which is not a valid character for boundary as per RFC 7231. Emitting an unquoted boundary with non-token characters is technically a malformed Content-Type and strict parsers will reject it.
Our apps are rejecting such malformed boundaries and hence we were facing unexplained HTTP 5xx errors in our apps. Can this be taken up and addressed at the earliest please?

sudheeshskrishna:
Closing this since it is getting tracked under support ticket.

PR #1493, opened 2026-06-16
fix: Make boundary parameter of multi-part form data requests safe for use in `Content-Type` headers

TartanLlama:
StarlingMonkey form data encoder uses random Base64 data as part of the boundary. Base64 can contain `+` and `/` characters, which can make them invalid when used unquoted in `Content-type` headers, and is known to cause issues with some clients.

While [RFC 2046, section 5.11](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1) allows for and recommends quoting the `boundary` parameter, the [HTML spec does not allow it](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm) and [some Web Platform tests](https://github.com/web-platform-tests/wpt/blob/a72c94d4a4df0811b49751b5dc922c8a517aeebc/fetch/api/response/response-init-contenttype.any.js#L30) fail with an implementation that does so.

Rather than quoting the strings, I've elected to patch StarlingMonkey to ensure that invalid characters are not produced, limiting the output to alphanumeric characters. This follows the approach taken in [WebKit](https://chromium.googlesource.com/chromium/src/+/refs/heads/master/third_party/blink/renderer/platform/network/form_data_encoder.cc#119), and uses the same approach of simply replacing `+` and `/` in the Base64 encoding table with `A` and `B`.

### DIFF
--- a/integration-tests/js-compute/fixtures/app/src/form-data.js
+++ b/integration-tests/js-compute/fixtures/app/src/form-data.js
@@ -2,21 +2,26 @@ import { assert, assertDoesNotThrow } from './assertions.js';
 import { routes } from './routes.js';
 
 routes.set('/form-data/quoted-boundary', async () => {
-    const data = new FormData();
-    data.append('test', 'field');
-    let response = await fetch('https://http-me.fastly.dev/anything', {
-        method: 'POST',
-        body: data
-    });
-    assert(response.ok, true, 'response.ok === true');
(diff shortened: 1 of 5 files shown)


dgryski:
Originally reported issue https://github.com/fastly/js-compute-runtime/issues/1492

zkat (approved):

PR #1500, opened 2026-06-29
feat: update StarlingMonkey

zkat:
This commit is a no-op due to a PR merging mistake, but the changes can be seen over at https://github.com/fastly/js-compute-runtime/pull/1490


dgryski (approved):
I'll assume this all works then.

PR #1508, opened 2026-07-21
Release 3.43.2

zkat:
We need to reset `main` to only have one (or a couple) of critical patches so we can release sooner, since what is _currently_ in `main` is blocked on several issues, plus the starlingmonkey upgrade.


dgryski (approved):
LGTM

dgryski (approved):
LGTM

PR #1518, opened 2026-07-22
fix(tests): Make test suite properly fail when a test fails

zkat:
(Note: if this is still failing, it's because #1520 hasn't been merged yet)


dgryski (approved):
LGTM

PR #1519, opened 2026-07-22
fix(cache): handle error

zkat:
This was making the build fail, but I think it may have only affected wasi-sdk-30 (the starlingmonkey upgrade branch). Either way, we shouldn't be ignoring errors like this.


dgryski (approved):
LGTM

Can we scan for other places we're failing to handle errors? (comby/semgrep/...)

zkat:
I scanned through and found a couple other places yeah

PR #1520, opened 2026-07-22
fix(tests): fix failing cache test

zkat:
I think this is our last test failure


dgryski (approved):
LGTM.  Actually read through it this time.

PR #1526, opened 2026-07-27
Release 3.45.0-pre.0

zkat:
This is a prerelease that includes a big StarlingMonkey upgrade that we want to do some testing on without affecting our wider user community unless they opt in.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,13 +99,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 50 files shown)


dgryski (approved):
I believe you.

briancarbone:
Will this update provide AbortSignal and AbortController?

TartanLlama:
@briancarbone not immediately; the StarlingMonkey upgrade provides the JS-level machinery needed for `AbortSignal` and `AbortController`, but there's additional work needed on the Fastly Compute platform to supply the underlying cancellation support. So it's a step in that direction, but won't be coming with this patch.

PR #1535, opened 2026-07-29
fix: dynamic backends disabled exception

TartanLlama:
There were two mistakes made in the creation of exceptions for when dynamic backends are disabled at the service level:

1. The wrong error message enum was used
2. One of the error message definitions was invalid

This PR fixes both of these. No test is added to CI because our CI service has dynamic backends enabled.


zkat (commented):
We should add tests against Viceroy, although that'll require changes to Viceroy itself. Still, it means we'll have actual tests.

zkat (approved):
Actually this is trivial enough that I'm happy to merge it for now, and just write tests soon?

Issue #1178, opened 2025-04-18
The `Vary` response header seems to be getting invalidated when using the HTTP Cache API


vcarvajal-sigsci:
There appears to be a bug with regard to how the `vary` response header is being handled in using the http caching option. When the flag is set, --enable-http-cache, upon building the package, vary header is invalidated regardless of the use of HTTP Cache API (ie afterSend), however if the flag is not used, it works as expected.

PR #1536, opened 2026-07-30
fix: Commit request prior to cache lookup when using guest caching

TartanLlama:
Currently, headers are not committed prior to `transaction_lookup` calls when using guest caching. As a result, if any headers are set that are marked as part of a `Vary` header on a cached origin response, the cache lookup does not take them into account. This PR ensures that headers are committed in time.

### DIFF
--- a/runtime/fastly/builtins/fetch/fetch.cpp
+++ b/runtime/fastly/builtins/fetch/fetch.cpp
@@ -895,6 +895,7 @@ std::optional<JSObject *> get_found_response(JSContext *cx, host_api::HttpCacheE
       HANDLE_ERROR(cx, *err);
       return nullptr;
     }
+    fastly_push_debug_message(std::string(vary_res.unwrap().value()));
     cache_options->vary_rule = std::move(vary_res.unwrap());
     auto surrogate_keys_res = cache_entry.get_surrogate_keys();
     if (auto *err = surrogate_keys_res.to_err()) {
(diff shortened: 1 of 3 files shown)


TartanLlama:
@cceckman-at-fastly do you think you could have a look at this and tell me if this is the right place to make the hostcalls necessary to update headers when using guest caching?

TartanLlama:
Fixes https://github.com/fastly/js-compute-runtime/issues/1178

runtime/fastly/builtins/fetch/fetch.cpp
cceckman-at-fastly, line 1233: Assuming `fetch_send_body` above flushes the headers for the non-guest-caching cases:

I think we need to make sure the host & guest are in sync as of _this_ call, which should map to the `is_request_cacheable` hostcall. There are some things the RFC says about how the `Authorization` header affects cacheability ("it's complicated" but we want to be sure we have the data here) -- and if they have changed the method, that plays in to it as well.

That touches on my other thought: is `commit_headers` just headers, or is it also method & URI? We should make sure those are flushed as well

TartanLlama: Got it, I've moved the call. `commit_headers` only commits the headers, but the method and URI are set eagerly, so they don't need flushed

cceckman-at-fastly (approved):

PR #1549, opened 2026-07-31
fix: starlingmonkey update already fixed urlsearchparams memory issue

zkat:
We previously fixed a urlsearchparams memory issue in main, but that issue has now been fixed in the updated starlingmonkey, so we don't need our patch anymore. This PR keeps the test, but removes the patch.


TartanLlama:
Can you add a description to the PR and make the title something that will show up well in the changelog?

zkat:
@TartanLlama done

PR #1565, opened 2026-08-17
feat!: make module mode the default, add --disable-top-level-await opt-out

harmony7:
This PR makes moduleMode (`--enable-experimental-top-level-await`) the default.

Users can use `--disable-top-level-await` if they want to switch to the old behavior of emitting an IIFE. Note that the flag is named for its most visible effect, but classic script mode also drops strict-mode-by-default (so code that assigns to undeclared variables builds again) and `import.meta`.

`--enable-experimental-top-level-await` is kept as a no-op so existing pipelines do not break, and `--module-mode` (unbundled ESM) is unchanged.

Targets 4.0.0: rebased onto `4ea3c3ba` and retargeted at `release-4.0.0`, so this does not land in 3.x.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -11,6 +11,7 @@ const strictOptionsMap = {
   enableExperimentalHighResolutionTimeMethods:
     '--enable-experimental-high-resolution-time-methods',
   enableExperimentalTopLevelAwait: '--enable-experimental-top-level-await',
+  disableTopLevelAwait: '--disable-top-level-await',
   enableStackTraces: '--enable-stack-traces',
   excludeSources: '--exclude-sources',
   debugIntermediateFiles: '--debug-intermediate-files',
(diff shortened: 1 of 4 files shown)


TartanLlama:
Can you target this at the https://github.com/fastly/js-compute-runtime/commits/release-4.0.0 branch?

src/config.ts
```diff
@@ -11,6 +11,7 @@ const strictOptionsMap = {
   enableExperimentalHighResolutionTimeMethods:
     '--enable-experimental-high-resolution-time-methods',
   enableExperimentalTopLevelAwait: '--enable-experimental-top-level-await',
+  iife: '--iife',
```
TartanLlama: I wonder if `--disable-top-level-await` would be a better name. What do you think @zkat ?

zkat: agree

harmony7: sure I can change it

harmony7:
> target this at the https://github.com/fastly/js-compute-runtime/commits/release-4.0.0 branch

will do

harmony7:
1. rebased to `release-4.0.0` branch
2. renamed the flag to `--disable-top-level-await` and updated its description.

please review

zkat (approved):

PR #1568, opened 2026-08-18
fix: Don't throw from getters

### LINKED ISSUE #1308: Reconsider throwing from property getters
In some cases, e.g. https://github.com/fastly/js-compute-runtime/pull/1306, property getters throw exceptions in places where it may be surprising to users. We should reconsider where and when we do this.

I think it's reasonable to throw an exception for errors that are outwith API contracts, but within reasonable use, property getters should return `null` instead of throwing on error. If the error information is important, those properties should be functions instead.

TartanLlama:
It is very surprising for users if property accesses throw. We should never require putting try catch blocks around property accesses. This PR changes all instances of this in the codebase (that I could find) to instead return `null`, `undefined`, or `false` depending on the context.

Fixes #1308

### DIFF
--- a/runtime/fastly/builtins/backend.cpp
+++ b/runtime/fastly/builtins/backend.cpp
@@ -1573,7 +1573,7 @@ bool Backend::is_ssl_get(JSContext *cx, unsigned argc, JS::Value *vp) {
   }
   auto res = backend->is_ssl();
   if (auto *err = res.to_err()) {
-    args.rval().setBoolean();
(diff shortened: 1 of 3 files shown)


zkat:
This might be something we need to target at v4 instead, since imo it's a breaking change. What do you think?

TartanLlama:
I think that's reasonable, yeah

zkat:
this has been merged into https://github.com/fastly/js-compute-runtime/pull/1551

PR #1570, opened 2026-08-18
fix: Avoid arithmetic overflow in content_stream_read_then_handler

TartanLlama:
There are a couple places in which overflowing an integer in `content_stream_read_handler` could cause buffer overrun issues. This PR checks for overflow to avoid this.

### DIFF
--- a/integration-tests/js-compute/fixtures/app/src/client.js
+++ b/integration-tests/js-compute/fixtures/app/src/client.js
@@ -124,3 +124,15 @@ routes.set('/client/ohFingerprint', (event) => {
     );
   }
 });
+
+routes.set('/client/clientSNI', (event) => {
+  if (isRunningLocally()) {
+    strictEqual(event.client.clientSNI, null);
+  } else {
+    strictEqual(
+      typeof event.client.clientSNI,
+      'string',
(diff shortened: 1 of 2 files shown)


integration-tests/js-compute/fixtures/app/src/client.js
```diff
@@ -124,3 +124,15 @@ routes.set('/client/ohFingerprint', (event) => {
...
+  if (isRunningLocally()) {
+    strictEqual(event.client.clientSNI, null);
+  } else {
+    strictEqual(
+      typeof event.client.clientSNI,
+      'string',
+      'typeof event.client.clientSNI',
+    );
+  }
+});
```
zkat: I don't think these changes were supposed to be included in this PR?

zkat (changes requested):

integration-tests/js-compute/fixtures/app/src/client.js
zkat:
```suggestion
```

zkat (changes requested):

zkat (approved):

zkat (approved):

PR #1575, opened 2026-08-24
fix: Improve restoring of global state

TartanLlama:
Restoring of global state is a bit ad-hoc right now, and misses some members. Ideally, all global state should be reset back to what it was at the end of wizer time at the end of each request. This PR adds a new global state framework for achieving this, and adapts the existing Fastly and Backend builtins.

The basic idea is that all global state is encapsulated into per-builtin `RequestState` types, which define how to initialize and snapshot them. Their lifecycle looks like this:

- Initialized exactly once on sandbox creation (at Wizer time)
- At the end of Wizer time, all `RequestState`s are snapshotted and stored aside
- During a request, all mutations occur on the current global state
- At the end of a request, the snapshot is restored to the current global state in preparation for the next request

Additionally, every global variable in the SDK should be commented with `// GLOBAL` for ease of searching. There are a few globals that I have not placed into this new scheme because they don't need it. All such globals should have a documented reason to not be snapshotted and restored.

### DIFF
--- a/runtime/fastly/handler.cpp
+++ b/runtime/fastly/handler.cpp
@@ -14,28 +14,16 @@ using std::chrono::duration_cast;
 using std::chrono::microseconds;
 using std::chrono::system_clock;
 
+// This is the set of builtins that have RequestState types
+// that need resetting upon every request. Simply add new builtins
+// to this list, and they will be picked up by the reset machinery.
+// Note that this variable is only used for its type; it is a way of
+// carrying around a variadic template parameter pack.
+static constexpr fastly::state::BuiltinsList<fastly::fastly::Fastly, fastly::backend::Backend>
+    builtins_with_request_state;
+
 namespace fastly::runtime {
-
 api::Engine *ENGINE;
-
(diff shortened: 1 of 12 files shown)


runtime/fastly/handler.cpp
```diff
@@ -134,12 +122,19 @@ bool handle_incoming(host_api::Request req) {
...
+  if (!fastly::state::Manager::snapshot_all_request_states(builtins_with_request_state,
+                                                           ENGINE->cx())) {
+    return -1;
+  }
 
   host_api::HttpReqPromise::DownstreamNextOptions options;
-  if (Fastly::reusableSandboxOptions.between_request_timeout()) {
-    options.timeout_ms = static_cast<uint32_t>(
-        Fastly::reusableSandboxOptions.between_request_timeout().value().count());
+  if (Fastly::request_state.get().reusable_sandbox_options.between_request_timeout()) {
```
zkat: did you want to use `->` instead of `.get().`?

zkat (approved):

PR #1581, opened 2026-09-03
fix: Commit cache transaction in SimpleCache.getOrSet

TartanLlama:
There was a missing call to `CacheTransaction.commit()` in `SimpleCache::getOrSet`, which can cause pending exceptions in some cases due to the `CacheTransaction` destructor attempting to cancel the transaction. This PR adds some more defensive code around the destructor and ensures the transaction is committed in the cache hit path. It also uses `CacheHandle::close` rather than `CacheHandle::transaction_cancel`, as this closes the read end as well as the write end.

### DIFF
--- a/runtime/fastly/builtins/cache-simple.cpp
+++ b/runtime/fastly/builtins/cache-simple.cpp
@@ -190,6 +190,12 @@ public:
       host_api::handle_api_error(this->cx, *err, this->line, this->func);
     }
 
+    if (!JS_IsExceptionPending(this->cx)) {
+      JS_ReportErrorUTF8(cx, "Cache transaction cancelled for unknown reason");
+    }
+
+    MOZ_ASSERT(JS::GetPromiseState(this->promise) == JS::PromiseState::Pending);
+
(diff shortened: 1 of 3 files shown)


runtime/fastly/builtins/cache-simple.cpp
```diff
@@ -444,6 +450,7 @@ bool process_pending_cache_lookup(JSContext *cx, host_api::CacheHandle::Handle h
     JS::RootedValue result(cx);
     result.setObject(*entry);
     JS::ResolvePromise(cx, promise_obj, result);
+    transaction.commit();
```
cceckman-at-fastly, line 453: In this setup, IIUC, the SDK will never call `handle.close()` on this handle. Correct?

cceckman-at-fastly (approved):

runtime/fastly/builtins/cache-simple.cpp
TartanLlama: I've changed `transaction_cancel` to `close`

zkat (approved):

PR #1586, opened 2026-09-07
fix: Close cache transaction if getOrSet handler throws

TartanLlama:
If the set handler for `SimpleCache.getOrSet` throws, the cache transaction is left open. The exception is rethrown, but if the guest catches this exception and continues to run, it will block other readers of that key. This PR ensures that the transaction is closed if the handler throws.


dgryski (approved):
LGTM

PR #1593, opened 2026-09-18
fix: handle potential NULL pointer dereference in debug logging

zkat:
No test because we can't really repro this in our infra, but it was good to have anyway.


dgryski (approved):
LGTM

